PR #235, opened 2015-04-16
Dev v1.1

wilmatrick:
1. The default fitting range in quasiisothermaldf.vmomentdensity() is 4*sigma. The default value for nsigma is 4. By changing 4*sigma to nsigma*sigma the default behaviour does not change, but by using the keyword nsigma other fitting ranges are also possible now. 

2. Cleaned up the coordinate transformation code used for the KuzminKutuzovStaeckelPotential by removing the Mathematica calculations (used for testing) again.

### DIFF
--- a/galpy/df_src/quasiisothermaldf.py
+++ b/galpy/df_src/quasiisothermaldf.py
@@ -472,21 +472,21 @@ class quasiisothermaldf(object):
             #Evaluate everywhere
             if isinstance(self._aA,(actionAngle.actionAngleAdiabatic,
                                     actionAngle.actionAngleAdiabaticGrid)):
-                vRgl= 4.*sigmaR1/2.*(glx+1.)
-                vzgl= 4.*sigmaz1/2.*(glx+1.)
+                vRgl= nsigma*sigmaR1/2.*(glx+1.)
+                vzgl= nsigma*sigmaz1/2.*(glx+1.)
                 vRglw= glw
                 vzglw= glw
             else:
-                vRgl= 4.*sigmaR1/2.*(glx12+1.)
+                vRgl= nsigma*sigmaR1/2.*(glx12+1.)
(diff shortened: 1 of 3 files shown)


jobovy:
Thanks, this looks good, I'll merge it. A few questions and comments:

- Why did you change the nsigma in vmomentdensity? It's good that you did, I'm just wondering why you did this now :-) There are a bunch of other functions in quasiisothermaldf that use 4 sigma (like pvz), that should eventually be fixed. This is really a fix that should be separate from the KuzminKutuzov potential addition.

- If you look at the test coverage (https://coveralls.io/builds/2351287), you see that's it's lower than that of my dev-v1.1, because some parts of your code are not tested. It would be good to add some additional tests to get the test coverage up again:

 - [ ] The first test of your notebook could be added to nose/test_potential.py, because you have a formula to compare to. So you could add something like test_KuzminKutuzovStaeckel_vcirc and then compare the galpy vcirc with the formula in a similar way as the other tests in that file. You can run the test with::

   nosetests -v nose/test_potential.py -x --nocapture

   to see whether they work.

 - [ ] You could add the test that the delta from estimateDeltaStaeckel is the same as the input delta in nose/test_actionAngle.py.

 - [ ] You should also add a test of the actions being conserved in nose/test_actionAngle.py. If you look at some of the functions that test the actionAngleStaeckel implementation, I think it should be straightforward to add this.

 - [ ] It would also be good to add tests of the new bovy_coords functions. These would probably be similar to the ones I wrote for Rz_to_uv etc.

I can help with all of this.


Once I've merged this, you probably want to merge in the changes I made in my 'wilma' branch to your dev-v1.1 branch. I think you can do the following

git remote add jobovy https://github.com/jobovy/galpy.git
git fetch jobovy

(these just mean that you have access to my branches). And then

git checkout dev-v1.1
git merge jobovy/wilma

Hopefully that won't give any merge conflicts.

PR #237, opened 2015-05-05
Test Suite for the KuzminKutuzovStaeckelPotential

wilmatrick:
Added a test suite for the KKS potential class.

### DIFF
--- a/galpy/potential_src/KuzminKutuzovStaeckelPotential.py
+++ b/galpy/potential_src/KuzminKutuzovStaeckelPotential.py
@@ -84,15 +84,10 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
-        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
-            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
-            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
-            out = nu.array([self._Rforce(rr,zz) for rr,zz in zip(R,z)])
-            return out
         l,n = bovy_coords.Rz_to_lambdanu    (R,z,ac=self._ac,Delta=self._Delta)
         jac = bovy_coords.Rz_to_lambdanu_jac(R,z,            Delta=self._Delta)
-        dldR = jac[0,0]
-        dndR = jac[1,0]
+        dldR = jac[0,0,:]
+        dndR = jac[1,0,:]
         return - (dldR * self._lderiv(l,n) + \
                   dndR * self._nderiv(l,n))
    
@@ -113,15 +108,10 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
-        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
-            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
-            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
-            out = nu.array([self._zforce(rr,zz) for rr,zz in zip(R,z)])
-            return out
         l,n = bovy_coords.Rz_to_lambdanu    (R,z,ac=self._ac,Delta=self._Delta)
         jac = bovy_coords.Rz_to_lambdanu_jac(R,z,            Delta=self._Delta)
-        dldz = jac[0,1]
-        dndz = jac[1,1]
+        dldz = jac[0,1,:]
+        dndz = jac[1,1,:]
         return - (dldz * self._lderiv(l,n) + \
                   dndz * self._nderiv(l,n))
 
@@ -141,18 +131,13 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
-        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
-            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
-            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
-            out = nu.array([self._R2deriv(rr,zz) for rr,zz in zip(R,z)])
-            return out
         l,n    = bovy_coords.Rz_to_lambdanu     (R,z,ac=self._ac,Delta=self._Delta)
         jac    = bovy_coords.Rz_to_lambdanu_jac (R,z,            Delta=self._Delta)
         hess   = bovy_coords.Rz_to_lambdanu_hess(R,z,            Delta=self._Delta)
-        dldR   = jac[0,0]
-        dndR   = jac[1,0]
-        d2ldR2 = hess[0,0,0]
-        d2ndR2 = hess[1,0,0]
+        dldR   = jac[0,0,:]
+        dndR   = jac[1,0,:]
+        d2ldR2 = hess[0,0,0,:]
+        d2ndR2 = hess[1,0,0,:]
         return d2ldR2       * self._lderiv(l,n)  + \
                d2ndR2       * self._nderiv(l,n)  + \
                (dldR)**2    * self._l2deriv(l,n) + \
@@ -175,18 +160,13 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
-        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
-            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
-            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
-            out = nu.array([self._z2deriv(rr,zz) for rr,zz in zip(R,z)])
-            return out
         l,n    = bovy_coords.Rz_to_lambdanu    (R,z,ac=self._ac,Delta=self._Delta)
         jac    = bovy_coords.Rz_to_lambdanu_jac(R,z,            Delta=self._Delta)
         hess   = bovy_coords.Rz_to_lambdanu_hess(R,z,            Delta=self._Delta)
-        dldz = jac[0,1]
-        dndz = jac[1,1]
-        d2ldz2 = hess[0,1,1]
-        d2ndz2 = hess[1,1,1]
+        dldz = jac[0,1,:]
+        dndz = jac[1,1,:]
+        d2ldz2 = hess[0,1,1,:]
+        d2ndz2 = hess[1,1,1,:]
         return d2ldz2       * self._lderiv(l,n)  + \
                d2ndz2       * self._nderiv(l,n)  + \
                (dldz)**2    * self._l2deriv(l,n) + \
@@ -210,20 +190,15 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
-        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
-            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
-            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
-            out = nu.array([self._Rzderiv(rr,zz) for rr,zz in zip(R,z)])
-            return out
         l,n    = bovy_coords.Rz_to_lambdanu    (R,z,ac=self._ac,Delta=self._Delta)
         jac    = bovy_coords.Rz_to_lambdanu_jac(R,z,            Delta=self._Delta)
         hess   = bovy_coords.Rz_to_lambdanu_hess(R,z,            Delta=self._Delta)
-        dldR = jac[0,0]
-        dndR = jac[1,0]
-        dldz = jac[0,1]
-        dndz = jac[1,1]
-        d2ldRdz = hess[0,0,1]
-        d2ndRdz = hess[1,0,1]
+        dldR = jac[0,0,:]
+        dndR = jac[1,0,:]
+        dldz = jac[0,1,:]
+        dndz = jac[1,1,:]
+        d2ldRdz = hess[0,0,1,:]
+        d2ndRdz = hess[1,0,1,:]
         return d2ldRdz              * self._lderiv(l,n)  + \
                d2ndRdz              * self._nderiv(l,n)  + \
                dldR*dldz            * self._l2deriv(l,n) + \
--- a/galpy/potential_src/KuzminKutuzovStaeckelPotential.py
+++ b/galpy/potential_src/KuzminKutuzovStaeckelPotential.py
@@ -84,10 +84,15 @@ class KuzminKutuzovStaeckelPotential(Potential):
         HISTORY:
             2015-02-13 - Written - Trick (MPIA)
         """
+        if isinstance(R,nu.ndarray) or isinstance(z,nu.ndarray):
+            if not isinstance(z,nu.ndarray): z = nu.ones_like(R)*z
+            if not isinstance(R,nu.ndarray): R = nu.ones_like(z)*R
+            out = nu.array([self._Rforce(rr,zz) for rr,zz in zip(R,z)])
+            return out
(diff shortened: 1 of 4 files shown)


jobovy:
Hi Wilma,

Thanks, this looks great! I think we're very close here. A few remaining issues:

- [ ] The test at the following link and the two asserts following that should probably have 10**-8.?: https://github.com/jobovy/galpy/pull/237/files#diff-8d0534a052faed8181aac8267e8974d9R243

- [ ] A few special cases of the coordinate transformation are not tested; can you come up with a test of those? https://coveralls.io/files/688326087#L1652 and https://coveralls.io/files/688326087#L1783

- [ ] The handling of the array input to the _zforce and the second derivatives for the Kuzmin-Kutuzov potential is not always tested: https://coveralls.io/files/688326029#L126 (and similar lower down). Is this handling of the array input really necessary? I have avoided this (as I mentioned in my comment on the previous pull request), so it would be better to avoid it here too. *If* we want to add this support, then it would be better to use a *decorator* (like the ones that I use in bovy_coords to parse the input to the coordinate transformation routines).

- [ ] I'm a little worried about the difference between Python and C, they shouldn't disagree by 10%. Are you sure that this doesn't happen when the coordinate gets very close to zero? Perhaps it would be better to combine tests of the relative and absolute difference: that is, test that abs_diff= numpy.fabs(C - Python)  < 10.**-5. + numpy.fabs(Python) * 10.**-5. (where 'Python' is the Python obtained with the Python code). '-5' is just an example exponential. This way, if the value of the coordinate is very small, the test will primarily be of the absolute difference, rather than the relative difference. I also thought that I had a test that integrating an orbit in C gave the same result as in Python for all potentials, but I cannot find it now...

jobovy:
Hi WIlma,

Have you made any progress on this? It would be good to wrap this up, because it's so close to being done.

In case this is helpful for the test suite, I started using a new coverage website that combines the output from the different travis runs (previously it was difficult to see whether a line was covered, because you had to check the results from each of 7 travis runs on coveralls; this new website combines them). This should make it easier to see where the 50 lines are that aren't covered by tests:

https://codecov.io/github/jobovy/galpy?branch=wilma

Best,

Jo

wilmatrick:
Hi Jo,

I'm really sorry, that I didn't finish this earlier, but I run into some problems again for which I could not find a solution quickly.

* I included your suggestion to compare not only the relative difference between the C and the Python integrated orbit, but also the absolute difference for coordinates close to zero. But it got only a tiny bit better... Could you check out if the test conditions are okay for you?
--> Starting on line 169 in test_orbitIntegrationC() in https://github.com/wilmatrick/galpy/blob/4d655c2a9dec670b5710f9c0efa580a163a37ce2/nose/test_kuzminkutuzov.py)

* I also changed the array treatment to something more elegant (I hoped). Please see:
https://github.com/wilmatrick/galpy/commit/50b5016f0e70ab48ccee3e3a66a3a40d711b6450 for the changes in KKSpot.py
https://github.com/wilmatrick/galpy/commit/eb826223ee0d4e9be6e09e7eaf84895f17b58354 for the changes in bovy_coords.py
This also solved the problem with the ridiculous long calculation time of the frequencies I wrote you about on May 8. Yay!

Unfortunately with this faster array treatment the orbit integration crashed, so I had to change it back to the slow array treatment. 
https://github.com/wilmatrick/galpy/commit/bb2ffde531689fd7a0a14e939ff88a83c445fbb5

The error message is:

nosetests -v nose/test_kuzminkutuzov.py -x --nocapture

test_kuzminkutuzov.test_vcirc ... ok
test_kuzminkutuzov.test_density ... ok
test_kuzminkutuzov.test_orbitIntegrationC ... ERROR

======================================================================
ERROR: test_kuzminkutuzov.test_orbitIntegrationC
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/trick/BovyCode/galpy/nose/test_kuzminkutuzov.py", line 166, in test_orbitIntegrationC
    o_P.integrate(ts,pot,method='leapfrog')  #python
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/Orbit.py", line 294, in integrate
    self._orb.integrate(t,pot,method=method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 76, in integrate
    self.orbit= _integrateRZOrbit(self.vxvv,pot,t,method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 483, in _integrateRZOrbit
    tmp_out= _integrateFullOrbit(this_vxvv,pot,t,method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py", line 577, in _integrateFullOrbit
    t,args=(pot,),rtol=10.**-8)
  File "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py", line 60, in leapfrog
    dt= _leapfrog_estimate_step(func,qo,po,dt,t[0],args,rtol,atol)
  File "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py", line 104, in _leapfrog_estimate_step
    force= func(qtmp,*args,t=to+3.*dt/4)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py", line 682, in _rectForce
    if x[1] < 0.: phi= 2.*nu.pi-phi
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

----------------------------------------------------------------------

Do you have an idea how to solve this problem?

Thanks!
Wilma

wilmatrick:
Hi Jo,

I'm really sorry, that I didn't finish this earlier, but I run into some problems again for which I could not find a solution quickly.

* I included your suggestion to compare not only the relative difference between the C and the Python integrated orbit, but also the absolute difference for coordinates close to zero. But it got only a tiny bit better... Could you check out if the test conditions are okay for you?
--> Starting on line 169 in test_orbitIntegrationC() in https://github.com/wilmatrick/galpy/blob/4d655c2a9dec670b5710f9c0efa580a163a37ce2/nose/test_kuzminkutuzov.py)

* I also changed the array treatment to something more elegant (I hoped). Please see:
https://github.com/wilmatrick/galpy/commit/50b5016f0e70ab48ccee3e3a66a3a40d711b6450 for the changes in KKSpot.py
https://github.com/wilmatrick/galpy/commit/eb826223ee0d4e9be6e09e7eaf84895f17b58354 for the changes in bovy_coords.py
This also solved the problem with the ridiculous long calculation time of the frequencies I wrote you about on May 8. Yay!

Unfortunately with this faster array treatment the orbit integration crashed, so I had to change it back to the slow array treatment. 
https://github.com/wilmatrick/galpy/commit/bb2ffde531689fd7a0a14e939ff88a83c445fbb5

The error message is:

nosetests -v nose/test_kuzminkutuzov.py -x --nocapture

test_kuzminkutuzov.test_vcirc ... ok
test_kuzminkutuzov.test_density ... ok
test_kuzminkutuzov.test_orbitIntegrationC ... ERROR

======================================================================
ERROR: test_kuzminkutuzov.test_orbitIntegrationC
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/trick/BovyCode/galpy/nose/test_kuzminkutuzov.py", line 166, in test_orbitIntegrationC
    o_P.integrate(ts,pot,method='leapfrog')  #python
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/Orbit.py", line 294, in integrate
    self._orb.integrate(t,pot,method=method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 76, in integrate
    self.orbit= _integrateRZOrbit(self.vxvv,pot,t,method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 483, in _integrateRZOrbit
    tmp_out= _integrateFullOrbit(this_vxvv,pot,t,method)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py", line 577, in _integrateFullOrbit
    t,args=(pot,),rtol=10.**-8)
  File "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py", line 60, in leapfrog
    dt= _leapfrog_estimate_step(func,qo,po,dt,t[0],args,rtol,atol)
  File "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py", line 104, in _leapfrog_estimate_step
    force= func(qtmp,*args,t=to+3.*dt/4)
  File "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py", line 682, in _rectForce
    if x[1] < 0.: phi= 2.*nu.pi-phi
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

----------------------------------------------------------------------

Do you have an idea how to solve this problem?

Thanks!
Wilma

wilmatrick:
(I'm so sorry, I did not want to close this. Still a bloody github beginner...)

jobovy:
On Tue, Jun 9, 2015 at 4:15 AM, Wilma Trick <notifications@github.com>
wrote:

> Hi Jo,
>
> I'm really sorry, that I didn't finish this earlier, but I run into some
> problems again for which I could not find a solution quickly.
>
>    -
>
>    I included your suggestion to compare not only the relative difference
>    between the C and the Python integrated orbit, but also the absolute
>    difference for coordinates close to zero. But it got only a tiny bit
>    better... Could you check out if the test conditions are okay for you?
>    --> Starting on line 169 in test_orbitIntegrationC() in
>    https://github.com/wilmatrick/galpy/blob/4d655c2a9dec670b5710f9c0efa580a163a37ce2/nose/test_kuzminkutuzov.py
>    )
>    -
>
>    I also changed the array treatment to something more elegant (I
>    hoped). Please see:
>    wilmatrick@50b5016
>    <https://github.com/wilmatrick/galpy/commit/50b5016f0e70ab48ccee3e3a66a3a40d711b6450>
>    for the changes in KKSpot.py
>    wilmatrick@eb82622
>    <https://github.com/wilmatrick/galpy/commit/eb826223ee0d4e9be6e09e7eaf84895f17b58354>
>    for the changes in bovy_coords.py
>    This also solved the problem with the ridiculous long calculation time
>    of the frequencies I wrote you about on May 8. Yay!
>
> Unfortunately with this faster array treatment the orbit integration
> crashed, so I had to change it back to the slow array treatment.
> wilmatrick@bb2ffde
> <https://github.com/wilmatrick/galpy/commit/bb2ffde531689fd7a0a14e939ff88a83c445fbb5>
>
> The error message is:
>
> nosetests -v nose/test_kuzminkutuzov.py -x --nocapture
>
> test_kuzminkutuzov.test_vcirc ... ok
> test_kuzminkutuzov.test_density ... ok
> test_kuzminkutuzov.test_orbitIntegrationC ... ERROR
>  ERROR: test_kuzminkutuzov.test_orbitIntegrationC
>
> Traceback (most recent call last):
> File "/usr/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
> self.test(
>
>
>
>
>
>
>
>
>
> *self.arg) File "/home/trick/BovyCode/galpy/nose/test_kuzminkutuzov.py",
> line 166, in test_orbitIntegrationC o_P.integrate(ts,pot,method='leapfrog')
> #python File "/home/trick/BovyCode/galpy/galpy/orbit_src/Orbit.py", line
> 294, in integrate self._orb.integrate(t,pot,method=method) File
> "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 76, in
> integrate self.orbit= _integrateRZOrbit(self.vxvv,pot,t,method) File
> "/home/trick/BovyCode/galpy/galpy/orbit_src/RZOrbit.py", line 483, in
> _integrateRZOrbit tmp_out= _integrateFullOrbit(this_vxvv,pot,t,method) File
> "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py", line 577, in
> _integrateFullOrbit t,args=(pot,),rtol=10.*
>
>
>
> *-8) File "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py",
> line 60, in leapfrog dt=
> _leapfrog_estimate_step(func,qo,po,dt,t[0],args,rtol,atol) File
> "/home/trick/BovyCode/galpy/galpy/util/bovy_symplecticode.py", line 104, in
> _leapfrog_estimate_step force= func(qtmp,*args,t=to+3.
>
> *dt/4) File "/home/trick/BovyCode/galpy/galpy/orbit_src/FullOrbit.py",
> line 682, in _rectForce if x[1] < 0.: phi= 2.*nu.pi-phi
> ValueError: The truth value of an array with more than one element is
> ambiguous. Use a.any() or a.all()
> ------------------------------
>
> Do you have an idea how to solve this problem?
>

Hi Wilma,

Okay, I will take a look at the C/Python issue. The error above is because
somehow an array is returned for the force in the leapfrog Python
integration. I'll take a look at your new implementation of the forces for
arrays as well.

Hope you had a good vacation!

Jo

jobovy:
Hi Wilma,

The comparison between the Python and C orbit integration works fine for me, and it also passes on travis:

https://travis-ci.org/jobovy/galpy/jobs/66011189#L2106

So this is probably okay.

Jo

jobovy:
Hi Wilma,

I think the issue with the fast array treatment is that the force in that treatment always returns an array, which makes the leapfrog integration crash. You could make sure to return a number rather than an array when dim == 1. So for example here (https://github.com/wilmatrick/galpy/commit/eb826223ee0d4e9be6e09e7eaf84895f17b58354#diff-51311b7d1b561ad8704e15c495f47cc0R1698) you could write:

if dim == 1: return jac[:,:,0]
else: return jac

All of the changes here (https://github.com/wilmatrick/galpy/commit/50b5016f0e70ab48ccee3e3a66a3a40d711b6450) should also be unnecessary, since Python will just assume [0,0] == [0,0,:] etc.

That said, I still don't understand why you need the array input for the actions/frequencies, since those methods do not use the Python code! Most of the other potentials in galpy do not accept array inputs and work fine.

Jo

wilmatrick:
Hi Jo,

did you also look at the test conditions in the orbit test? I had to tweak
them somehow to make the test work fine.

E.g. for the vz coordinate the test condition is

rel_diff = |(Python-CC)/CC)| < 0.01
abs_diff = |(Python-CC)|     < 0.01   *   |Python| < 0.01
test_condition = rel_diff+abs_diff

If I choose something smaller than 0.01 it crashes again.

I compared the C with the python code of the KKSPot again, and both
implement exactly the same formula. So I'm really surprised, that there
are differences at all.

Do you still think it's okay?

Cheers,
Wilma

wilmatrick:
Hi Jo,

thanks so much - now everything works fine with the array treatment.

I didn't know that Python indeed assumes [0,0] == [0,0,:]. I was always
dealing with this by hand.
But now the KKS pot can deal with arrays even if it is not explicitely in
the code.

It's really good to have someone to also discuss tiny (galpy-related)
coding problems with!!! :-) Thanks for taking the time.

If you're still happy with the orbit integration, I can finally pronounce
the KKS Potential FINISHED! Yay!

Cheers,
Wilma

P.S.: On your comment:

> That said, I still don't understand why you need the array input for the
> actions/frequencies, since those methods do not use the Python code! Most
> of the other potentials in galpy do not accept array inputs and work fine.

Hm, maybe I'm doing something wrong in my "RoadMapping" code? I'm
pre-calculating the frequencies together with the actions for each
potential. I do this by calling (just for illustration):

value,actions,frequencies =
qdf(R,vR,vT,z,vz,log=True,_return_actions=True,_return_freqs=True)

And I'm pretty sure, that this uses indeed the python KKSpot code to
calculate the frequencies. Therefore I needed the array input.

The potential I use in my paper is a 2-component KKSpot, i.e. array of
potentials:
pot = [KKSpot_disk,KKSpot_halo]
Is there a possibility to tell this potential, that it should use C-code
instead of Python code to evaluate actions and frequencies?

jobovy:
Hi Wilma,

Looks good! I will take another look at the C/Python test, I didn't realize that the condition for success was so weak.

On the frequencies: I'm sorry, I got confused between the orbital frequencies (\Omega), which are computed in C, and the epicycle frequencies, which are indeed computed in Python. So never mind my comment!

Jo

PR #238, opened 2015-05-28
Codecov coverage reporting

stevepeak:
Hey there, I found your repo and saw some complex methods to generate coverage reports. I wanted to see how Codecov (I'm the founder of Codecov) would handle the lcov and cobertura together. It merges them automatically so no need for `coveralls-merge`

If your interested in switching to Codecov that would be great. You can see the report from [my build here](https://codecov.io/github/stevepeak/galpy?ref=cc537e91ecddff3794f24e9097242100f14eb2f7)

Cheer!

Steve

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,6 @@ before_install:
 # command to install dependencies
 install:
  - pip install nose-cov
- - pip install coveralls-merge
  - gem install coveralls-lcov
  - easy_install --upgrade requests
  - easy_install --upgrade coveralls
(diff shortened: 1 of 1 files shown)


jobovy:
Hi Steve,

Thanks for the pull request. I'm very interested in using Codecov, which looks great and much better than coveralls. For now I'd like to keep the uploads to coveralls though, which require the merge between the C and Python reports. Your change also does not upload both the Python and C coverage reports (only the C report is uploaded to Codecov); I believe I first have to do

codecov

and then

bash <(curl -s https://codecov.io/bash) -v

in the after_success section.

I had actually already fixed this in my dev-v1.1 branch (which is the development branch):

https://github.com/jobovy/galpy/blob/dev-v1.1/.travis.yml


One thing here that might be useful for you is that the Codecov upload (codecov and the bash ... invocation) needs to be done **before** the coveralls-merge and coveralls upload. This might be useful for helping other people who want to use both coveralls and Codecov for now.

Thanks for the great work on Codecov, I'm sure this will be very useful for me!

Jo

stevepeak:
@jobovy thanks for the feedback and help getting this working.

Now that you mention the lack of Python report I know why. It's because the report was not generated by `coverage xml` which is called in the Bash uploader, but [only when no reports were found](https://github.com/codecov/codecov-bash/blob/master/codecov#L302-L323).

I'll make that adjustment to the uploader to always try to generate a coverage report for python.

Anything you need please do not hesitate to ask. 

Cheers,
Steve

PR #241, opened 2015-06-10
Finalize adding Kuzmin-Kutuzov potential

jobovy:
@wilmatrick

I added a few more tests and tweaked the orbit-integration in Python and C test (which seems fine to me, the differences in z and vz are just somewhat bigger than those in R,vR, and vT because the values are typically smaller). I will merge this into the dev-v1.1 branch and then soon merge everything into master and release a new galpy version.

Thanks!

Jo


wilmatrick:
Wow! You added a LOT of tests. Thanks for this.
I'm glad, we finally have finished it. :-)

jobovy:
I only added two more tests, the commits listed are mostly yours :-)!

Thanks,

Jo

PR #250, opened 2015-08-16
Hernquist analytic kicks with tests

jls713:


jobovy:
That looks great, thanks for the tests!

PR #251, opened 2015-08-18
Kicks computed with full orbit path of stars

jls713:
I have written two more impulse_deltav* routines. One uses pre-computed samples of an orbit at time samples, and the second uses a galpy potential to compute the orbits. This second one could certainly be improved by:
1. Simplify the interface to galpy.orbit?
2. Don't have the user choose the time samples but instead integrate (both forwards and backwards in time) until the acceleration from the sub-halo falls below some threshold and automatically choose a suitable resolution for the time sampling.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,7 @@ install:
  - gem install coveralls-lcov
  - easy_install --upgrade requests
  - easy_install --upgrade coveralls
- - pip install codecov
- - if $REQUIRES_PYNBODY; then pip install git+git://github.com/pynbody/pynbody.git@cda64c8ac9fa59b83663b155d14d184eb33826a5; fi
+ - if $REQUIRES_PYNBODY; then pip install git+git://github.com/pynbody/pynbody.git; fi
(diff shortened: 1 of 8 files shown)


jls713:
There is a bug here. I didn't realise that times=0 wasn't the initially passed point for orbit so the particles are only interacted forward in time. Not sure why the tests pass though...

jobovy:
Yes, the initial condition is at times[0], not where times == 0. I have some code in ``galpy/actionAngle_src/actionAngleIsochroneApprox.py`` that performs both forward and backward integration (it's what I use in the generating-function code). One would have to hack it to use it here (because it's a member function) or re-write it. I'll try to take a look at this code after Wayne's defense this afternoon...

jobovy:
Should I merge this now? Can't quite tell whether this is done yet...

jls713:
The orbit integration is done. I am not quite sure whether the streamgapdf.impulse_deltav_general_orbitintegration is working correctly. It passes the test but in the comparison of the kicks plot (attached) the delta v_x kicks are quite different from the other calculations and it is not clear to me why. 

image

PR #270, opened 2016-04-18
V t int limit

wilmatrick:
I would like to merge my changes in quasiisothermaldf.py with galpy master. The only changes are that one can choose the integration limits over the velocity components now by hand when calculating the tracer density. Default is nsigma=4. and vTmax=1.5. There are special cases where you need larger integration ranges, for example if one needs really high accuracy for the integral, or in regions far away from the sun. I think it is a useful change.


jobovy:
Thanks, this is great and we will want to merge this into ``master``. However, as you see, one of the tests has failed and this is one that tests that results shown in the ``galpy`` paper still work, so we want to figure out what's going on with that: https://travis-ci.org/jobovy/galpy/jobs/123888814#L2209

It seems to me that this is related to the way you've dealt with the normalization. I see a bunch of comments in the code asking about the normalization (e.g., [here](https://github.com/jobovy/galpy/pull/270/files#diff-4d26d7c9a34b2dddf5055719755d9e1bR1436)). Why did you not keep the normalization factor the same? Do you think the previous normalization factor was wrong? If not, could you change the code such that the normalization for the default parameters is the same as before in all functions? Thanks!

Note that you should be able to run the test that failed locally using:
```
nosetests -v nose/test_galpypaper.py -x --nocapture
```
such that you can test whether it still fails before committing or pushing to GitHub. If you're running it more than a few times, you probably want to do
```
nosetests -v nose/test_galpypaper.py -x --nocapture -e diskdf -e oort
```
to not run the long test of ``diskdf`` every time.

jobovy:
Hi Wilma,

Is all you did in the latest commit merge your branch with master?

There's still a discrepancy between the old and new normalization: seems like it's off by a factor of two, see [here](https://travis-ci.org/jobovy/galpy/jobs/279494653#L3027). Note that the test suite now uses ``pytest``, which gives more useful error messages (now it's clear that the difference is a factor of two).

Do you think the factor of two is correct (and the old code was wrong)?

In case you want to see whether your code will pass the tests, you should be able to run
```
pytest -v tests/test_galpypaper.py -xs -k test_qdf
```

(this just runs the single test that currently fails).

wilmatrick:
Hi Jo,

sorry that I only now (that my thesis is done and I plan modifications in the galpy qDF) finally came back to this issue. As a reminder: I have changed the prefactors in the pvR, pvz, pvT etc. functions to properly account for the integration limits in the case of flexible integration ranges.
These prefactors do not matter much because these functions are most of the time only used as distributions for which the absolute value doesn't matter. Still, they should, ultimately, be correct.

First, I have problems running the nose test you suggested above:

`nosetests -v nose/test_galpypaper.py -x --nocapture`

I get the following error message:

```
Traceback (most recent call last):
  File "/usr/bin/nosetests", line 9, in <module>
    load_entry_point('nose==1.3.7', 'console_scripts', 'nosetests-3.4')()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 560, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2647, in load_entry_point
    raise ImportError("Entry point %r not found" % ((group, name),))
ImportError: Entry point ('console_scripts', 'nosetests-3.4') not found
```

Any idea what is missing?

Second, I did some tests you suggested some time ago: Comparing vmomentdensity with the integral over pvR etc. I have made some plots that demonstrate the following:
(1) Your implementation of pvR and pvz agrees with mine for the default values (nsigma=4), because the prefactors cancel each other out.
(2) I think you missed a factor of nsigma (even in the default case with nsigma=4) in pvT.

plot_testing_pv_prefactors_jo_4 0
plot_testing_pv_prefactors_wilma_4 0
plot_testing_pv_prefactors_jo_2 0
plot_testing_pv_prefactors_wilma_2 0

I think, it is pretty clear that, when we leave it to the user to decide on the integration ranges, we should include my changes in galpy.

Now I have to figure out why coveralls is not throwing errors...

wilmatrick:
I just saw that you commented while I was commenting too...

Do I need to install something special to be able to run pytest?

jobovy:
Hi Wilma,

You just have to install ``pytest`` to be able to use the test suite now: see [here](https://docs.pytest.org/en/latest/getting-started.html).

Okay, I think I agree with your normalizations (and I'm sure you've thought about them much more than me at this point 😄), so we should use those. The only issue is that that will change some of the examples given in the galpy paper (which were using the wrong normalization). So we will have to make a note of that in the ``test_galpypaper.py`` code. This way if anybody runs the code in the paper and is wondering why the result is different, they can find the reason in the test code.

So I think all you need to do to finish this PR is
* Remove the comments about checking the normalization :smile: in ``quasiisothermaldf``
* Edit the tests in ``tests/test_galpypaper.py`` that fail because you are now using the correct normalization. Just replace the numbers in the ``assert numpy.fabs(FUNCTION-number)...`` statements where the tests fail. Also make a note everytime you do this about why this is different from the galpy paper.
* Add an entry in ``HISTORY.txt`` with this PR's changes to the normalization.

Once you get past the first test that now fails, there might be others as well (not sure, there's a ``qdf.pvz`` but it seems like this will still be working now). Once the ``test_qdf`` passes, please run
```
pytest -v tests/test_galpypaper.py -xs
```
to make sure that all galpy paper tests run, before pushing to GitHub again and running the travis test suite.

wilmatrick:
Just for documentation reasons, here a final comment and test:

These are all the Gauss-Legendre prefactors that I added to quasiisothermaldf.py functions:
[nsigma^2/2] --> pvT(), 
[nsigma/2] --> pvRvT(), 
[nsigma/2] --> pvTvz(), 
[vTmax/2] --> pvRvz()

For the default values (nsigma=4, vTmax=1.5), this means that in the old version there were missing prefactors of 4, 2, 2 and 1.5/2, respectively. 

To demonstrate that the calculations are now overall consistent within quasiisothermaldf.py, I'm comparing vmomentdensity (n=0,m=0,o=0) at (R,z=1kpc) with the integrals of the pv... functions over the velocities. By adding the new prefactors, the functions give all the same result.
plot_testing_pv_prefactors_jo_4 0_1 5

jobovy:
Hi Wilma,

Thanks for this. It looks like this is now ready to be merged? If you confirm that it is, I will merge it. I think some tests fail because of some unrelated numpy/astropy issue (astropy/astropy#7051), but everything related to your change seems to work.

wilmatrick:
Hi Jo, yes, it is ready to merge. The test `pytest -v tests/test_galpypaper.py -xs` was throwing some warnings, but no errors. Also, I have commented and tested everything properly. Cheers, Wilma

jobovy:
Great, thanks!

PR #271, opened 2016-04-20
NFW's a is not in units of Ro when conc is None

JovanVeljanoski:


jobovy:
Thank you for your contribution! I know the documentation of parameters in terms of Ro or not is confusing, so thanks for adding the clarification!

PR #281, opened 2016-08-04
Scf c

SeaifanAladdin:
Updated tests, added documentations. Added axi symmetric optimization.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
  - hash -r
  - conda config --set always_yes yes --set changeps1 no
  - conda update -q conda
- - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib setuptools pip cython>=0.20 nose==1.3.4
+ - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy=0.17.0 matplotlib setuptools pip cython>=0.20 nose==1.3.4
  - source activate test-environment
  - export DISPLAY=:99.0 #For plotting tests, courtesy Ryan Varley
  - sh -e /etc/init.d/xvfb start $For plotting tests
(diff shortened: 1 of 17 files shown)


jobovy:
Could you merge ``merge-scf-twopowertri`` from the main galpy branch into your SCF_C branch first and then push those changes? Right now there is a conflict, which means that the tests don't run.

jobovy:
Thanks! The build now fails for what seems to be an issue with the latest version of scipy that I reported here: scipy/scipy#6458

I think we'll have to use a work-around until this is fixed. Could you replace the line in ``.travis.yml`` that reads
```
conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy matplotlib setuptools pip cython>=0.20 nose==1.3.4
```
with
```
conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION numpy scipy=0.17.0 matplotlib setuptools pip cython>=0.20 nose==1.3.4
```
? This way we'll be using the older version.

jobovy:
Thanks, this looks good. Something weird has happened to the test coverage reporting that I think is related to one of the software packages used in the test-coverage reporting being updated in a backward incompatible way :rage::rage::rage: I'll merge this and try to fix this.

jobovy:
Okay, that fixed it and the ``merge-scf-twopowertri`` branch now does test coverage correctly again; you might want to merge this last change into your own branch to get this in in case we make other changes.

The test coverage is here, looks great only ten lines now covered! https://codecov.io/gh/jobovy/galpy/tree/6037e50cfc1dbbbd74053ea6433ba30f947d6ad7/galpy
As far as I can tell, the C code cannot be touched (the lines are for the evaluation of the potential in the non-axisymmetric case, which currently does not get used anywhere). There's a few lines in the SCFPotential.py file that could probably be touched, but it's not a big deal. Great job!

The documentation is also online now:

http://galpy.readthedocs.io/en/merge-scf-twopowertri/reference/potential.html

Maybe check this quickly to make sure it looks the same as what you had, but it should.

PR #283, opened 2016-08-05
Increased test coverage for SCFPotential: axisymmetric case

SeaifanAladdin:
Added an extra test that should cover the three missed lines from 

[https://codecov.io/gh/jobovy/galpy/src/6037e50cfc1dbbbd74053ea6433ba30f947d6ad7/galpy/potential_src/SCFPotential.py](https://codecov.io/gh/jobovy/galpy/src/6037e50cfc1dbbbd74053ea6433ba30f947d6ad7/galpy/potential_src/SCFPotential.py)

Did not make any test to cover that last miss since roundoff isn't used(although it could in the future?).


jobovy:
Looks great, thanks!

PR #284, opened 2016-08-22
Updated documentation to SCFPotential

SeaifanAladdin:
Added references and fixed algebraic mistakes in the documentation.

Also added SCFPotential into HISTORY.txt


jobovy:
Looks great; I'll merge and then we should be able to see the documentation changes online.

PR #286, opened 2016-09-17
Add FerrersPotential

smoh:
This is to add a new potential FerrersPotential.
- When getting lower limit of the integrals, I could not figure out the real positive root from the closed form solution of cubic function, and used numpy.roots.
- This assumes the short axis to be in z-direction.
- The 2nd derivatives work for arbitrary time, and from the comparison of densities from formula and Poission equation (forcePoission option), I think they are correct.
- Since there's finite total mass to the corresponding density profile, I added an option in `__init__` to set `amp` from total mass.


jobovy:
Hi,

Thanks for the pull request! One quick first comment: should we just make ``amp`` always be the total mass because it is finite? That would remove the ``totalmass=True`` option and be perhaps less ambiguous. Would be a rather small change to the code and documentation.

Could you also add a file to ``doc/source/reference/`` for the potential (similar to other ``potential*.rst`` files in that directory) and add the potential to the ``Triaxial potentials`` section of ``doc/source/reference/potential.rst?

Finally, the tests on this and #287 failed to run because there's an issue with Anaconda's matplotlib version (see ContinuumIO/anaconda-issues#1068). Could you merge the fix that is in master in both this and #287 before you next push changes? That way the test suite will run automatically.

Thanks!

smoh:
Hi Jo,

- I rebased all changes to the new master, and added files to doc/.
- I made amplitude total mass (and passed directly to `Potential`), and now other normalizations (of actual potential/force/2nd derivatives/density) are taken care of within the class.

The same two tests seem to fail for both python 2.7/3. Could you comment on what's wrong?

jobovy:
Thanks! One of the test that fails is that which test that the second derivative of the potential is correctly implemented (by comparing to numerical integration); this is in ``nose/test_potential.test_2ndDeriv_potential``. It seems like this is just because the tolerance is too small. You can fix this by adding a line just before the ``for p in pots`` loop in that function:
```
tol['FerrersPotential']= -4.
```
You can run the test locally as 
```
nosetests -v nose/test_potential.py -x --nocapture
```
to check that this worked.

The other tests that failed because energy conservation fails in ``nose/test_orbit.py`` in ``test_energy_jacobi_conservation``. You can also similarly add a line
```
tol['FerrersPotential']= -4.
```
to just before ``for p in pots`` to lower the tolerance (of course, you should be convinced that energy is conserved enough; it's probably rather badly conserved because of the numerical integration to get the forces).

``nose/test_orbit.test_liouville_planar`` has a similar problem. ``numpy.fabs(tjac-1.)`` in line 509 is above the default tolerance, so you need to again set it by hand.

The ``test_fixedstepsize`` fail is unrelated to your changes (it's just a bad test that often fails).

Try to run the ``test_potential.py`` and ``test_orbit.py`` tests locally before submitting, that way we know that the tests should fail.

Overall it's impressive that so many of the tests pass for this new potential (because a lot more tests are automatically run on FerrersPotential)!

smoh:
Hi Jo, 

I was able to resolve `nose/test_potential.test_2ndDeriv_potential` without increasing the tolerance by setting error tolerance in `scipy.integrate.quad`. I think doing this makes sense because the default for `quad` is ~1E-8, and the testing routines are doing differentiation by moving position on that order.

On tests in `test_orbit.py`, I have noticed the error message reads integration failure with `dopr54_c`, which is not possible because I didn't implement the potential in C. It turns out even for other potentials the error messages can be misleading because the integration routines changes integration method silently (exactly the other issue I raised).
I added some print statement to `test_orbit.py` and in `_integrateFullOrbit` function to demonstrate this [here](https://gist.github.com/smoh/ee8ac2e2fd5f57e47ea89b0c997c4741)

The stuff I added has `>>>` prepended, and I grepped them in a separate file.
You can see that the first time the integration method `test_orbit.py` thinks it uses and the actualy method that's used in the integration routine differs is 
line 117-119 
```
>>> potential = RazorThinExponentialDiskPotential
>>> integrator = dopr54_c ttol = -10.0
>>> _integrateFullOrbit method =  odeint
```
and it also happens for `TwoPowerTriaxialPotential` as well.

This matters in `test_orbit.py` because you set the times for integration depending on the integration method

```python
    times= numpy.linspace(0.,280.,10001) #~10 Gyr at the Solar circle
    fasttimes= numpy.linspace(0.,28.,1001) #~1 Gyr at the Solar circle
    integrators= ['dopr54_c', #first, because we do it for all potentials
                  'odeint', #direct python solver
                  'leapfrog','leapfrog_c',
                  'rk4_c','rk6_c',
                  'symplec4_c','symplec6_c']
...

        for integrator in integrators:
            print('>>> integrator =', integrator, 'ttol =', ttol)
            if integrator == 'dopr54_c' \
                    and not 'MovingObject' in p: ttimes= times
            else: ttimes= fasttimes
```

I have a comment and two questions:

* I think it'll be best to set the default method in all integration routines to odeint (or something else globally available), and raise error if user selected method is unavailable. If you agree, I will do that when I get to the other pull request.

* `FerrersPotential` would have passed if `fasttimes` was selected (integration for 1 Gyr)

image

But when using `odeint` and integrating for 10 Gyrs, the energy has sudden jumps

image

Is this expected from numerical degradation or do I have a problem in the potential I wrote?
If latter, do you know where I should be looking to find the problem?

* In test, for each potential you loop over integrators, yet from the output, only one integrator seems to be tested except for special cases -- where can I see this in the code?

Thanks for your help.

jobovy:
Hi Semyeong,

Thanks for continuing to look into this!

Yes, I know that the tests for potentials not implemented in C will simply fall back onto ``odeint``. Because most users will want to take advantage of the C integrators without having to explicitly select them, I prefer to keep a C integrator as the default. But I very much like your other pull request that raises a warning when the integrator falls back onto a Python version in the absence of a C implementation, so I think we should include that. Perhaps we should discuss there whether to also raise an Error when a user explicitly requests an integrator that is unavailable, but I think a warning like what you implemented would be a good start.

In ``test_orbit.py``, the integration time is not so much set based on the integrator. It's just set to a shorter integration for all but one of the integrators (which happens to be ``dopr54_c`` for most potentials). Ideally, we want all potentials to conserve energy well enough over 10 Gyr even when using ``odeint``. 

I think your test must be failing because the orbit gets to a part of phase-space where the force is not quite the same as the derivative of the potential because of numerical issues with the integration in FerrersPotential. Have you looked at the orbit and what happens at those times?

Since the energy is well conserved except for the spikes, I think we can be quite sure that the forces are implemented correctly and that the potential is also good enough (from the fact that there is a spike in the energy, rather than a lasting change, the problem is probably in the potential and not the forces). So what we could do in this case is to use ``fasttimes`` for the FerrersPotential, because then the test would pass. You could do that by editing line

https://github.com/jobovy/galpy/blob/master/nose/test_orbit.py#L140

to also have ``and if not 'Ferrers' in p``. Other functions might also need that then.

Yes, only for one or two potentials (I forget) are all of the integrators tested (NFW and maybe another one), simply not to waste time testing all integrators for all potentials. This is done by breaking out of the integrator loop for the other potentials, for example, in this line:

https://github.com/jobovy/galpy/blob/master/nose/test_orbit.py#L360

(and there are other similar lines higher up for 2D and 1D potentials).

smoh:
Thanks for the quick response! I've looked at the orbit (and also compared to `TwoPowerTriaxial` case) yet didn't find anything odd, but I will investigate what you said around the spikes and get back to you. If it becomes certain there isn't much I can improve upon, I will add statements you suggested so that it can pass the test.

smoh:
Hi Jo, just so you know I haven't abandoned this, it seems that this only happens because the test orbit is radial, and gets very far out, and I think things are sound with potential/force implementation. I intend to make tweaks you suggested so that all tests pass soon, and code ready to be merged.

jobovy:
Sounds good. I've arrived at the CCA now, so if you're here sometime and you want to work the last parts out together, that might be easiest.

jobovy:
Hi Semyeong,

I was wondering what the status of this PR is? From what I remember, it was pretty close to done and it would be a shame not to merge this in. So it would be great if you could let me know if it's ready to be merged.

Note that I added a very simple bar potential in galpy recently: [SoftenedNeedleBarPotential](http://galpy.readthedocs.io/en/latest/reference/potentialsoftenedneedle.html). I also started a section of the ``doc/source/reference/potential.rst`` documentation for bar potentials, which is probably where the FerrersPotential docs should be moved as well, but I can do that later.

Finally, before you finish the PR, please add the FerrersPotential to the ``HISTORY.txt`` file and add yourself and the student you were working with to the ``AUTHORS.txt`` file.

smoh:
Hi Jo,

I think all tests should pass now (I checked locally), and did what you said about author/history. I think it's ready to be merged.

jobovy:
Hi Semyeong,

This all looks good; I will merge it in. Thanks so much for adding this!

PR #287, opened 2016-09-17
issue warning when integration method changes

smoh:
I noticed that the integration routines silently falls back to python methods when C methods are selected, and thought it'd be better to issue a warning of the change explicitly.

I think there are couple of other integration routines for e.g., planar orbits, but I submit this example to see if you think this is a good idea. (Although right now, the methods in classes default to C, and you are warned when C integrations are actually used..)

### DIFF
--- a/nose/test_orbit.py
+++ b/nose/test_orbit.py
@@ -3023,6 +3023,27 @@ def test_orbit_c_sigint_planardxdv():
             raise AssertionError("Full orbit integration using %s should have been interrupted by SIGINT (CTRL-C), but was not because p.poll() == %i" % (integrator,p.poll()))
     return None
 
+def test_orbitint_pythonfallback():
+    # Check if a warning is raised when the potential has no C integrator
+    from galpy.orbit import Orbit
+    bp= BurkertPotentialNoC() # BurkertPotentialNoC is already imported at the top of test_orbit.py
+    bp.normalize(1.)
+    ts= numpy.linspace(0.,1.,101)
+    import warnings
+    for orb in [Orbit([1.,0.1,1.1,0.1,0.,1.]),Orbit([1.,0.1,1.1,0.1,0.]),
+                Orbit([1.,0.1,1.1,1.]),Orbit([1.,0.1,1.1])]:
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always", galpyWarning)
+            #Test w/ dopr54_c
+            orb.integrate(ts,bp, method='dopr54_c')
+            # Should raise warning bc of python fallback, might raise others
+            raisedWarning= False
+            for wa in w:
+                raisedWarning= ( "Cannot use C integration because some of the potentials are not implemented in C" in str(wa.message) )
+                if raisedWarning: break
+            assert raisedWarning, "Orbit integration did not raise fallback warning"
+    return None
+
 def test_linear_plotting():
     from galpy.orbit import Orbit
     from galpy.potential_src.verticalPotential import RZToverticalPotential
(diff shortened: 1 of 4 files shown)


jobovy:
Thanks, much appreciated! Yes, I think this kind of warning would be a good idea, so if you could go ahead and also implement it for the other integration methods (from RZOrbit, planarOrbit, planarROrbit), that would be great. One question though that I can't immediately figure out the answer to from looking at the diff: Does the warning only get raised when you are trying to use a *_c method? It seems like it might get raised when you are simply using 'odeint' or 'leapfrog', although I might be reading the code wrong.

Ideally, you would also add a test to ``nose/test_orbit.py`` to make sure these warnings are working as expected. For this test you could use ``BurkertPotentialNoC`` in ``nose/test_potential.py`` (to get a potential for which there is no C implementation, or you could use the Ferrers one I guess, but that one isn't in galpy yet). This test would go something like:
```
def test_orbitint_pythonfallback():
    bp= BurkertPotentialNoC(normalize=1.) # BurkertPotentialNoC is already imported at the top of test_orbit.py
    orb= Orbit([1.,0.1,1.1,0.1,0.,1.])
    ts= numpy.linspace(0.,1.,101)
    import warnings
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always",galpyWarning)
        #Test w/ dopr54_c
        orb.integrate(ts,bp)
       # Should raise warning bc of python fallback, might raise others
        raisedWarning= False
        for wa in w:
            raisedWarning= (str(wa.message) == "Cannot use C integration because some of the potentials are not implemented in C (using odeint instead)")
            if raisedWarning: break
        assert raisedWarning, "Orbit integration did not raise fallback warning"
```
and similar for other integration methods and orbit types (could easily run this in a loop). If you put it at the top of ``nose/test_orbit.py`` at first, it will be the first test to be run when you do:
```
nosetests -v nose/test_orbit.py -x --nocapture
```
so you can immediately verify whether it works or not and move it to the back later (or leave it at the top for now, doesn't really matter).

The travis builds are having some weird issue with Anacaonda right now, so the test suite is not running properly for either this pull request or your previous one, but I will try to re-run it later when this is sorted out and comment on the ``FerrersPotential`` one.

smoh:
I've added it for all other integration functions, and test using `BurkertPotentialNoC` at the bottom of `test_orbit.py`. (Oddly, `BurkertPotentialNoC` complained not to have `normalize` keyword.)

Let me know if there's test potentials for `RZOrbit`, `planarOrbit`, or `planarROrbit`, but otherwise I think it's good to be merged.

nose/test_orbit.py
```diff
@@ -3554,3 +3554,20 @@ def check_integrate_t_asQuantity_warning(o,funcName):
         assert raisedWarning, "Orbit method %s wit unitless time after integrating with unitful time should have thrown a warning, but didn't" % funcName
     return None  
 
+def test_orbitint_pythonfallback():
+    # Check if a warning is raised when the potential has no C integrator
+    from galpy.orbit import Orbit
+    bp= BurkertPotentialNoC() # BurkertPotentialNoC is already imported at the top of test_orbit.py
+    orb= Orbit([1.,0.1,1.1,0.1,0.,1.])
```
jobovy: This looks great and is almost ready to merge, but it would be better to test all four new warnings.

I think you should be able to test all 4 new warnings by just looping through all of the possible orbits. That is, start a loop here with
```
for orb in [Orbit([1.,0.1,1.1,0.1,0.,1.]),Orbit([1.,0.1,1.1,0.1,0.]),Orbit([1.,0.1,1.1,1.]),Orbit([1.,0.1,1.1])]:
```
and then put everything in this function within the loop. Probably best practice to move the ``ts= `` and ``import warnings`` lines to before the loop.

Yes, ``BurkertPotentialNoC`` has no ``normalize`` keyword, but you can call
```
bp.normalize(1.)
```
after you initialize ``bp`` to do the normalization.

I can make these final changes as well if you allow me to edit the pull request (see https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/).

jobovy:
Looks good now, thanks @smoh!

PR #292, opened 2016-10-31
SCFPotential for v1 and v2 of GSL

SeaifanAladdin:
Replaced deprecated gsl_sf_legendre_Plm_array with gsl_sf_legendre_array_e for GSL v2.

Optimized axisymmetric case by using gsl_sf_legendre_Pl_array.

### DIFF
--- a/galpy/potential_src/potential_c_ext/SCFPotential.c
+++ b/galpy/potential_src/potential_c_ext/SCFPotential.c
@@ -349,10 +349,14 @@ inline void computeNonAxi(double a, int N, int L, int M,
     int Pindex = 0;
     for (l = 0; l < L; l++)
     {
-        
+        v1counter = 0;
         for (m = 0; m<=l; m++)
         {
-            v1counter = m*(m + 1)/2 + l; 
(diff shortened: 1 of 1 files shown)


jobovy:
Thanks! Everything seems to work for GSL v2, but the orbit integration test with mockSCFDensityPotential fails for GSL v1: https://travis-ci.org/jobovy/galpy/builds/172158579. I think there must be a problem with how the associated Legendre polynomials are computed in the code for GSL v1 (because the axisymmetric densities seem to work, so must be the m =/= 0 terms).

SeaifanAladdin:
I think the problem is because I started the for loop in the v1 case with `m=1` rather than `m=0`.
Changed it to be
 ```
for (m = 0; m < M; m++)
```

I don't have GSL v1 installed so I cant test it without using Travis

jobovy:
The tests all pass now, so it seems like everything is okay. Well done, thanks! I'm going to merge this into my gslv2 branch, test it a little further under GSL v1 locally, and then merge it into master.

PR #293, opened 2016-11-18
Updated the order of function parameters

lambainsaan:
In v1.2 the order of parameters of the function evaluatezforce changed from (R,z,Pot) to (Pot,R,z)


jobovy:
Thanks for noticing and fixing!

PR #297, opened 2017-01-27
Wilkinson Evans gravitational potential

ignotur:
This is a halo potential for the second fitted potential in Irrgang et al. (2013) article.
I added both Python and C++ implementation. I programmed it approximately one year ago for galpy v 1.1. Today I tried to adapt changes for galpy v 1.2, still C++-Python connection might not work smoothly.

### DIFF
--- a/galpy/orbit_src/integrateFullOrbit.py
+++ b/galpy/orbit_src/integrateFullOrbit.py
@@ -182,13 +182,12 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
                 elif hztype == 'sech2':
                     pot_args.extend([1,hz.get('h',0.0375)])
 
-          elif isinstance(p,potential.WilkinsonEvansPotential):
+        elif isinstance(p,potential.WilkinsonEvansPotential):
 	    #print 'we are here - trying to use c potential'
-#	    warnings.warn('We are here 27', galpyWarning)
+	    #warnings.warn('We are here 27', galpyWarning)
+	    #print p._Mh,p._ah
             pot_type.append(27)
             pot_args.extend([p._Mh,p._ah])
-
-
 
 
     pot_type= nu.array(pot_type,dtype=nu.int32,order='C')
--- a/galpy/orbit_src/integrateFullOrbit.py
+++ b/galpy/orbit_src/integrateFullOrbit.py
@@ -181,6 +181,16 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
                     pot_args.extend([0,hz.get('h',0.0375)])
                 elif hztype == 'sech2':
                     pot_args.extend([1,hz.get('h',0.0375)])
+
+          elif isinstance(p,potential.WilkinsonEvansPotential):
+	    #print 'we are here - trying to use c potential'
+#	    warnings.warn('We are here 27', galpyWarning)
+            pot_type.append(27)
+            pot_args.extend([p._Mh,p._ah])
+
+
+
+
     pot_type= nu.array(pot_type,dtype=nu.int32,order='C')
     pot_args= nu.array(pot_args,dtype=nu.float64,order='C')
     return (npot,pot_type,pot_args)
--- a/galpy/orbit_src/orbit_c_ext/integrateFullOrbit.c
+++ b/galpy/orbit_src/orbit_c_ext/integrateFullOrbit.c
@@ -38,6 +38,7 @@ void parse_leapFuncArgs_Full(int npot,
     potentialArgs->i2dzforce= NULL;
     potentialArgs->accxzforce= NULL;
     potentialArgs->accyzforce= NULL;
+
     switch ( *pot_type++ ) {
     case 0: //LogarithmicHaloPotential, 2 arguments
       potentialArgs->Rforce= &LogarithmicHaloPotentialRforce;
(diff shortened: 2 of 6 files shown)


jobovy:
Hi Andrei,

The tests seem to fail for the C implementation of the new potential. In particular, energy conservation is not satisfied (test ``test_orbit.test_energy_jacobi_conservation``). I *think* that the issue is that you don't include the ``_amp`` parameter when you pass the potential to C. The tests use a normalized version of the potential that does not have ``amp=1``  [(this line](https://github.com/jobovy/galpy/pull/297/files#diff-63c6a8f18875fca26ce3a35b7be52907R53) in the code changes the amp). I think you could fix the issue by replacing [this line](https://github.com/jobovy/galpy/pull/297/files#diff-1bfc5551329e96d294921af6957beee1R190) with
```
pot_args.extend([p._Mh*p._amp,p._ah])
```

However, I would suggest you remove the ``Mh`` input to the new potential altogether and replace it by ``amp`` (this is more in the galpy spirit). This you can then pass to the ``Potential.__init__`` function in [this line](https://github.com/jobovy/galpy/pull/297/files#diff-63c6a8f18875fca26ce3a35b7be52907R47) as
```
Potential.__init__(self,amp=amp,ro=ro,vo=vo,amp_units='mass')
```
(``__init__`` should also have ``ro=None,vo=None`` keywords; writing it this way will allow people to input a mass as ``10 * astropy.units.Msun``). Then the C glue above would become
```
pot_args.extend([p._amp,p._ah])
```
In the Python implementation of the potential, you wouldn't include ``amp`` (because this is multiplied in by the general ``Potential`` functions), so you would just replace ``Mh`` by 1 in your code.

Another test that fails is ``test_liouville_planar``. I think you need to add a line like [this line](https://github.com/jobovy/galpy/blob/master/nose/test_orbit.py#L457) but for your potential to exclude it from this test (because the necessary functions are not implemented for this functionality).

The other tests that fail are probably related to the ``amp`` problem above.

Before pushing the code again, you can try whether the tests pass by running
```
nosetests -v nose/test_orbit.py -x --nocapture
```
Note that these can take a while to run, but this way you will know whether there is still a problem before pushing the code to GitHub. This only runs the orbit tests, but those are the ones that fail. If this is taking too long, I think you can also test the basic issue with a simple
```
from galpy.orbit import Orbit
from galpy.potential import WilkinsonEvansPotential
wp= WilkinsonEvansPotential(normalize=1.)
import numpy
ts= numpy.linspace(0.,100.,1001)
o= Orbit([1.,0.1,1.1,0.1,0.1,0.])
o.integrate(ts,wp)
o.plotE()
```
and the energy should be close to conserved here.

jobovy:
Hi Andrei,

I noticed that you had another paper where you use this galpy potential and was wondering whether you are still interested in contributing it to galpy? I think the current PR was close to ready to merge, but somehow never got finished. If it's not too much work (and I don't think it would be), it would be nice to add this model to galpy; that way other people can also easily use it and it would be easier for you to keep your version of galpy up-to-date.

There's been quite a few commits to galpy since you initiated this PR and in particular the C number for the potential (27) that you chose is now used by a different potential, so that has to be changed (these numbers are just assigned first come, first served).

At this point I think it would be best to add the potential to a clean branch (say, ``tfpotential``) started from the current master branch and then add the new potential in that branch (the files with the potential in Python and C and then just add the Python/C glue again). You can then start a new PR (referencing this one) and close this one.

Some suggestions:

* I normally avoid using people's names for a potential and this one might be better named the "TruncatedFlatRotationCurvePotential", which is what Wilkinson & Evans named it.
* As discussed above, the ``Mh`` parameter should just be the ``amp`` parameter which every galpy parameter has.
* It would be great to also implement the density, which is quite simple.

Note that the test suite now uses pytest, so to run tests locally, you have to install pytest and run the tests as
```
pytest -v tests/test_potential.py
```
and
```
pytest -v tests/test_orbit.py
```

If you allow me to push changes to this PR I can also help out with making additional changes.

ignotur:
Dear Jo,

I still want to implement it. Unfortunately, I was very busy with finishing
my Ph.D. last few months.
Not the thesis is submitted, so indeed I can return back to it. I will try
to make some changes into potential on this weekend.

Best,
Andrei

jobovy:
Okay, great! Let me know if there's anything I can do to help with this.

ignotur:
Dear Jo,

I am sorry again for the delay with the potential implementation.

I am in a middle of moving to another country to start the postdoc in
Israel. I will return to this potential as soon as possible.

Best regards,
Andrei Igoshev

jobovy:
Hi Andrei,

Not a problem, good luck with the move!

Jo

jobovy:
Model II from Irrgang et al. (2013) is now part of the new ``galpy.potential.mwpotentials`` module as ``Irrgang13II``. Rather than specifically implementing the Wilkinson Evans potential, I implemented it simply by computing its basis function expansion using the ``SCFPotential`` class, which works well enough.

I can keep this open if you think you will eventually contributed the actual Wilkinson Evans model, but with the ``Irrgang13II`` model now in ``galpy.potential.mwpotentials``, the urgency is probably low.

ignotur:
Thank you very much for implementing the Irrgang model!

>I can keep this open if you think you will eventually contributed the
actual Wilkinson Evans model, but with the Irrgang13II model now in
galpy.potential.mwpotentials, the urgency is probably low.

You can definitely close this now.

PR #302, opened 2017-05-05
Minor spelling mistake in documentation: shoenrich to schoenrich

JeffreySimpson:


jobovy:
Thanks for fixing! If you find any other mistakes in the documentation (which is much appreciated!), please add ``[ci skip]`` at the end of each commit message; that way the test suite doesn't run.

PR #305, opened 2017-06-28
Add SpiralArmsPotential

jackhong6:


jobovy:
Hi Jack,

This PR looks great, well done! I will leave some further comments on the code, but here's a few more steps before this can be merged:

- Add an entry to HISTORY.txt describing that we added SpiralArmsPotential and a brief description of what it is. Refer to this PR #305.
- Add yourself to AUTHORS.txt
- Add an entry to the API documentation. That is, add 'spiral' to the title of this [subsection](http://galpy.readthedocs.io/en/latest/reference/potential.html#triaxial-and-bar-potentials) of the documentation and add a file similar to the ones in that list: . The documentation is at ``doc/source/reference/potential.rst`` and you will have to add a file ``doc/source/reference/potentialspiralarms.rst`` that looks like ``doc/source/reference/potentialdehnenbar.rst``.
- The actual documentation that gets shown gets taken from the class definition in SpiralArmsPotential.py. You can generate the documentation locally after installing sphinx and running ``make html`` in the ``doc/`` directory. This will generate the documentation in a directory one directory up from the ``galpy`` repository called ``galpy-docs`` and you can look at it there to see whether it looks okay. Can you also include the definitions of ``Kn``, ``Bn``, etc. in the documentation, such that the documentation is self-contained? Maybe also mention the two ``Cs`` settings from the paper.
- We need to test that inputting parameters as quantities works. There's a bunch of tests in ``tests/test_quantity.py`` of this and you should add similar ones to those for ``DehnenBarPotential`` for ``SpiralArmsPotential``.
- All of the tests currently run for the default set of parameters (in particular, ``N=2``, ``Omega=0``, and ``Cs=[1]``. I think we want to add tests varying these to make sure everything works (I'm sure it does, but it's always good to check!). To do this, you can add specific SpiralArms potentials by subclassing SpiralArmsPotential and calling ``__init__`` with different values, like [this example](https://github.com/jobovy/galpy/blob/master/tests/test_potential.py#L2350). Then you add them to tests like [this line](https://github.com/jobovy/galpy/blob/master/tests/test_potential.py#L100). Let's do a variation with ``Omega=1.`` and then also (``Omega=1.,N=4)`` and ``(Omega=1.,Cs=the other option from the paper). Also add the ``Omega=1.`` test to ``test_orbit.test_energy_jacobi_conservation``.

Thanks!

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+
+double D(double R, double H, double n, double N, double sin_alpha);
+
+double dK_dR(double R, double n, double N, double sin_alpha);
+
+double dB_dR(double R, double H, double n, double N, double sin_alpha);
+
+double dD_dR(double R, double H, double n, double N, double sin_alpha);
+
+double SpiralArmsPotentialEval(double R, double z, double phi, double t,
```
jobovy: It's great that you've added this function, but currently it will not get used by ``galpy`` (because evaluating a potential in C is only done for action-angle calculations, which we don't support for complicated potentials such as this one). You can exclude this code black from the report that checks whether the line is covered by the tests by adding a line before this definition that says
```
//LCOV_EXCL_START
```
and another line after it that I will list below.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+        double Kn = K(R, n, N, sin_alpha);
+        double Bn = B(R, H, n, N, sin_alpha);
+        double Dn = D(R, H, n, N, sin_alpha);
+
+        sum += Cn / Kn / Dn * cos(n * g) / pow(cosh(Kn * z / Bn), Bn);
+    }
+
+    return -amp * H * exp(-(R - r_ref) / Rs) * sum;
+}
+
```
jobovy: Add
```
//LCOV_EXCL_STOP
```

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+        double Bn = B(R, H, n, N, sin_alpha);
+        double Dn = D(R, H, n, N, sin_alpha);
+
+        sum += N * n * Cn / Dn / Kn / pow(cosh(z * Kn / Bn), Bn) * sin(n * g);
+    }
+
+    return -amp * H * exp(-(R - r_ref) / Rs) * sum;
+}
+
+double SpiralArmsPotentialR2deriv(double R, double z, double phi, double t,
```
jobovy: Again, great that you added this function, because it may becomes useful in the future, but currently galpy does not use the general forms of the second derivatives in C (only the planar ones), so add
```
//LCOV_EXCL_START
```
 before this function.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+                  + sin_ng * (z * tanh(zKB) * (dKn_dR / Kn - dBn_dR / Bn)
+                              + 1 / Kn * (-dBn_dR * log(sechzKB)
+                                          + dKn_dR / Kn
+                                          + dDn_dR / Dn
+                                          + 1 / Rs)));
+    }
+
+    return -amp * H * exp(-(R - r_ref) / Rs) * sum;
+}
+
```
jobovy: Add
```
//LCOV_EXCL_STOP
```

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+    phi = phi - omega * t;
+
+    double g = gam(R, phi, N, phi_ref, r_ref, tan_alpha);
+
+    // Return value of the potential.
+    double sum = 0;
+
+    int n;
+    for (n = 1; n <= nCs; n++) {
+        double Cn = *args++;
```
jobovy: It's best practice to declare variables *outside* of the loop, so move these ``double`` declarations outside of the loop.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+
+    double g = gam(R, phi, N, phi_ref, r_ref, tan_alpha);
+    double dg_dR = dgam_dR(R, N, tan_alpha);
+
+    // Return the Rforce (-dPhi / dR)
+    double sum = 0;
+
+    int n;
+    for (n = 1; n <= nCs; n++) {
+        double Cn = *args++;
```
jobovy: Same here: move the declarations outside of the loop.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+    phi = phi - omega * t;
+
+    double g = gam(R, phi, N, phi_ref, r_ref, tan_alpha);
+
+    // Return the zforce (-dPhi / dz)
+    double sum = 0;
+    int n;
+
+    for (n = 1; n <= nCs; n++) {
+        double Cn = *args++;
```
jobovy: Again move declarations outside of the loop

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+    double dg_dR = dgam_dR(R, N, tan_alpha);
+    double d2g_dR2 = N / R / R / tan_alpha;
+
+    // Return the second derivative of the potential wrt R (d2Phi / dR2)
+    double sum = 0;
+
+    int n;
+
+    for (n = 1; n <= nCs; n++) {
+        double Cn = *args++;
```
jobovy: For future use, best to move these declarations outside of the loop.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -0,0 +1,763 @@
...
+    double N = *args++;
+    double sin_alpha = *args++;
+    double tan_alpha = *args++;
+    double r_ref = *args++;
+    double phi_ref = *args++;
+    double Rs = *args++;
+    double H = *args++;
+    double omega = *args++;
+
+    phi = phi - omega * t;
```
jobovy: Can't we just pass ``phi-omega*t`` to the g function, rather than assigning ``phi=phi-omega*t``? I think ``phi`` is only used once.

galpy/orbit_src/orbit_c_ext/integrateFullOrbit.c
```diff
@@ -231,7 +231,18 @@ void parse_leapFuncArgs_Full(int npot,
...
       potentialArgs->phiforce= &ZeroForce;
       potentialArgs->nargs= (int) *(pot_args) + 3;
-      break;      
+      break;
+    case 27: // SpiralArmsPotential, 10 arguments + array of Cs
+      potentialArgs->Rforce = &SpiralArmsPotentialRforce;
+      potentialArgs->zforce = &SpiralArmsPotentialzforce;
+      potentialArgs->phiforce = &SpiralArmsPotentialphiforce;
+      potentialArgs->R2deriv = &SpiralArmsPotentialR2deriv;
+      //potentialArgs->z2derib = &SpiralArmsPotentialz2deriv;
```
jobovy: Should be deriv not derib

galpy/orbit_src/orbit_c_ext/integrateFullOrbit.c
```diff
@@ -231,7 +231,18 @@ void parse_leapFuncArgs_Full(int npot,
       potentialArgs->zforce= &DiskSCFPotentialzforce;
       potentialArgs->phiforce= &ZeroForce;
       potentialArgs->nargs= (int) *(pot_args) + 3;
-      break;      
+      break;
+    case 27: // SpiralArmsPotential, 10 arguments + array of Cs
+      potentialArgs->Rforce = &SpiralArmsPotentialRforce;
+      potentialArgs->zforce = &SpiralArmsPotentialzforce;
+      potentialArgs->phiforce = &SpiralArmsPotentialphiforce;
+      potentialArgs->R2deriv = &SpiralArmsPotentialR2deriv;
```
jobovy: Comment out the second derivatives, because they aren't used yet.

galpy/potential_src/SpiralArmsPotential.py
```diff
@@ -0,0 +1,584 @@
...
+
+def check_inputs_not_arrays(func):
+    """
+    Decorator to check inputs and throw TypeError if any of the inputs are arrays.
+    Methods potentially return with silent errors if inputs are not checked.
+    """
+    @wraps(func)
+    def func_wrapper(self, R, z, phi, t):
+        if hasattr(R, '__len__') or hasattr(z, '__len__') or hasattr(phi, '__len__') or hasattr(t, '__len__'):
+            raise TypeError('Methods in SpiralArmsPotential do not accept array inputs. Please input scalars.')
```
jobovy: We need to add a test of this Error. Write a little test function in ``test_potential.py`` similar to [this function](https://github.com/jobovy/galpy/blob/master/tests/test_potential.py#L1626) for that.

galpy/potential_src/SpiralArmsPotential.py
```diff
@@ -0,0 +1,584 @@
...
+        self._phi_ref = phi_ref
+        self._Rs = Rs
+        self._H = H
+        self._Cs = np.array(Cs)
+        self._ns = np.arange(1, len(Cs) + 1)
+        self._omega = omega
+        self._rho0 = 1 / (4 * np.pi)
+        self._HNn = self._H * self._N * self._ns
+
+        if normalize or (isinstance(normalize, (int, float)) and not isinstance(normalize, bool)):
```
jobovy: I think we should get rid of the ``normalize`` keyword and these lines. Normalizing this potential doesn't make sense.

tests/test_SpiralArmsPotential.py
```diff
@@ -0,0 +1,864 @@
...
+        assert sp._Cs == [1]
+        assert sp._omega == 0
+        assert sp._rho0 == 1 / (4 * pi)
+        assert sp.isNonAxi == True
+        assert sp.hasC == True
+        assert sp.hasC_dxdv == True
+        assert sp._ro == 8
+        assert sp._vo == 220
+
+        sp = spiral(N=3, alpha=10*u.deg, r_ref=1, phi_ref=0, Rs=0.5, H=0.5, Cs=[1], omega=0)
```
jobovy: Let's test that units are handled properly by adding a test in ``test_quantity.py`` (as mentioned in the comment to the PR). Right now the ``test_SpiralArmsPotential.py`` because astropy is not installed for that test and it's cleaner to separate the units test.

doc/source/reference/potentialdehnenbar.rst
```diff
@@ -1,5 +1,5 @@
 Dehnen bar potential
-========================
+====================
```
jobovy: For future reference, it would be better to not make any changes to files in a PR that you don't actually need to change. I suspect your editor has changed the length of these title headings (the ===) automatically. It doesn't matter much here, but in general this would be better to avoid (you can restore files that you didn't want to change if you notice them with ``git status`` by doing ``git checkout -- FILE``).

galpy/potential_src/SpiralArmsPotential.py
```diff
@@ -33,42 +33,60 @@ def func_wrapper(self, R, z, phi, t):
...
+        \\Phi(r, \\phi, z) = -4 \\pi GH \\rho_0 exp \\left( -\\frac{r-r_0}{R_s} \\right) \\sum{\\frac{C_n}{K_n D_n} cos(n \\gamma) sech \\left( \\frac{K_n z}{B_n} \\right)^{B_n}}
 
+    where
+
+    .. math::
+        K_n &= \\frac{n N}{r sin(\\alpha)} \\\\
+        B_n &= K_n H (1 + 0.4 K_n H) \\\\
+        D_n &= \\frac{1 + K_n H + 0.3 (K_n H)^2}{1 + 0.3 K_n H} \\\\
+
+    From (`Cox and Gomez 2002 <https://arxiv.org/abs/astro-ph/0207635>`__):
```
jobovy: Not entirely sure how this looks when rendered, but do you also mention the default case C=1?

galpy/potential_src/SpiralArmsPotential.py
```diff
@@ -33,42 +33,60 @@ def func_wrapper(self, R, z, phi, t):
...
             :param H: scale height of the stellar arm perturbation (can be Quantity)
-            :param Cs: list of constants multiplying the cos(n*gamma) term in the mass density expression
-            :param omega: rotational speed of the spiral arms (can be Quantity)
+            :param Cs: list of constants multiplying the :math:`cos(n \\gamma)` term in the mass density expression
+            :param omega: rotational pattern speed of the spiral arms (can be Quantity)
         OUTPUT:
             (none)
         HISTORY:
-            2017-05-12  Jack Hong (UBC)
+            2017-07-04  Jack Hong (UBC)
```
jobovy: Keep the full history here: date when you started it, and additional dates when you made big changes.

galpy/potential_src/potential_c_ext/SpiralArmsPotential.c
```diff
@@ -308,28 +365,35 @@ double SpiralArmsPotentialz2deriv(double R, double z, double phi, double t,
...
-        double tanh2_zKB = tanh(zKB) * tanh(zKB);
+        zKB = z * Kn / Bn;
+        tanh2_zKB = tanh(zKB) * tanh(zKB);
 
         sum += Cn * Kn / Dn * ((tanh2_zKB - 1) / Bn + tanh2_zKB) * cos(n * g) / pow(cosh(zKB), Bn);
     }
 
     return -amp * H * exp(-(R - r_ref) / Rs) * sum;
 }
+//LCOV_EXCL_STOP
```
jobovy: Think you need a 
```
//LCOV_EXCL_START
```
before phi2deriv as well (and any second derivative below that is not planar, like Rphideriv).

doc/source/reference/potential.rst
```diff
@@ -139,7 +139,7 @@ Axisymmetric potentials
    potentialsnapshotrzpotential.rst
 
 Triaxial and bar potentials
```
jobovy: Change to
```
Triaxial, spiral, and bar potentials
```

jobovy:
The ``test_phiforce`` test in ``test_SpiralArmsPotential.py`` failed: https://travis-ci.org/jobovy/galpy/jobs/250162854#L2764

Also, you seem to have removed the special and rotatingSpiralArmsPotential in the orbit-integration test? The orbit integration for these needs to be tested (and the rotating one should also be added to ``test_liouville``).

Otherwise everything is looking good.

jobovy:
All looks good, thanks @jackhong6 ! :tada::tada::tada:

PR #330, opened 2018-01-04
Staeckelecc - documentation and no_median

jmackereth:
Added documentation for new analytic orbit parameters and EccZmaxRperiRap actionAngle method.

Also added no_median option to estimateDeltaStaeckel with relevant tests.


doc/source/examples/dierickx_eccentricities.py
```diff
@@ -0,0 +1,192 @@
+import os, os.path
+import pexpect
```
jobovy: Is this actually used? Doesn't seem to be a standard library module, so users would have to install it.

tests/test_actionAngle.py
```diff
@@ -1257,6 +1258,24 @@ def test_actionAngleStaeckel_indivdelta_actions():
...
+	# Briefly integrate orbit to get multiple points
+	o= Orbit([1.,0.1,1.1,0.001,0.25,1.])
+	ts= numpy.linspace(0.,1.,101)
+	o.integrate(ts,MWPotential2014)
+	#generate no_median deltas
+	nomed = estimateDeltaStaeckel(MWPotential2014, o.R(ts[:10]), o.z(ts[:10]), no_median=True)
+	#and the individual ones
+	indiv = numpy.array([estimateDeltaStaeckel(MWPotential2014, o.R(ts[i]), o.z(ts[i])) for i in range(10)])
+	#check that values agree
+	assert (numpy.fabs(nomed-indiv) < 1e10).all(), 'no_median option returns different values to individual Delta estimation'
```
jobovy: This seems like a good test, but I think it should be ``1e-10`` rather than ``1e10``!

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
...
+
+Fast orbit characterization
+-----------------------
+
+It is also possible to use galpy for the fast estimation of orbit parameters via the Staeckel 
+approximation (originally used by `Binney (2012) <http://adsabs.harvard.edu/abs/2012MNRAS.426.1324B>`_
+for the appoximation of actions in axisymmetric potentials), without performing any orbit integration. 
+The method uses the geometry of the orbit tori to estimate the orbit parameters. After initialising 
+an ``Orbit`` instance, this is done (as in the previous section) specifying ``analytic=True`` and 
+selecting ``type='staeckel'`` (default in vX.X of galpy).
```
jobovy: No need for the 

(default in vX.X of galpy)

The documentation online is automatically the right one for this version)

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
 
 .. image:: images/lp-orbit-integration-EzJz.png
 
+.. _fastchar:
+
+Fast orbit characterization
+-----------------------
+
```
jobovy: Somewhere early in this section need to say that this is from Mackereth & Bovy (2018, in prep) which we will later replace with the link. Don't want people to just cite Binney's paper!

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
...
+selecting ``type='staeckel'`` (default in vX.X of galpy).
+
+>>> o.e(analytic=True, type='staeckel')
+
+if running the above without integrating the orbit, the potential should also be specified
+in the usual way
+
+>>> o.e(analytic=True, type='staeckel', pot=mp)
+
+again, where ``mp`` is the Miyamoto-Nagai potential of :ref:`Introduction:
```
jobovy: Sorry to be nitpicking, but this ``mp`` has already been defined earlier in this page, so we can just assume that the user has done that (just like the ``o`` object itself)

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
...
+Rotation curves <rotcurves>`. This interface automatically computes the necessary Delta
```
jobovy: estimates rather than computes? Also 'delta' rather than 'Delta', because the actual keyword is delta

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
...
+parameter based on the initial condition of the ``Orbit`` object.
+
+While this is useful and fast for individual ``Orbit`` objects, it is likely that users will
+want to rapidly evaluate the orbit parameters of large numbers of objects. It is possible
+to perform the orbital parameter estimation above through the :ref:`actionAngle <actionangle>` 
+interface. To do this, we need arrays of the phase-space points ``R``, ``vR``, ``vT``, ``z``, ``vz``, and 
+``phi`` for the objects. We can then optionally estimate the individual Delta parameter 
+for these phase-space points using
+
+>>> from galpy.actionAngle import estimateDeltaStaeckel
```
jobovy: Can we move this to later so that the logical flow is

1) Use actionAngleStaeckel for arrays
2) Default of that is to use a constant delta --> show example where aAS= defined using single delta and then used
3) Point out that you can give ``EccZmaxRperiRap`` an array which can be computed using ``estimateDeltaStaeckel``

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
...
+``EccZmaxRperiRap`` method with the data points and the estimated delta array:
+
+>>> aAS = actionAngleStaeckel(pot=mp, delta=0.4)
+>>> e, Zmax, rperi, rap = aAS.EccZmaxRperiRap(R, vR, vT, z, vz, phi, delta=deltas)
+
+Here, delta can alternatively be either a single scalar value for all points, negating the need for
+the estimation at each point. This method is also applicable in the ``actionAngleIsochrone``, 
+``actionAngleSpherical``, and ``actionAngleAdiabatic`` modules. This method returns parameters at
+speeds as fast as 3 microseconds per object. 
+
```
jobovy: Is it really 3 microseconds? You could put an example with ``timeit`` to explicitly show this (so people can easily test this themselves)

Can we also add an example of using the actionAngleStaeckelGrid approach and how fast that is?

doc/source/orbit.rst
```diff
@@ -346,6 +346,52 @@ behavior
 
 .. image:: images/lp-orbit-integration-EzJz.png
 
+.. _fastchar:
+
+Fast orbit characterization
+-----------------------
```
jobovy: My sphinx build complains that the --- aren't long enough (they need to extend at least as far as the title)

doc/source/orbit.rst
```diff
@@ -507,32 +553,70 @@ a set of thick disk stars. We start by downloading the sample of SDSS
...
+...     except UnboundError:
+...         #parameters cannot be estimated analytically
+...         e_ana[i] = np.nan
+...     #integrate the orbit and return the numerical e value
+...     orbit.integrate(ts, lp)
+...     e_int[i] = orbit.e(analytic=False)
+
+We then find the following eccentricity distribution (from the numerical eccentricities)
+
+.. image:: images/dierickx-integratedehist.png
```
jobovy: This figure is enormous on my screen, maybe we should make it smaller? I think you can add
```
    :scale: 50 %
```
or similar. Same for the figures below.

doc/source/examples/dierickx_eccentricities.py
jmackereth: think that got left in there from an earlier version i was tinkering with - can definitely be thrown out!

jobovy:
Thanks, this all looks good now.

I made some more small changes to the documentation, but I'm having problems pushing them to your branch. Did you click the option "Allow commits from maintainer" in the PR?

jmackereth:
Thanks! Yes, i just checked and that option is on, so I'm not sure what the problem is there?

jobovy:
Thanks for checking! I figured it out, for some reason git wasn't allowing me to push to your fork's ``staeckelecc`` from a branch that wasn't called ``staeckelecc`` itself (I had your changes in ``jmackereth-staeckelecc``).

jobovy:
One more thing, could you add an entry in the ``HISTORY.txt`` file about the ``no_median`` change and add yourself to the ``AUTHORS.txt``? Please include ``[ci skip]`` in the commit messages for those changes; this causes the test suite to not be rerun.

PR #334, opened 2018-03-19
Addrtide

webbjj:
Hi Jo

As promised, here is the pull request. Even though I had 4 commits, the first few were mostly me learning. The bulk of the changes were in the last commit and as you can see its not that extensive. Let me know if there are any issues.

Jeremy


galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+            Tidal Radius
+        
+        HISTORY:
+            
+            2018-02-12 - Written - Webb (UofT)
+            
+        """
+        if self.isNonAxi:
+            #Make sure potential is axisymmetric
+            print("Error - cannot calculate rtide for non-axisymmetric potential")
```
jobovy: Use a ``PotentialError`` here:
```
raise PotentialError("Cannot calculate rtide for non-axisymmetric potential")
```

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
         except AttributeError:
             raise AttributeError('NEMO acceleration parameters not supported for %s' % self.__class__.__name__)
 
+    @potential_physical_input
+    @physical_conversion('position',pop=True)
+    def rtide(self,R,z,M=None,M_units=None):
```
jobovy: Remove ``M_units``

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+        """
+        if self.isNonAxi:
+            #Make sure potential is axisymmetric
+            print("Error - cannot calculate rtide for non-axisymmetric potential")
+            return 0.0
+        elif M==None:
+            #If units not given, assume galpy! GM units?
+            print("Error - mass parameter has not been set")
+            return 0.0
+        else:
```
jobovy: We should just assume the following two cases:

1. M is a scalar --> assumed to be in galpy units --> divide by bovy_conversion.mass_in_msol
2. M is a Quantity --> try units of mass or Gxmass

So I think this entire piece of code should be
```
### NO ELSE TO BEGIN
if _APY_LOADED and isinstance(M,units.Quantity):
    try:
        M= M.to(units.Msun).value/bovy_conversion.mass_in_msol(self._vo,self._ro)
    except units.UnitConversionError: pass
    else:
        try:
            M= M.to(units.pc*units.km**2/units.s**2).value/bovy_conversion.mass_in_msol(self._vo,self._ro)/bovy_conversion._G
        except units.UnitConversionError: pass
            raise units.UnitConversionError('M= parameter of rtide should have units of mass')
```

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+                            .value\
+                            /bovy_conversion.mass_in_msol(self._vo,self._ro)\
+                            /bovy_conversion._G
+                    except units.UnitConversionError: pass
+                    else:
+                        unitFound= True
+                        if not M_units == 'mass':
+                            raise units.UnitConversionError('M= parameter of %s should have units of %s, but has units of G x mass instead' % (type(self).__name__,M_units))
+                                    
+            #To calculate omegac in spherical coordinates we need the first derivative of the potential with respect to r
```
jobovy: Switch back to the regular method indentation (you're doing this under the ``else`` clause here)

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+            r=math.sqrt(R**2.0+z**2.0)
+            theta=nu.arctan2(R, z)
+            
+            Rderiv=-1.0*self.Rforce(R,z,use_physical=False)
+            zderiv=-1.0*self.zforce(R,z,use_physical=False)
+
+            omegac2=(Rderiv*(R/r)+zderiv*(z/r))/r
```
jobovy: ``Potential`` instances have a method ``rforce`` that gives the radial force (minus the derivative of the potential wrt spherical radius), isn't that just what you want to use here?

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+            
+            #To calculate the epicyclic frequency kappa we need the second derivative of the potential with respect to r
+            R2deriv= self.R2deriv(R,z,use_physical=False)
+            z2deriv=self.z2deriv(R,z,use_physical=False)
+            Rzderiv=self.Rzderiv(R,z,use_physical=False)
+            
+            r2deriv=(R2deriv*nu.sin(theta)+Rzderiv*nu.cos(theta))*nu.sin(theta)+(Rzderiv*nu.sin(theta)+z2deriv*nu.cos(theta))*nu.cos(theta)
```
jobovy: I think it would be good to introduce a new method ``r2deriv`` (like the way ``rforce`` is implemented) that gives the second derivative wrt the spherical radius and then use that here. This function might be useful more generally.

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+
+            #rt=(GM/(omegac2*nu))**(1/3)
+            kappa2=3.0*omegac2+r2deriv
+            nuu=4.0-kappa2/omegac2
+            return (M/(omegac2*nuu))**(1.0/3.0)
+
+    @potential_physical_input
+    @physical_conversion('forcederivative',pop=True)
+    def ttensor(self,R,z,phi):
```
jobovy: Make ``phi`` optional as ``phi=0.``  to allow this to be easily used with axisymmetric potentials

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+            Tidal Tensor
+        
+        HISTORY:
+        
+            2018-03-07 - Written - Webb (UofT)
+            
+        """
+        if self.isNonAxi:
+            #Make sure potential is axisymmetric
+            print("Error - cannot calculate tidal tensor for non-axisymmetric potential")
```
jobovy: Use ``PotentialError`` again

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+            return 0.0
+        else:
```
jobovy: No ``else`` here

galpy/potential_src/Potential.py
```diff
@@ -1510,6 +1510,166 @@ def nemo_accpars(self,vo,ro):
...
+
+            #rt=(GM/(omegac2*nu))**(1/3)
+            kappa2=3.0*omegac2+r2deriv
+            nuu=4.0-kappa2/omegac2
+            return (M/(omegac2*nuu))**(1.0/3.0)
+
+    @potential_physical_input
+    @physical_conversion('forcederivative',pop=True)
+    def ttensor(self,R,z,phi):
+        """
```
jobovy: Would it be useful to make it an option to return the eigenvalues instead? I.e., aren't we often interested in the eigenvalues?

galpy/potential_src/Potential.py
```diff
@@ -2755,3 +2915,164 @@ def kms_to_kpcGyrDecorator(func):
     def kms_to_kpcGyr_wrapper(*args,**kwargs):
         return func(args[0],velocity_in_kpcGyr(args[1],1.),args[2],**kwargs)
     return kms_to_kpcGyr_wrapper
+
+@potential_physical_input
+@physical_conversion('position',pop=True)
+def rtide(Pot,R,z,M=None,M_units=None):
```
jobovy: Basically all comments from ``rtide`` above apply to this function as well.

galpy/potential_src/Potential.py
```diff
@@ -2755,3 +2915,164 @@ def kms_to_kpcGyrDecorator(func):
...
+        Tidal Tensor
+        
+    HISTORY:
+        
+        2018-03-19 - Written - Webb (UofT)
+            
+    """
+    if Pot.isNonAxi:
+        #Make sure potential is axisymmetric
+        print("Error - cannot calculate tidal tensor for non-axisymmetric potential")
```
jobovy: Use ``PotentialError`` instead

webbjj:
Ah shoot, I actually committed too soon. Sorry. Left in a comment that I
wanted to ask you about. Pretty sure I do not need to do the axisymmetric
check for rtide since we just need rforce and r2deriv, which shouldn't
depend on whether or not the potential is axisymmetric. It just matters
what the values are at the cluster's current position because the
calculation itself assumes the cluster has a circular orbit its
instantaneous position. For ttensor though the check is still necessary. In
time / if necessary I could add the phi dependence to ttensor.

Jeremy

galpy/potential_src/Potential.py
```diff
@@ -389,7 +389,46 @@ def rforce(self,R,z,phi=0.,t=0.):
...
         return self.Rforce(R,z,phi=phi,t=t,use_physical=False)*R/r\
             +self.zforce(R,z,phi=phi,t=t,use_physical=False)*z/r
+
+    @potential_physical_input
+    @physical_conversion('forcederivative',pop=True)
+    def r2deriv(self,R,z,phi=0.,t=0.):
+        """
+        NAME:
+
+           rderiv
```
jobovy: --> r2deriv

galpy/potential_src/Potential.py
```diff
@@ -1539,73 +1580,49 @@ def rtide(self,R,z,M=None,M_units=None):
         
         HISTORY:
             
-            2018-02-12 - Written - Webb (UofT)
+            2018-03-21 - Written - Webb (UofT)
             
         """
+        #DO WE NEED THIS? PHI IS WORKED INTO RFORCE AND R2DERIV?
```
jobovy: If the result is sensible when just evaluating the forces and derivatives at R,z,phi,t then I agree that we don't need to do the check, so you can remove it.

galpy/potential_src/Potential.py
```diff
@@ -1539,73 +1580,49 @@ def rtide(self,R,z,M=None,M_units=None):
...
         if self.isNonAxi:
             #Make sure potential is axisymmetric
-            print("Error - cannot calculate rtide for non-axisymmetric potential")
+            raise PotentialError("Cannot calculate rtide for non-axisymmetric potential")
             return 0.0
         elif M==None:
-            #If units not given, assume galpy! GM units?
-            print("Error - mass parameter has not been set")
+            #Make sure an object mass is given
+            print("RTIDE Error - mass parameter has not been set")
```
jobovy: Also raise PotentialError here instead

galpy/potential_src/Potential.py
```diff
@@ -1619,55 +1636,59 @@ def ttensor(self,R,z,phi):
...
         
-            2018-03-07 - Written - Webb (UofT)
-            
+            2018-03-21 - Written - Webb (UofT)
+
         """
         if self.isNonAxi:
             #Make sure potential is axisymmetric
-            print("Error - cannot calculate tidal tensor for non-axisymmetric potential")
+            raise PotentialError("Cannot calculate ttensor for non-axisymmetric potential")
```
jobovy: So again everything is just computed based on local forces and second derivatives, so why can this not be done for a non-axisymmetric potential?

galpy/potential_src/Potential.py
```diff
@@ -2918,100 +2939,79 @@ def kms_to_kpcGyr_wrapper(*args,**kwargs):
...
     if Pot.isNonAxi:
         #Make sure potential is axisymmetric
-        print("Error - cannot calculate rtide for non-axisymmetric potential")
+        raise PotentialError("Cannot calculate rtide for non-axisymmetric potential")
         return 0.0
     elif M==None:
-        #If units not given, assume galpy! GM units?
-        print("Error - mass parameter has not been set")
+        #Make sure an object mass is given
+        print("RTIDE Error - mass parameter has not been set")
```
jobovy: PotentialError here too.

webbjj:
Hi Jo

As you will likely see I made another commit. I believe this address all
your comments and I added a phi component to ttensor, but set zphideriv=0.0
since as we discussed this doesn't exist yet.

Ive attached my derivation of the tidal tensor components in case you want
to check it. The x and y derivatives I was able to cross check with your
planar_orbit.py file. The xz, yz, zx, and zy derivates I wasn't able to
cross check but they weren't terribly hard (I don't think!).

Let me know if you have more comments.

Jeremy

Issue #337, opened 2018-04-17
.pyd image not found


senorPepe:
Dear Jo,

I recently downloaded and installed GalPy for a university project. However, I get the following warnings when running the examples from your paper. I do not know how to solve this and what to do next. Hope you can help, thanks in advance!

Ps. This is from running figure 21:
galpyWarning: interppotential_c extension module not loaded, because galpy_interppotential_c.cp36-win_amd64.pyd image was not found
galpyWarning: integratePlanarOrbit_c extension module not loaded, because galpy_integrate_c.cp36-win_amd64.pyd image was not found
galpyWarning: integrateFullOrbit_c extension module not loaded, because galpy_integrate_c.cp36-win_amd64.pyd image was not found
galpyWarning: actionAngleAdiabatic_c extension module not loaded, because galpy_actionAngle_c.cp36-win_amd64.pyd image was not found
galpyWarning: actionAngleStaeckel_c extension module not loaded, because galpy_actionAngle_c.cp36-win_amd64.pyd image was not found

jobovy:
Hi,

It seems like you are on a Windows system? If you installed the ``conda`` or ``pip`` version of the code, these warnings are expected, because the C extensions were not properly compiled in the latest version. Depending on what you want to do with the code, these warnings may be unimportant (almost all of the code still works, it just won't be as fast).

The latest ``master`` version of the code can now actually be compiled on Windows, I finally fixed this after the latest release. See [this section of the documentation](http://galpy.readthedocs.io/en/latest/installation.html#installing-from-source-on-windows) for instructions on how to do that. Note that compiling code on Windows can be quite difficult if you haven't done it before.

henrysky:
Here are two galpy wheels I have built if you have difficulty: galpy 1.3 (27 Feb 2018) and galpy 1.4dev (18 April 2018) you can use for Win64 py36

**Use with caution as I am not doing any serious thing with galpy and I will not guarantee the wheels I built work correctly**
https://drive.google.com/drive/folders/1tai7diGXywauP4Sn-_qVUg_SMkTAiey0

senorPepe:
Thank you both! However, it still not works. Firstly, I cannot seem to ignore the warnings as suggested by mr. Bovy. 
And secondly, using the wheels henrysky built, I get the following error:
[...]

PR #338, opened 2018-05-02
Fix frame as a positional argument is deprecated in astropy

henrysky:
Fix frame as a positional argument is deprecated in astropy, adding frame="xxx" to prevent `AstropyDeprecationWarning` from displaying every time

Related:
#337


jobovy:
Thanks! I had actually already fixed this, but hadn't pushed it to GitHub yet. That's done now, so will close this.

PR #344, opened 2018-07-10
Add Orbit.from_name class method

mwbub:
Orbit.from_name class method, which uses astroquery to generate Orbits from the name of an object in an astronomical catalogue.


mwbub:
Hi Jo,

The test suite fails on the import of astroquery:
```
if not _ASTROQUERY_LOADED:
>           raise ImportError(('astroquery needs to be installed to use '
                               'Orbit.from_name'))
E           ImportError: astroquery needs to be installed to use Orbit.from_name
```
I'm not sure if that's something I need to fix myself or not. Let me know if there's anything I have to change.

Thanks,
Mathew

jobovy:
Yes, ``astroquery`` needs to be installed on both travis (which runs the Linux/Mac tests) and appveyor (which runs the Windows tests); I've added a commit to do this, so please pull this before doing antything further.

jobovy:
Note that every time you push the code to GitHub now that the PR is open, the test suite gets run. If you change something that doesn't change the code (like adding blank lines), please end the commit message with ``[ci skip]``, which will cause the tests to be skipped if they don't have to be run.

mwbub:
Ok, sorry, I didn't know there was an option to skip the tests. I won't be making any more changes anyway. Thanks.

galpy/orbit_src/Orbit.py
```diff
@@ -3774,6 +3779,140 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+        OUTPUT:
+
+            orbit containing the phase space coordinates of the named object
+
+        HISTORY:
+
+            2018-07-09 - Written - Mathew Bub (UofT)
+
+        """
+        if not _APY_LOADED:
```
jobovy: Add ``# pragma: no cover``

galpy/orbit_src/Orbit.py
```diff
@@ -3774,6 +3779,140 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+            raise ImportError(('astropy needs to be installed to use '
+                               'Orbit.from_name'))
+        if not _ASTROQUERY_LOADED:
```
jobovy: Add ``# pragma: no cover``

galpy/orbit_src/Orbit.py
```diff
@@ -3774,6 +3779,140 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+            raise ValueError('failed to find {} in SIMBAD'.format(name))
+        simbad_table= simbad_table.as_array().view((nu.float64,6))
+
+        # try to cross-match the SIMBAD coordinates with Gaia DR2
+        if gaiadr2:
+            from astroquery.gaia import Gaia
+            if searchr is None:
+                searchr= 1
+            if isinstance(searchr, units.Quantity):
+                searchr= searchr.to(units.deg).value
```
jobovy: This line is currently not hit by any of the tests, please add or edit one of the tests to give ``r`` with units so this gets hit.

tests/test_orbit.py
```diff
@@ -4280,6 +4280,163 @@ def test_full_plotting():
     else: raise AssertionError("plot3d(d3='vy') applied to RZOrbit did not raise AttributeError")
     return None
 
+def test_from_name():
```
jobovy: Split these tests into a few more specific tests:
  * Basic test that the correct object is found
  * Tests of the various errors
  * Tests of the warnings
  * ...

This way it would be easier to tell if one of the tests fail which part of the code is the issue.

tests/test_orbit.py
```diff
@@ -4280,6 +4280,163 @@ def test_full_plotting():
...
+        "PMRA of Lacaille 8760 does not match Gaia value"
+    assert numpy.isclose(o.pmdec(), -1145.3957291901002), \
+        "PMDec of Lacaille 8760 does not match Gaia value"
+    assert numpy.isclose(o.vlos(), 20.561701480854417), \
+        "radial velocity of Lacaille 8760 does not match Gaia value"
+
+    # test GJ 440 from Gaia
+    with pytest.raises(ValueError) as excinfo:
+        with pytest.warns(galpyWarning) as record:
+            warnings.filterwarnings('ignore', 'unclosed', ResourceWarning, 'socket')
```
jobovy: ``ResourceWarning`` is not defined on python2, is this really necessary? If not, just remove it, if yes, then you need to test the python version...

mwbub: OK, I will make those changes.

I put in a filter for ResourceWarnings in the tests because socket throws a warning that the connection is unclosed. This is normally ignored by default in Python, but it is caught when I explicitly test for the warnings. I will remove the ResourceWarning part and replace it with a general filter for warnings from socket.

mwbub:
Hi Jo,

I notice that SIMBAD has completed a limited cross-match of their data with the second Gaia release. Would it be worthwhile to try to use their cross-matched data either in addition to or as a replacement of the current cross-matching functionality of from_name?

jobovy:
Yes, if it would be possible to avoid the explicit Gaia query, that would be better, so perhaps you should look into that and perhaps replace the current Gaia match. What is limited about SIMBAD's cross match?

mwbub:
I only meant that the cross-match is 'limited' in the sense that certain limits were placed on the cross-match to ensure that the results were correct. For instance, they don't look at any stars with a neighbour closer than 3 arcseconds, etc. Full details are here: http://cdsweb.u-strasbg.fr/news.php?fn_mode=fullnews&fn_incl=0&fn_id=702. For all practical purposes this shouldn't make any difference, however. 

I will put this on hold for the moment and look into possibly replacing the Gaia cone search with directly grabbing the Gaia object identifier from SIMBAD.

jobovy:
Okay, thanks. Looking at that documentation, it seems like it would indeed be better to use the SIMBAD crossmatch (for one thing, they use the *Gaia* proper motions to propagate the positions to 2000 coordinates, rather than SIMBAD proper motions to go to 2015.5; *Gaia* should be better).

I think the PR is otherwise ready, but I'll leave it open for this change.

mwbub:
Just to clarify, my current plan regarding the change is to retreive the Gaia source_id from SIMBAD, then use that to grab the object coordinates from the Gaia archive. In that case I would still be getting them in 2015.5. Is that what you had in mind, or were you thinking that I shouldn't query the Gaia archive at all?

jobovy:
It looks to me like the 'plx' and 'pm' values in SIMBAD are now just those from Gaia DR2? For example, http://simbad.u-strasbg.fr/simbad/sim-id?Ident=Lacaille+8760: the catalog given as the source for parallax, proper motion, and vlos is Gaia DR2.

mwbub:
Yeah, I think you're right. It looks like they updated their sources. I guess the quesion is whether there is any value in the query to Gaia? Do you think I should keep it in or should I cut it down to just use the values from SIMBAD?

jobovy:
I don't think there is now any real reason to keep the *Gaia* query, which is good because it significantly simplifies the code (also making it much easier to maintain this in the future!). Good that you found this before we merged this in!

mwbub:
In that case, I have some leeway to make the method more versatile if you would like. For instance, I was thinking that the method could check if SIMBAD has any non-parallax distances in case parallax is unavailable for some object. Would you like an extra feature like that and/or some other feature, or should I just leave it as is?

jobovy:
I'm happy for the function to use any distance that is available (preferring the parallax if it's available). The point is to be able to setup some reasonable Orbit instance using available data, if people really need to be sure that what they are using is correct, they should look into it more closely, this is just to get *some* Orbit.

If there's any way to have ``Orbit.from_name('LMC')`` to work, that would also be great...

mwbub:
Hi Jo,

I removed the Gaia query and added a check for alternative distance measurements. The new code is quite a bit simpler and more readable. Let me know if you have any additional concerns.

Thanks,
Mathew

jobovy:
Thanks, this is great! Just had a lot of fun doing things like ``o= Orbit.from_name('LMC')`` or ``o= Orbit.from_name('Omega Cen')`` (a prominent globular cluster) and plotting their orbits in the past and future! All seems to be working well. SIMBAD seems to time out a lot (sometimes I couldn't get anything to work for a few minutes), but that seems to just be an issue with the service.

I'll merge this, thanks!

Issue #348, opened 2018-08-16
Windows galpy pure python installation issue


henrysky:
Today, me and @samuelwong100 encountered a galpy installation issue on Windows that it is impossible to install galpy without a compiler at all and the solution is to rename `gsl-config.bat` such that setup.py cannot find it. This issue is caused by the changes made for issue #331 .

I personally think it is a bug because of galpy think compiler is presented as long as gsl can be found which is not true at least for @samuelwong100 laptop.

Not sure if you think it should be fixed and one of my proposed solution is to initialize a test compiler and see if it fails which the solution works in my Windows 10 virtual machine where gsl is installed but no compiler. I can make a PR if you think it should be fixed

```
import distutils.ccompiler
from distutils.errors import DistutilsPlatformError 

try:
    test_compiler = distutils.ccompiler.new_compiler()
    test_compiler.initialize()
    compiler_is_here_flag = True
except DistutilsPlatformError:   # this error will be raised if no compiler in the system
    compiler_is_here_flag = False
```

jobovy:
Interesting! Yes, it would be good to have a decent fix to this in the ``setup.py`` file and what you suggest seems like a good start, so please make a PR. I guess what you'll want to do when there is no compiler is to set ``ext_modules=None`` or ``ext_modules=[]`` just before invoking ``setup`` (not sure which of these works).

Maybe call the flag ``no_compiler``?

jobovy:
Fixed in #349.

PR #349, opened 2018-08-18
Check compiler

henrysky:
check `msvc` compiler exists and can be initialize on windows even if `gsl` is installed.

A PR for issue #348

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ from distutils.core import Extension
 import sys
 import distutils.sysconfig as sysconfig
 import distutils.ccompiler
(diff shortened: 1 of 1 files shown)


jobovy:
Looks good 👍 ! Because the tests don't cover this case, can you confirm that this works when one doesn't have a compiler on windows? Or maybe @samuelwong100 can? If you confirm, I'll merge.

henrysky:
Yes I can confirm that this works when one doesn't have a compiler on windows in a virtual machine, `galpy` installed with pure python.

jobovy:
Thanks! 👍

PR #350, opened 2018-08-24
Add qdf sampleV_interpolate

samuelwong100:
I added a sample velocity function to qdf that samples for many positions at once and is faster than the old sampleV function because it uses interpolation.

### DIFF
--- a/galpy/df/quasiisothermaldf.py
+++ b/galpy/df/quasiisothermaldf.py
@@ -1773,14 +1773,8 @@ class quasiisothermaldf(df):
                 grid_max_vT[i][j]= optimize.fmin_powell((lambda x: -self(
                         R,0.,x,z,0.,log=True, use_physical=False)),1.)
         #Determine degree of interpolation
-        if R_number>3:
-            ky= 3
-        else:
-            ky= R_number-1
-        if z_number>3:
-            kx= 3
-        else:
-            kx= z_number-1
+        ky= numpy.min([R_number-1,3])
+        kx= numpy.min([z_number-1,3])
         #Generate interpolation object
         ip_max_vT= interpolate.RectBivariateSpline(z_linspace,R_linspace,
                                                     grid_max_vT,kx=kx,ky=ky)
--- a/galpy/df/quasiisothermaldf.py
+++ b/galpy/df/quasiisothermaldf.py
@@ -1685,6 +1685,73 @@ class quasiisothermaldf(df):
         else:
             return out
 
+    @potential_physical_input
+    def sampleV_preoptimized(self,R,z,maxVT,n=1):
+        """
+        NAME:
+
+           sampleV_preoptimized
+
+        PURPOSE:
+
+           sample a radial, azimuthal, and vertical velocity at R,z;
+           R,z can be an array of positions
+
+        INPUT:
+
+           R - Galactocentric distance (can be Quantity)
+
+           z - height (can be Quantity)
+           
+           maxVT - an array of pre-optimized maximum vT at corresponding R,z
+
+           n= number of distances to sample
+
+        OUTPUT:
+
+           list of samples
+
+        HISTORY:
+
+           2012-12-17 - Written - Bovy (IAS)
+
+        """
+        #Determine the maximum of the velocity distribution
+        maxVR= 0.
+        maxVz= 0.
+        maxVT= optimize.fmin_powell((lambda x: -self(R,0.,x,z,0.,log=True,
+                                                     use_physical=False)),
+                                    1.)
+        logmaxVD= self(R,maxVR,maxVT,z,maxVz,log=True,use_physical=False)
+        #Now rejection-sample
+        vRs= []
+        vTs= []
+        vzs= []
+        while len(vRs) < n:
+            nmore= n-len(vRs)+1
+            #sample
+            propvR= numpy.random.normal(size=nmore)*2.*self._sr
+            propvT= numpy.random.normal(size=nmore)*2.*self._sr+maxVT
+            propvz= numpy.random.normal(size=nmore)*2.*self._sz
+            VDatprop= self(R+numpy.zeros(nmore),
+                           propvR,propvT,z+numpy.zeros(nmore),
+                           propvz,log=True,use_physical=False)-logmaxVD
+            VDatprop-= -0.5*(propvR**2./4./self._sr**2.+propvz**2./4./self._sz**2.\
+                                 +(propvT-maxVT)**2./4./self._sr**2.)
+            VDatprop= numpy.reshape(VDatprop,(nmore))
+            indx= (VDatprop > numpy.log(numpy.random.random(size=nmore))) #accept
+            vRs.extend(list(propvR[indx]))
+            vTs.extend(list(propvT[indx]))
+            vzs.extend(list(propvz[indx]))
+        out= numpy.empty((n,3))
+        out[:,0]= vRs[0:n]
+        out[:,1]= vTs[0:n]
+        out[:,2]= vzs[0:n]
+        if _APY_UNITS and self._voSet:
+            return units.Quantity(out*self._vo,unit=units.km/units.s)
+        else:
+            return out
+
     @actionAngle_physical_input
     @physical_conversion('phasespacedensityvelocity2',pop=True)
     def pvR(self,vR,R,z,gl=True,ngl=_DEFAULTNGL2,nsigma=4.,vTmax=1.5):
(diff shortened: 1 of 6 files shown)


jobovy:
cc @jmackereth.

samuelwong100:
@jobovy  For your second high-level comment, if the input zs are all positive, then it isn't right to set z_min=0. So I plan to just work with the absolute value of zs, then the grid will naturally start somewhere close to z_min=0 if zs span both positive and negative numbers.

jobovy:
I'm having a hard time imaging a use-case where you wouldn't want to include stars near enough to ``z=0`` that you wouldn't just want to set ``zmin=0``. Then you could also just remove the option ``zmin=None`` and avoid people setting things like ``zmin=-1``, which would be annoying to deal with.

jobovy:
@samuelwong100---let me know when this is ready for further review.

samuelwong100:
@jobovy For sure. I finished all your recommendations. Working on a problem with the unit currently.

jobovy:
This looks pretty close at this point (modulo the units issue, but that should be easy to resolve). Two more comments on the latest edits:
* Great that you now store the grid for re-use! However, this is not covered by the tests I think. Could you add a test where the grid gets re-used?
* I think the way the outliers are determined is now incorrect: you now first determine outliers based on (mean,std.dev.), but then build the grid based on ``R_min``, etc. if they are given, which means that if ``R_min`` etc. are wider than those obtained from (mean,std.dev.) there are outliers that are within the grid. So I think you need to determine the outliers *after* determining the grid edges as those points that fall outside of the grid.

galpy/df/quasiisothermaldf.py
```diff
@@ -1685,6 +1688,194 @@ def sampleV(self,R,z,n=1):
...
+        if R_max is None:
+            R_max_temp= numpy.max(R)
+        else:
+            R_max_temp= R_max
+        if z_max is None:
+            z_max_temp= numpy.max(z)
+        else:
+            z_max_temp= z_max
+        mask= numpy.any([numpy.abs(R - mean_R) > num_std*std_R, 
+                       numpy.abs(z - mean_z) > num_std*std_z,
```
jobovy: I don't understand why you keep the ``numpy.abs(R - mean_R) > num_std*std_R`` and ``numpy.abs(z - mean_z) > num_std*std_z`` conditions here. Why not do something like (note that I also just include the mean calc. here, so that makes a few of the earlier lines unnecessary)
```
if R_min is None:
    R_min= numpy.mean(R) - num_std*numpy.std(R)
if R_max is None:
    R_max= numpy.mean(R) + num_std*numpy.std(R)
if z_max is None:
    z_max= numpy.mean(z) + num_std*numpy.std(z)
mask= numpy.any([R < R_min, R > R_max, z > z_max],axis = 0)
```
and then just keep ``R_min`` etc. to build the grid. I think you're worried that ``R_min`` might be much less than the actual minimum R of the normal points? But I think we should just ignore that possibility.

galpy/df/quasiisothermaldf.py
```diff
@@ -1685,6 +1688,194 @@ def sampleV(self,R,z,n=1):
...
+        outliers_z= z[mask]
+        normal_R= R[~mask]
+        normal_z= z[~mask]
+        #Sample the velocity of outliers directly (without interpolation)
+        outlier_coord_v= numpy.empty((outliers_R.size, 3))
+        for i in range(outliers_R.size):
+            outlier_coord_v[i]= self.sampleV(outliers_R[i], outliers_z[i])[0]
+        #Prepare for optimizing maxVT on a grid
+        #Edges of grid determined by input, unless prespecified by users
+        if R_min is None:
```
jobovy: Remove this re-determination of the grid

galpy/df/quasiisothermaldf.py
```diff
@@ -1685,6 +1688,194 @@ def sampleV(self,R,z,n=1):
...
+            z_max= numpy.max(normal_z)
+        z_min= 0. #Always start grid at z=0 for stars close to plane
+        #Get the new hash of the parameters of grid
+        new_hash= hashlib.md5(numpy.array([R_min,R_max,z_max,R_pixel,z_pixel])).hexdigest()
+        #Reuse old interpolated object if new hash matches the old one
+        if new_hash == self._maxVT_hash:
+            ip_max_vT= self._maxVT_ip
+        #Generate a new interpolation object if different from before
+        else:
+            R_number= int((R_max - R_min)/R_pixel)
```
jobovy: Should probably first do ``floor`` or ``ceil`` here, before turning into an int (sorry for not spotting before).

galpy/df/quasiisothermaldf.py
```diff
@@ -1685,6 +1688,194 @@ def sampleV(self,R,z,n=1):
...
+        normal_coord_v= self.sampleV_preoptimized(normal_R,normal_z,normal_max_vT)
+        #Combine normal and outlier result, preserving original order
+        coord_v[mask]= outlier_coord_v
+        coord_v[~mask]= normal_coord_v
+        if _APY_UNITS and self._voSet:
+            return units.Quantity(coord_v,unit=units.km/units.s)
+        else:
+            return coord_v
+
+    def sampleV_preoptimized(self,R,z,maxVT):
```
jobovy: Should this function be an internal function, starting with ``_``? Then we could remove the units (see comment below).

galpy/df/quasiisothermaldf.py
```diff
@@ -1685,6 +1688,194 @@ def sampleV(self,R,z,n=1):
...
+            vT_accept= propvT[accept_indx]
+            vz_accept= propvz[accept_indx]
+            #Get the indexing of rows of output array that need to be updated
+            #with newly accepted velocity
+            to_change= numpy.copy(remain_indx)
+            to_change[remain_indx]= accept_indx
+            out[to_change]= numpy.stack((vR_accept,vT_accept,vz_accept), axis = 1)
+            #Removing accepted sampled from remain index
+            remain_indx[remain_indx]= ~accept_indx
+        if _APY_UNITS and self._voSet:
```
jobovy: You apply the units here, and then the units get applied again in the ``sampleV_interpolate`` function I think, so in the end the units are (km/s)**2? Making this an internal function would help: then we can just not apply units here and the units only get applied once, at the end of ``sampleV_interpolate``. Although, you seem to have a test that covers this, so maybe I'm misreading the code...

jobovy: Okay, thanks for the clarification. Could you
* explicitly strip off the units as ``coord_v[mask]= outlier_coord_v.to(units.km/units.s).value`` (and similar), so it's clear that this happens and that things are in km/s
* Maybe add a note near line 1811 explaining this? I'm worried that without such a note, this code will be difficult to comprehend.

galpy/df/quasiisothermaldf.py
samuelwong100: Yes, I am worried about that. Right now, `num_std` is set to a default value of 3. So most of the time, there are no outliers, and the maximum of normal points is usually smaller than `numpy.mean(R) + num_std*numpy.std(R)`. So I fear this assumption would slow it down.

galpy/df/quasiisothermaldf.py
samuelwong100: That's true. `ceil` is better since `kx` and `ky` will be less likely to become zero (and give error) in cases where the pixels are too large.

samuelwong100: @jobovy Speaking of that, should I raise an error if pixel is too large that kx or ky become zero?

galpy/df/quasiisothermaldf.py
jobovy: Okay, but then maybe just do
```
R_min= numpy.amax([numpy.mean(R)-num_std*numpy.std(R),numpy.amin(R)])
```
and similar?
and similar?

galpy/df/quasiisothermaldf.py
jobovy: ``R_number`` should always be at least 2, no? Even if ``R_max-R_min`` = ``R_pixel``, no?

galpy/df/quasiisothermaldf.py
samuelwong100: @jobovy I found out that unlike `normal_coord_v`, the unit of `outlier_coord_v` get stripped off earlier, when we assign it the result of `qdf.sampleV`. Should I explicitly strip off the units on line 1761 as well?

jobovy: Yes, I just think it's clearer what does and does not have units if you explicitly strip the units when they are supposed to get removed.

PR #355, opened 2018-10-12
Implement general potential framework for MovingObjectPotential

jamesmlane:
MovingObjectPotential now takes a `pot` keyword, which could be a `galpy` `Potential` object or a list of such objects. Generalized methods for evaluating the potential, forces, and the density are implemented. The `amp`, `GM`, `softening`, `softening_model`, and `softening_length` keywords are no longer accepted, and the user is warned if they are supplied.

`tests/test_potential.py` has been updated to reflect the new class. `mockMovingObjectPotential` now explicitly creates a Plummer potential, and `mockMovingObjectExplSoftPotential` has been removed.

### DIFF
--- a/galpy/potential/MovingObjectPotential.py
+++ b/galpy/potential/MovingObjectPotential.py
@@ -1,36 +1,28 @@
 ###############################################################################
 #   MovingObjectPotential.py: class that implements the potential coming from
 #                             a moving object
-#                                                           GM
-#                              phi(R,z) = -  ---------------------------------
-#                                                        distance
 ###############################################################################
 import copy
 import numpy as nu
-from .Potential import Potential, _APY_LOADED
+import warnings
+from .Potential import Potential, _APY_LOADED, _isNonAxi, flatten, \
+    evaluatePotentials, evaluateRforces, evaluatezforces, evaluateDensities, \
+    evaluatephiforces
+from .PlummerPotential import PlummerPotential
 if _APY_LOADED:
     from astropy import units
 from .ForceSoftening import PlummerSoftening
+from galpy.util import galpyWarning
 class MovingObjectPotential(Potential):
-    """Class that implements the potential coming from a moving object
-
-    .. math::
-
-        \\Phi(R,z,\\phi,t) = -\\mathrm{amp}\\,GM\\,S(d)
-
-    where :math:`d` is the distance between :math:`(R,z,\\phi)` and the moving object at time :math:`t` and :math:`S(\\cdot)` is a softening kernel. In the case of Plummer softening, this kernel is
-
-    .. math::
-
-        S(d) = \\frac{1}{\\sqrt{d^2+\\mathrm{softening\_length}^2}}
-
-    Plummer is currently the only implemented softening.
-
     """
-    def __init__(self,orbit,amp=1.,GM=.06,
+    Class that implements the potential coming from a moving object by combining
+    any galpy potential with an integrated galpy orbit.
+    """
+    def __init__(self,orbit,pot=None,
                  ro=None,vo=None,
-                 softening=None,
-                 softening_model='plummer',softening_length=0.01):
+                 amp=None, GM=None, softening=None,
+                 softening_model=None, softening_length=None
+                 ):
         """
         NAME:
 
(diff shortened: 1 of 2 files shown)


galpy/potential/MovingObjectPotential.py
```diff
@@ -67,17 +49,34 @@ def __init__(self,orbit,amp=1.,GM=.06,
            2011-04-10 - Started - Bovy (NYU)
 
         """
-        Potential.__init__(self,amp=amp*GM,ro=ro,vo=vo,amp_units='mass')
-        if _APY_LOADED and isinstance(softening_length,units.Quantity):
-            softening_length= softening_length.to(units.kpc).value/self._ro
-        # Make sure we aren't getting physical outputs
+
+        Potential.__init__(self,ro=ro,vo=vo)
```
jobovy: This should set ``amp=amp`` as well.

galpy/potential/MovingObjectPotential.py
```diff
@@ -67,17 +49,34 @@ def __init__(self,orbit,amp=1.,GM=.06,
...
+        isList = isinstance(pot,list)
```
jobovy: I don't think there is a real need for ``isList``. I would do:
* if pot is None: # setup Plummer potential
* else:
    pot= flatten(pot)
    if _isNonAxi(pot): # error...
    self._pot= pot # No need to copy I think, because it shouldn't get changed anyway

galpy/potential/MovingObjectPotential.py
```diff
@@ -1,36 +1,28 @@
...
-
-    .. math::
-
-        S(d) = \\frac{1}{\\sqrt{d^2+\\mathrm{softening\_length}^2}}
-
-    Plummer is currently the only implemented softening.
-
     """
-    def __init__(self,orbit,amp=1.,GM=.06,
+    Class that implements the potential coming from a moving object by combining
```
jobovy: Keep the defaults from before (currently, there is no reasonable default): ``GM=0.06``, ``softening_length=0.01``; I think we should remove ``softening_model``.

galpy/potential/MovingObjectPotential.py
```diff
@@ -67,17 +49,34 @@ def __init__(self,orbit,amp=1.,GM=.06,
...
+                self._pot = copy.deepcopy(pot)
+            else:
+                raise RuntimeError("List input to 'MovingObjectPotential' must be of Potential-istances")
+        elif isinstance(pot,Potential):
+            if pot.isNonAxi:
+                raise NotImplementedError('MovingObjectPotential for non-axisymmetric potentials is not currently supported')
+            self._pot = copy.deepcopy(pot)
+        elif pot == None:
+            # Initialize a Plummer potential by default for comptatability & tests
+            pot = PlummerPotential(amp=1.0,b=1.0,ro=ro,vo=vo)
```
jobovy: This needs to have ``amp=GM``, ``b=softening_length`` so these get setup correctly. Not doing this is probably what makes the test in ``test_quantity`` fail on travis.

galpy/potential/MovingObjectPotential.py
```diff
@@ -67,17 +49,34 @@ def __init__(self,orbit,amp=1.,GM=.06,
...
+            self._pot = copy.deepcopy(pot)
+        else:
+            raise RuntimeError("Input to 'MovingObjectPotential' is neither a Potential-instance or a list of such instances")
+        # Warn the user if they supplied deprecated keywords
+        if ( (amp!=None) or (GM!=None) or (softening!=None) or
```
jobovy: I don't think it's necessary, because we will still support the deprecated keywords.

tests/test_potential.py
```diff
@@ -3285,33 +3285,16 @@ def __init__(self,rc=0.75,maxt=1.,nt=50):
...
+        self._o1p= potential.MovingObjectPotential(o1, pot=oplum)
+        self._o2p= potential.MovingObjectPotential(o2, pot=oplum)
         testMWPotential.__init__(self,[self._o1p,self._o2p])
         self.isNonAxi= True
         return None
     def phi2deriv(self,R,z,phi=0.,t=0.):
         raise AttributeError
     def OmegaP(self):
         return 1./self._rc
-from galpy.potential.ForceSoftening import PlummerSoftening
```
jobovy: I don't think this test should be removed, because it should now just explicitly set ``pot=Plummer...`` to mimic what what was going on before.

Issue #357, opened 2018-10-28
Orbit integration with non-existing integration methods should raise error


jobovy:
As reported by Rendy Darma, the following raises a difficult to understand error message, but the issue is that ``method='rk4'`` does not exist:
```
from galpy.potential import *
from galpy.orbit import Orbit
import numpy as np
o = Orbit(vxvv = [1.0, 0.1, 0.1, 0.5, 0.1, 0.0])
t = np.arange(0.0, 10.0, 0.001)
o.integrate(t, MWPotential2014, method = 'rk4')
```

``Orbit.integrate`` should check that the ``method=`` input is an existing method (probably in the high-level call, to make it easier to maintain when new methods are added).

PR #359, opened 2018-11-05
Add validation and test for Orbit.integrate method

smoh:
This is a simple fix to #357 that just adds validation in `Orbit.integrate` method.

- All the method in the docstring of `Orbit.integrate` seems to be available for all individual Orbit classes (that populates `_orb`) so it seems reasonable to add validation at the top-level `Orbit` that is exposed to users.
- On another note, the docstrings for `integrate` methods in individual orbit subclasses do not seem consistent and accurate (some methods available are not mentioned). I know this is technically not exposed in package documentation but just thought I'd mention it.


smoh:
Oops, sorry a typo messed up all tests; fixed now (I did `git commit --amend` as it needn't be a separate one).

PR #362, opened 2018-11-24
Handle ctrl-c on windows too

henrysky:
I have added the functionality to handle ctrl-c on windows

It works fine for Visual Studio 2017 15.9.2 on my Windows 10 v1809 x64 with Python 3.7 which leads to `KeyboardInterrupt: Orbit integration interrupted by CTRL-C (SIGINT)` on windows while pressing CTRL-C during integration with C code

### DIFF
--- a/galpy/util/bovy_rk.c
+++ b/galpy/util/bovy_rk.c
@@ -98,14 +98,7 @@ void bovy_rk4(void (*func)(double t, double *q, double *a,
   action.sa_handler= handle_sigint;
   sigaction(SIGINT,&action,NULL);
 #else
-    if (SetConsoleCtrlHandler(CtrlHandler, TRUE))
-    {}
-    else
-    {
-        // for whatever reason cant install handler, print to warn
-        printf("\nERROR: Could not set control handler");
-        return 1;
-    }
+    if (SetConsoleCtrlHandler(CtrlHandler, TRUE)) {}
 #endif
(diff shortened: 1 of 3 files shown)


jobovy:
Thanks, this looks great! Few questions:

* You now load ``signal.h`` (by removing the enclosing ``#ifndef _WIN32`` etc. from those statements), but is ``signal.h`` really necessary? From, for example, [this help page](https://docs.microsoft.com/en-us/windows/console/registering-a-control-handler-function), it seems like it's not (I'm worried about this library not existing on some people's Windows machines). Or maybe this is where the handler functions are actually defined?
* I would rather not print the error (``printf("\nERROR: Could not set control handler");``) and return when the handler cannot be activated, because the code can still run (it just won't be able to be interrupted). So maybe change all of those statements to just ``if (SetConsoleCtrlHandler(CtrlHandler, TRUE)) {}``

henrysky:
> but is signal.h really necessary? From, for example, this help page, it seems like it's not
 
Yes that control handler does not require `signal.h` but it is required for the variable `interrupted` as it is type of `sig_atomic_t`

Compiled galpy wheels work fine on a fresh Windows 10 with only anaconda installed. `signal.h` will only be an issue for people who want to compile galpy on Windows. But `signal.h` comes with Windows SDK which is installed along with Visual Studio by default.

jobovy:
Ok, wonderful, all looks good now. Thanks a lot for this!

PR #363, opened 2018-11-27
Add python and C implementation of dop853

henrysky:
This PR adds python and C implementation of DOP853 integrator to galpy and test. DOP853 is an explicit Runge-Kutta method of order 8(5,3) by Dormand & Prince with dense output.

Here is a simple testing I have done so far to see the performance and accuracy of `leapfrog` and `dop853` in galpy

```
from galpy.potential import MWPotential2014
from galpy.orbit import Orbit
import numpy as np
import time
from astropy import units as u

# Orbit Integration using galpy for the Sun
op = Orbit([0., 0., 0., 0., 0., 0.], radec=True, ro=8., vo=220.)
ts = np.linspace(0, 20, 10000) * u.Gyr
start_time = time.time()
op.integrate(ts, MWPotential2014, method="method_here")
print(time.time() - start_time, "s")
E = op.E(ts)
print("Energy accurancy: ", np.std(E)/np.fabs(np.median(E)))
```

For `leapfrog`:
```
27.408s
Energy accurancy: 4.894e-09
```

For `dop853`:
```
8.446s
Energy accurancy: 1.361e-13
```

For `dop853_c`:
```
0.082s
Energy accurancy: 1.177e-12
```

For `dopr54_c`:
```
0.199s
Energy accurancy: 2.625e-10
```

For `symplec4_c`:
```
0.220s
Energy accurancy: 4.528e-12
```

### DIFF
--- a/galpy/orbit/Orbit.py
+++ b/galpy/orbit/Orbit.py
@@ -464,7 +464,7 @@ class Orbit(object):
 
            method= 'odeint' for scipy's odeint
                    'leapfrog' for a simple leapfrog implementation
-                   'dop853' for a DOP8(5, 3) implementation
+                   'dop853' for a Dormand-Prince 8(5,3) implementation
                    'leapfrog_c' for a simple leapfrog implementation in C
                    'symplec4_c' for a 4th order symplectic integrator in C
                    'symplec6_c' for a 6th order symplectic integrator in C
--- a/galpy/orbit/Orbit.py
+++ b/galpy/orbit/Orbit.py
@@ -464,6 +464,7 @@ class Orbit(object):
 
            method= 'odeint' for scipy's odeint
                    'leapfrog' for a simple leapfrog implementation
+                   'dop853' for a DOP8(5, 3) implementation
                    'leapfrog_c' for a simple leapfrog implementation in C
                    'symplec4_c' for a 4th order symplectic integrator in C
                    'symplec6_c' for a 6th order symplectic integrator in C
@@ -484,7 +485,7 @@ class Orbit(object):
            2015-06-28 - Added dt keyword - Bovy (IAS)
 
         """
-        if method.lower() not in ['odeint', 'leapfrog', 'leapfrog_c',
+        if method.lower() not in ['odeint', 'leapfrog', 'dop853', 'leapfrog_c',
                 'symplec4_c', 'symplec6_c', 'rk4_c', 'rk6_c',
                 'dopr54_c']:
             raise ValueError('{:s} is not a valid `method`'.format(method))
--- a/galpy/orbit/Orbit.py
+++ b/galpy/orbit/Orbit.py
@@ -470,7 +470,8 @@ class Orbit(object):
                    'symplec6_c' for a 6th order symplectic integrator in C
                    'rk4_c' for a 4th-order Runge-Kutta integrator in C
                    'rk6_c' for a 6-th order Runge-Kutta integrator in C
-                   'dopr54_c' for a Dormand-Prince integrator in C (generally the fastest)
+                   'dopr54_c' for a Dormand-Prince 5(4) integrator in C (generally the fastest)
+                   'dop853_c' for a Dormand-Prince 8(5, 3) integrator in C
 
            dt= (None) if set, force the integrator to use this basic stepsize; must be an integer divisor of output stepsize (only works for the C integrators that use a fixed stepsize) (can be Quantity)
 
@@ -487,7 +488,7 @@ class Orbit(object):
         """
         if method.lower() not in ['odeint', 'leapfrog', 'dop853', 'leapfrog_c',
                 'symplec4_c', 'symplec6_c', 'rk4_c', 'rk6_c',
-                'dopr54_c']:
+                'dopr54_c', 'dop853_c']:
             raise ValueError('{:s} is not a valid `method`'.format(method))
         pot= flatten_potential(pot)
         _check_potential_dim(self,pot)
--- a/galpy/orbit/RZOrbit.py
+++ b/galpy/orbit/RZOrbit.py
@@ -484,7 +484,7 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
     if method.lower() == 'leapfrog' \
             or method.lower() == 'leapfrog_c' or method.lower() == 'rk4_c' \
             or method.lower() == 'rk6_c' or method.lower() == 'symplec4_c' \
-            or method.lower() == 'symplec6_c' or method.lower() == 'dopr54_c':
+            or method.lower() == 'symplec6_c' or method.lower() == 'dopr54_c' or method.lower() == 'dop853_c':
         #We hack this by upgrading to a FullOrbit
         this_vxvv= nu.zeros(len(vxvv)+1)
         this_vxvv[0:len(vxvv)]= vxvv
--- a/galpy/orbit/RZOrbit.py
+++ b/galpy/orbit/RZOrbit.py
@@ -495,7 +495,7 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
         l= vxvv[0]*vxvv[2]
         l2= l**2.
         init= [vxvv[0],vxvv[1],vxvv[3],vxvv[4]]
-        if method.lower == "dop853":
+        if method.lower() == "dop853":
             intOut = dop853(_RZEOM, init, t, args=(pot, l2))
         else:
             intOut = integrate.odeint(_RZEOM, init, t, args=(pot, l2),
--- a/galpy/orbit/RZOrbit.py
+++ b/galpy/orbit/RZOrbit.py
@@ -480,7 +480,7 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
                 warnings.warn("Cannot use C integration because C extension not loaded (using %s instead)" % (method), galpyWarning)
             else:
                 warnings.warn("Cannot use C integration because some of the potentials are not implemented in C (using %s instead)" % (method), galpyWarning)
-    if method.lower() == 'leapfrog' \
+    if method.lower() == 'leapfrog' or method.lower() == 'dop853' \
             or method.lower() == 'leapfrog_c' or method.lower() == 'rk4_c' \
             or method.lower() == 'rk6_c' or method.lower() == 'symplec4_c' \
             or method.lower() == 'symplec6_c' or method.lower() == 'dopr54_c':
--- a/galpy/orbit/RZOrbit.py
+++ b/galpy/orbit/RZOrbit.py
@@ -10,6 +10,7 @@ import galpy.util.bovy_symplecticode as symplecticode
 from .FullOrbit import _integrateFullOrbit
 from .integrateFullOrbit import _ext_loaded as ext_loaded
 from galpy.util.bovy_conversion import physical_conversion
+from galpy.util.leung_dop853 import dop853
 from .OrbitTop import OrbitTop
 class RZOrbit(OrbitTop):
     """Class that holds and integrates orbits in axisymetric potentials 
@@ -480,7 +481,7 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
                 warnings.warn("Cannot use C integration because C extension not loaded (using %s instead)" % (method), galpyWarning)
             else:
                 warnings.warn("Cannot use C integration because some of the potentials are not implemented in C (using %s instead)" % (method), galpyWarning)
-    if method.lower() == 'leapfrog' or method.lower() == 'dop853' \
+    if method.lower() == 'leapfrog' \
             or method.lower() == 'leapfrog_c' or method.lower() == 'rk4_c' \
             or method.lower() == 'rk6_c' or method.lower() == 'symplec4_c' \
             or method.lower() == 'symplec6_c' or method.lower() == 'dopr54_c':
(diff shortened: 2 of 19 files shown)


tests/test_jeans.py
```diff
@@ -47,7 +47,7 @@ def test_sigmar_wlog_constbeta_asbetafunc():
 
 def test_sigmar_wlog_linbeta():
     # for log halo, dens ~ r^-gamma, and beta = -b x r --> 
-    # sigmar = vc sqrt( scipy.special.gamma(-gamma)*scipy.special.gammaincc(-gamma,2*b*r)/[(2*b*r)**-gamma*exp(-2*b*r)]
+    # sigmar = vc sqrt( scipy.misc.gamma(-gamma)*scipy.misc.gammaincc(-gamma,2*b*r)/[(2*b*r)**-gamma*exp(-2*b*r)]
```
jobovy: Still one left here...

jobovy: although this strangely enough didn't fail in the tests...

jobovy:
``travis`` python 2.7 builds are failing with
```
from galpy.util.leung_dop853 import dop853
     File "/home/travis/build/jobovy/galpy/galpy/util/leung_dop853.py", line 467
       return np.concatenate((x[len(x) // 2:], func(x[:len(x) // 2], *args, t)))
   SyntaxError: only named arguments may follow *expression
```

which seems to work in python 3, but not in python 2.7.

tests/test_orbit.py
```diff
@@ -582,6 +584,7 @@ def test_energy_symplec_longterm():
     #Basic parameters for the test
     times= numpy.linspace(0.,10000.,100001) #~360 Gyr at the Solar circle
     integrators= ['leapfrog_c', #don't do leapfrog, because it takes too long
+                  'dop853',
```
henrysky: should `dop853` be added here? it passes the test but it is not symplectic which is this test is about???

jobovy: Good that it passes the test! But if it's not symplectic, there should be some long-term energy error growth (do ``o.plotE()``), so it shouldn't be in this test (which is for symplectic integrators).

galpy/orbit/planarOrbit.py
```diff
@@ -506,7 +507,7 @@ def _integrateROrbit(vxvv,pot,t,method,dt):
                 warnings.warn("Cannot use C integration because C extension not loaded (using %s instead)" % (method), galpyWarning)
             else:
                 warnings.warn("Cannot use C integration because some of the potentials are not implemented in C (using %s instead)" % (method), galpyWarning)
-    if method.lower() == 'leapfrog':
+    if method.lower() == 'leapfrog' or method.lower() == 'dop853':
```
jobovy: This ``dop853`` should be moved to the ``odeint`` elif statement

galpy/orbit/Orbit.py
```diff
@@ -464,6 +464,7 @@ def integrate(self,t,pot,method='symplec4_c',dt=None):
 
            method= 'odeint' for scipy's odeint
                    'leapfrog' for a simple leapfrog implementation
+                   'dop853' for a DOP8(5, 3) implementation
```
jobovy: Make this 'for a Dormand-Prince 8(5,3) implementation'

galpy/orbit/RZOrbit.py
```diff
@@ -480,7 +480,7 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
                 warnings.warn("Cannot use C integration because C extension not loaded (using %s instead)" % (method), galpyWarning)
             else:
                 warnings.warn("Cannot use C integration because some of the potentials are not implemented in C (using %s instead)" % (method), galpyWarning)
-    if method.lower() == 'leapfrog' \
+    if method.lower() == 'leapfrog' or method.lower() == 'dop853' \
```
jobovy: Should move the ``dop853`` case to the elif case for ``odeint`` below (with Dormand-Prince, we can use the ``_RZEOM`` equation of motion)

galpy/util/leung_dop853.h
```diff
@@ -0,0 +1,92 @@
+/*
+C implementations of symplectic integrators
```
jobovy: This should be edited

galpy/util/leung_dop853.c
```diff
@@ -0,0 +1,517 @@
+/*
+C implementations of symplectic integrators
```
jobovy: This should be edited

jobovy:
Started testing this, looking at how the actual orbit (like ``o.R(ts)``) compares to other integrators rather than just energy. It seems like there's something wrong with the C implementation, because the radius has large deviations compared to other integrators, which the python implementation does not have. Example:
```
import numpy
from galpy.potential import MWPotential2014
from galpy.orbit import Orbit
%pylab inline
o= Orbit()
ts= numpy.linspace(0.,100.,1001)
o.integrate(ts,MWPotential2014,method='dop853')
od= o()
od.integrate(ts,MWPotential2014,method='dop853_c')
os= o()
os.integrate(ts,MWPotential2014,method='symplec4_c')
plot(ts,od.R(ts,use_physical=False)-os.R(ts,use_physical=False),label='dop853_c - symplec4_c')
plot(ts,o.R(ts,use_physical=False)-os.R(ts,use_physical=False),label='dop853 - symplec4_c')
xlabel(r'$t$')
ylabel(r'$\Delta R$')
legend();
```
gives the following:
screen shot 2018-12-02 at 4 34 48 pm

The ``appveyor`` build also fails because the orbit tests time out (they take an hour, used to take 25 minutes), which happened twice, so likely something to do with the new C code (travis is fine though...).

henrysky:
> Started testing this, looking at how the actual orbit (like `o.R(ts)`) compares to other integrators rather than just energy. It seems like there's something wrong with the C implementation, because the radius has large deviations compared to other integrators, which the python implementation does not have. Example:

I gave dense output in c implementation the wrong time which leads to wrong result. Now it behaves exactly like `dop853` python implementation. I also changed `rtol` and `atol` from 1e-14 to whatever galpy gives. i did not know that is exponential but fixed now
download

> The `appveyor` build also fails because the orbit tests time out (they take an hour, used to take 25 minutes), which happened twice, so likely something to do with the new C code (travis is fine though...).

SIGINT test gets stuck. It happens on my Windows too. Seems like the issue is python cannot get exit code correctly on Windows. CTRL+C works in practice tho. So I removed them from test for now until I figure it out whats the deal on Windows.

galpy/orbit/RZOrbit.py
```diff
@@ -490,12 +491,15 @@ def _integrateRZOrbit(vxvv,pot,t,method,dt):
...
         #tmp_out is (nt,6)
         out= tmp_out[:,0:5]
-    elif method.lower() == 'odeint':
+    elif method.lower() == 'odeint' or method.lower() == 'dop853':
         l= vxvv[0]*vxvv[2]
         l2= l**2.
         init= [vxvv[0],vxvv[1],vxvv[3],vxvv[4]]
-        intOut= integrate.odeint(_RZEOM,init,t,args=(pot,l2),
-                                 rtol=10.**-8.)#,mxstep=100000000)
+        if method.lower == "dop853":
```
jobovy: parentheses after ``method.lower`` are missing here

jobovy:
Great, that test seems to pass now!

You should also add ``dop853_c`` to the 'liouville' test in ``test_orbit``.

There are a few lines in ``leung_dop853.c`` not being touched by the tests now:
* There are two interruption blocks? Because there is only a single interrupt in the tests only one of these gets hit. Maybe you can remove one? We don't care that the interruption happens immediately, so perhaps it can just be in the outer loop?
* A special case, probably difficult to hit with a test...

jobovy:
Started looking at this again and this all looks quite good to me now (I also did an explicit test of the Python DOP853 implementation with a differential equation with an analytical solution, which passed). But I did one more tests that gives somewhat odd results (although may be nothing...):
```
import numpy
from galpy.potential import MWPotential2014
from galpy.orbit import Orbit
%pylab inline
o= Orbit([1.3,-0.5,1.,0.])
ts= numpy.linspace(0.,300.,1001)
o.integrate(ts,[MWPotential2014],method='dop853')
od= o()
od.integrate(ts,[MWPotential2014],method='dop853_c')
od.plotE()
o.plotE(overplot=True)
```
gives
dop853-test2
Apologies for the lack of labels, but the green line is the Python implementation of DOP853 and the blue line the C implementation: the Python version has a much more stable energy error. Do you understand why this is?

Finally, could you add an entry in HISTORY.txt about this addition and add yourself in the ``Direct contributions to the code base:`` section of AUTHORS.txt about this as well?

henrysky:
I found a bug that the time step is being reset to `hmax` and the steps are rejected and scaled down until acceptable error. `dop853_c` is now a bit faster due to time step not being reset to `hmax` in every loop. But this does not seems to change the result. Still not sure why C implementation seems worse than python implementation

jobovy:
If anything, with this latest change things seem a little worse still (but again, these are still very small errors):
download

henrysky:
I try to trace every variable and found python and C version differ even in very early in initial step estimation.

And I check with the initial condition array given by galpy, the initial condition array in c-implementation is `[1.3        0.        0.5         1.]` and in python is `[1.3        0.5        0.         0.76923077]`. Is it supposed to happen??

jobovy:
Yes, that's expected. In C the equation of motion is integrated in rectangular coordinates, so the initial conditions are ``x,y,vx,vy`` (I think you mean ``-0.5``); in Python the equation of motion is integrated in cylindrical coordinates and the initial conditions are ``R,vR,phi,vT/R``.

jobovy:
But this is a good point and I think the difference in the equations of motion might explain the difference, because the potential is axisymmetric and angular momentum is conserved, integrating the system in cylindrical coordinates is more stable. So there is likely nothing wrong with the DOP853 integrator!

henrysky:
I thought the C version also integrate in different coordinate between symplectic and non-symplectic like the python one you told me? Because I see the calls to C integrators are different between RK/DOP54 and symplectic integrators. I do find lowering the error tolerance of `dop853_c` to something like 1e-14 can bring it much closer to the python implementation and I do think the python and c implementation is working the same way. I cant find any difference anymore.

henrysky:
The cause is clearly coordinates now. I use my standalone python implementation of `dop853` in solar system. The integration using rectangular and polar coordinates show the exact same trend.

```python
import pylab as plt
import numpy as np
from leung_dop853 import dop853
from scipy.integrate import odeint


G = 4*np.pi**2 # define G for convenience.

# Set initial position and speed of satellite.
# Compute and display some values for a circular orbit to help decide on initial conditions.
Ro = 1.0    # Radius of the orbit 
M  = 1.0    # mass of the central mass

v = np.sqrt(G*M/Ro) # v assuming circular orbit

# Set initial conditions and define needed array.
x0 = 1.
vy0 = 5.
t0 = 0.
tf = 10.
tau = 1e-4

X0 = [x0, 0, 0, vy0]       # set initial state of the system
t = np.arange(t0,tf,tau)   # create time array starting at t0, ending at tf with a spacing tau

# Define function to return f(X,t)
def f_func(state, time):
    f = np.zeros(4)
    f[0] = state[2]
    f[1] = state[3]
    r = np.sqrt(state[0]**2 + state[1]**2)
    f[2] = -(G*M*state[0])/r**3
    f[3] = -(G*M*state[1])/r**3
    return f

X_dop853 = dop853(func=f_func, x=X0, t=t, rtol=1e-12)
x_dop853 = X_dop853[:,0]  # Giving a ':' as the index specifies all of the elements for that index so
y_dop853 = X_dop853[:,1]  # x, y, vx, and vy are arrays that specify their 
vx_dop853 = X_dop853[:,2] # values at any given time index
vy_dop853 = X_dop853[:,3]

X0 = [x0, 0, 0, vy0]       # set initial state of the system, same as retengular but in [r, phi, vr, vphi]

# Define function to return f(X,t) in polar
def f_func_polar(state, time):
    f = np.zeros(4)
    f[0] = state[2]
    f[1] = state[3]
    f[2] = -(G*M)/state[0]**2 + state[0]*state[3]**2
    f[3] = (-2 * state[2]*state[3])/state[0]
    return f

X_dop853_polar = dop853(func=f_func_polar, x=X0, t=t, rtol=1e-12)

# Plot the results
plt.figure(figsize=(10, 10))
plt.clf()

# calculate KE + PE
eps_dop853 = 0.5*(vx_dop853**2 + vy_dop853**2) - G*M/np.sqrt(x_dop853**2 + y_dop853**2)
eps_dop853_polar = 0.5*(X_dop853_polar[:,2]**2 + X_dop853_polar[:,0]**2*X_dop853_polar[:,3]**2) - G*M/X_dop853_polar[:,0]

# Plot the energy
plt.subplot(2,1,2)
plt.title("Energy Plot", fontsize=15)
plt.plot(t, eps_dop853, label='dop853 in rectangular')
plt.plot(t, eps_dop853_polar, label='dop853 in polar')
# plt.axis([min(t),max(t),eps_plot_min,0])
# plt.ylim((-31.4784, -31.47845))
plt.xlabel('Time', fontsize=15)
plt.ylabel('Normalized Energy', fontsize=15)
plt.legend(loc="best", fontsize=15)
plt.tight_layout()
plt.savefig("standalone_integration.png")
```
standalone_integration

jobovy:
Okay, great, thanks for checking this in detail! I think all is good then and will merge this.

Yes, the integration in C is different for symplectic and non-symplectic integrators, because they use very different algorithms, so they can't be the same. But I think they are always in rectangular coordinates, even though the calls are different.

PR #371, opened 2019-02-23
Better handling of error in orbits function

henrysky:
It just came up to me that I try to estimate eccentricity for the whole APOGEE DR14 with the new `Orbits` class, I get stuck in a orbit which is unbound. And it is very uninformative with just an error msg which orbit is unbound so user cannot debug.

So I modified to:
#. add which index the error msg is thrown so user know which orbit in `Orbits` is causing problem
#. instead of raising error, print warning so the whole operation can be finished
#. have done performance test, the new loops are as fast as the loop `Orbits` currently using (approx 4:29 to 4:31 minutes timed with hands on 140000 APOGEE stars eccentricity estimation on my laptop)
#. added a test to make sure an operation can be done to all orbit even some of them are problematic

### DIFF
--- a/galpy/orbit/Orbits.py
+++ b/galpy/orbit/Orbits.py
@@ -247,11 +247,29 @@ class Orbits(object):
         """
         attribute = getattr(Orbit(), name)
         if callable(attribute):
-            return lambda *args, **kwargs: [
-                getattr(orbit, name)(*args, **kwargs) for orbit in self._orbits
-            ]
+            def __lambda(*args, **kwargs):
+                idx = 0  # use to keep track of indices
+                listt = []
+                for orbit in self._orbits:
+                    try:
+                        listt.append(getattr(orbit, name)(*args, **kwargs))
+                        idx += 1
+                    except Exception as e:
+                        # show informative error on which orbit is problematic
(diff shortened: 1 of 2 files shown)


henrysky:
Oh I thought the `orbit seems to be unbound` error is caused by some high velocity stars. It seems the test failed to cover that and what happen actually is galpy ignored the problematic orbit in the list that returned although the operation is completed

jobovy:
Thanks for this! I'm not going to merge this (or at least not yet), because the plan is to implement all of the ``Orbit`` functions specifically for ``Orbits`` and thus not need the ``__getattr__`` function anymore. In particular, I just implemented the ``e()`` function, so you should be able to compute the eccentricity with that. It might have the same issue though, so might have to deal with this eventually.

henrysky:
> I just implemented the `e()` function, so you should be able to compute the eccentricity with that. It might have the same issue though, so might have to deal with this eventually.

I just tested it. The problematic orbits now return Nan in eccentricity and it is way much faster too!!

jobovy:
I'm going to close this, because all ``Orbit`` functions have now been implemented in ``Orbits``, so this isn't relevant anymore.

PR #373, opened 2019-03-08
SkyCoord('all')

nstarman:
added a string option into Orbits.SkyCoords so that SkyCoords('all') uses SkyCoords(t=self.t).
This just means that an Orbits may be integrated and converted into a SkyCoord over its full integration range without manually re-calling the time
ex:
```python
o = Orbits()
o.integrate(...)
sc = o.SkyCoord('all')  # equivalent to o.SkyCoord(o.time())
```


nstarman:
'full' may be a better key than 'all'
I went with 'all' for brevity, but if SkyCoord is expanded to include slice options for the underlying orbits, then 'all' might connote *all* orbits, rather than the *full* SkyCoords

PR #375, opened 2019-04-06
PowerShell script in docs for Windows

henrysky:
I just realized the newer Anaconda for Windows migrated to PowerShell from CMD now so the compiling on Windows guide is not working straightforward. I have added an equivalent script for PowerShell so people can just copy and paste if using PowerShell.

I also tested to compile galpy with the new MSVC 2019 compiler and things seem fine.


jobovy:
Thanks!

PR #376, opened 2019-04-11
Allow Orbits to be subclassed 

nstarman:
small change to Orbits such that Orbits may be fully subclassed.
Now __getitem__ will return an instance of child, not parent, class.


nstarman:
I don't why the coverage has decreased so much. It has passed all the builds and I ran the full suite of tests on python 2.7 and 3.7.

PR #378, opened 2019-05-16
Warning for r < minr

### LINKED ISSUE #356: Raise warning when r < minr in ChandrasekharDynamicalFrictionForce?
When ``r < minr``, dynamical friction is turned off, which may come as a surprise when looking at orbits at small ``r``. Perhaps raising a warning in this case would be best.

ritzvik:
Fixes #356 
@jobovy

### DIFF
--- a/galpy/potential/ChandrasekharDynamicalFrictionForce.py
+++ b/galpy/potential/ChandrasekharDynamicalFrictionForce.py
@@ -7,7 +7,7 @@ import copy
 import hashlib
 import numpy
 from scipy import special, interpolate
-from galpy.util import bovy_conversion
+from galpy.util import bovy_conversion, galpyWarning
 from .DissipativeForce import DissipativeForce
 from .Potential import _APY_LOADED, evaluateDensities
 from .Potential import flatten as flatten_pot
@@ -185,7 +185,7 @@ class ChandrasekharDynamicalFrictionForce(DissipativeForce):
     def _calc_force(self,R,phi,z,v,t):
         r= numpy.sqrt(R**2.+z**2.)
         if r < self._minr:
-            warnings.warn("Dynamical friction is turned off, as r < minr", RuntimeWarning)
+            warnings.warn("Dynamical friction is turned off, as r < minr", galpyWarning)
             self._cached_force= 0.
         else:
             vs= numpy.sqrt(v[0]**2.+v[1]**2.+v[2]**2.)
--- a/galpy/potential/ChandrasekharDynamicalFrictionForce.py
+++ b/galpy/potential/ChandrasekharDynamicalFrictionForce.py
@@ -2,6 +2,7 @@
 #   ChandrasekharDynamicalFrictionForce: Class that implements the 
 #                                        Chandrasekhar dynamical friction
 ###############################################################################
+import warnings
 import copy
 import hashlib
 import numpy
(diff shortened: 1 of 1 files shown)


jobovy:
Thanks! Few things:

* Please make this a ``galpyWarning`` rather than a ``RuntimeWarning`` (``galpyWarning`` is located in ``galpy.util``)
* I worry that raising the warning during an orbit integration will *significantly* slow it down, even when the warning settings are such that it only gets printed once. Could you time an orbit integration (maybe following the [LMC example](https://galpy.readthedocs.io/en/latest/orbit.html#new-in-v1-4-example-the-orbit-of-the-large-magellanic-cloud-in-the-presence-of-dynamical-friction) in the docs) where ``minr`` is such that for most of the orbit integration ``r < minr`` for the case with and without the warning?
* If raising the warning upon every evaluation does slow things down, we should try to come up with a way to only do it once per orbit integration. How to do that is a little tricky, because we don't want to turn off the warning for _every_ subsequent use of the dynamical friction force....

ritzvik:
@jobovy Thanks for the insights.
  - Regarding the first point raised, it's done. (I should have figured it out myself)
  - Regarding point 2, I couldn't understand anything :( Maybe it requires theoretical knowledge.
  

>
> * If raising the warning upon every evaluation does slow things down, we should try to come up with a way to only do it once per orbit integration. How to do that is a little tricky, because we don't want to turn off the warning for _every_ subsequent use of the dynamical friction force....

  - For this I was thinking maybe we can make new function `_calc_force_with_warning()` and the call this separately at the start of every orbit integration.

Like this:
```
def _calc_force_with_warning(-----):
    r= numpy.sqrt(R**2.+z**2.)
    if r < self._minr:
        warnings.warn("Dynamical friction is turned off, as r < minr", galpyWarning)
    self._calc_force(-------)
```
That is `_calc_force_with_warning()` would be called once, and rest of the time `_calc_force()` can be called. `_calc_force_with_warning()` calls the same `_calc_force()` after emitting the warning.

Regarding the third issue, it would be great if you could point out where the orbit integration is happening.

jobovy:
Something like what you suggest might work, but the issue is how to make sure to just call the ``_calc_force_with_warning`` once (a larger issue is that you might only reach ``r < minr`` after a while in the orbit integration, so you can't decide on the warning at the starting point alone).

Anyway, before embarking down a potentially complicated path, I was just wondering whether you could run and time the original code and your edited code for an orbit  with ``r < minr`` (for example, by using the LMC example and explicitly setting ``minr`` to a value larger than the LMC's r. That way we can see how much slower it is to raise the warning every time the force is evaluated. During orbit integration, the force is evaluated thousands of times, so even if raising the warning takes only a fraction of a second, that will add up to a lot of time over the course of an orbit integration.

jobovy:
@ritzvik: are you still working on this?

ritzvik:
> @ritzvik: are you still working on this?

@jobovy Sorry to keep you hanging, but I am quite down with end-sems, projects, assignments(I procrastinate a lot, things pile up 😢 ). I would be able to touch this after 10th june. Earlier, I tried to run benchmarks but it failed to do so on windows(Some GNU scientific library dependencies, I think). So next time, I would be doing it on ubuntu(Which I had happily kicked out of my system, don't know why)

Again, sorry for the delay!! I would do this.

jobovy:
@ritzvik: Okay, thanks, no problem!

PR #387, opened 2019-06-15
Edited DehnenBarPotential so it can accept arrays and included a new t…

webbjj:
…est to make sure array/float inputs return same potential for r<rb and r> rb

### DIFF
--- a/galpy/potential/DehnenBarPotential.py
+++ b/galpy/potential/DehnenBarPotential.py
@@ -163,14 +163,26 @@ class DehnenBarPotential(Potential):
             smooth= 1.
         r2= R**2.+z**2.
         r= numpy.sqrt(r2)
-        if r <= self._rb:
-            return self._af*smooth*numpy.cos(2.*(phi-self._omegab*t-self._barphi))\
-                *((r/self._rb)**3.-2.)*R**2./r2
+        if isinstance(r,float):
+            if r <= self._rb:
+                return self._af*smooth*numpy.cos(2.*(phi-self._omegab*t-self._barphi))\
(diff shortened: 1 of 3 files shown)


jobovy:
@webbjj ---Looks good! I fixed a slight ``np.`` --> ``numpy.`` issue (so pull before you make any further changes). Could you also add a test that ``SpiralArmsPotential`` works fine for arrays? Also, the tests are still running and test coverage hasn't been determined yet, but I think we'll need similar tests for all ``DehnenBarPotential`` and ``SpiralArmsPotential`` methods, because the changes won't all be touched by the test that you added.

webbjj:
Thanks Jo. Good catch on the np --> numpy.

 I actually didn't mean for the SpiralArmsPotential edits to be part of the
pull request. I pushed it to my forked version of galpy after I initiated
the pull request, not thinking that it would be included. It was probably
final though. So worked out. Will add a similar test as the
DehnenBarPotential test asap. So you think I should do a test for
everything then. Rforce, zforce, etc....? If so Ill add that too.

Jeremy

jobovy:
You could mimic the tests in ``tests/test_potential.py`` : ``test_ExpDisk_special``, which do the same for the ``DoubleExponentialDiskPotential`` I think.

Yes, when you push further commits to the branch that you launched the PR from, they get added to the PR (if you want to avoid this, make a separate branch for the PR before launching it).

Lots of ``SpiralArmsPotential`` tests failed... (see the [report](https://travis-ci.org/jobovy/galpy/builds/546180101?utm_source=github_status&utm_medium=notification)).

webbjj:
Closing this, as a new pull request was created.

PR #388, opened 2019-06-17
Using arrays with DehenBar and SpiralArms

webbjj:
Array tests for both DehenBar and SpiralArms are now in test_potential.py. Removed DehnenBar test that had been implemented in test_quantity.py. Commented out test for array input fail in test_SpiralArmsPotential.py. Changed the isinstance check in SpiralArms from a check if float to check if numpy array, as integer inputs were causing many tests to fail


webbjj:
test_potential.py updated. The DehnenBar_special and SpiralArms_special tests should work now.

jobovy:
@webbjj: I just added a general test of all potentials to see whether they accept array input. In doing so, I noticed that `DehnenBarPotential` still doesn't work when the input time is an array (which it should be if the other coordinates are). This test is ``tests/test_potential.py::test_potential_array_input``. The beginning of this test also has a convenient list of potentials that still do not accept array input and so couldn't be used in AMUSE I think...

webbjj:
So the way DehnenBarPotential is written now, we assume a user has input R,z,phi as single values or arrays. We do not allow or check for combinations like your edits to SpiralArmPotential now do. Are you thinking to allow the time input to be either single value or an array, regardless of the R,z,phi input? This would probably be necessary for AMUSE, as I think only a single time value is sent along with the N-dimensional position arrays.

The most straightforward approach would then to have an if statement at the start of each function for the calculation of smooth, depending on if t is an array or not. And then when we return out, t will have to be repeated if R,z,phi are arrays and t is not. I think t can be an array if R,z,phi are not. Make sense to you?

jobovy:
I think the ``DehnenBarPotential`` currently works for R an array and the others not and even for R a scalar and z or phi an array, because then they just go into the 'R not an array' statement and the expression produces an array if any of z, phi, or smooth is an array. So I think the only thing necessary is to make smooth an array if t is one; if t is not an array, then the current code works?

Might be useful to pull out the smooth part of each function into each own little function
```
def _smooth(self,t):
    ...
```
such that you only have to right the array handling once.

Didn't realize AMUSE doesn't give t as an array, so I guess this isn't necessary for AMUSE.

webbjj:
Looking closer at the various combinations in DehnenBarPotential, I think the only one that won't work will be if one of R or z are an array and phi is not, as there is a phi[indx] inside if isinstance(r,numpy.ndarray):. Want me to add a check for phi too?

jobovy:
I think we can get around that by just pulling everything that doesn't need `indx` out of the assignment to ``out[indx]``, like
```
out[indx]= ((r[indx]/self._rb)**3.-2.)*R[indx]**2./r2[indx]
indx=numpy.invert(indx)
out[indx]= -(self._rb/r[indx])**3.*R[indx]**2./r2[indx]
out*= self._af*smooth*numpy.cos(2.*(phi-self._omegab*t-self._barphi))
```
(which would also fix the issue of t being an array or not).

webbjj:
Ok, new DehnenBarPotential has been pushed that passes both test_potential_array_input and test_DehnenBar_special. _smooth is now its own function that checks for t being an array or not within. I couldn't get around checking that R and z are arrays if r is an array, as they are both called with [indx] in several cases. I started going through on a case by case basis, as sometimes the R or z term can be factored out, but thought this may get confusing.

tests/test_potential.py
```diff
@@ -1020,6 +1020,103 @@ def test_amp_mult_divide():
...
+    pots= [p for p in dir(potential) 
+           if ('Potential' in p and not 'plot' in p and not 'RZTo' in p 
+               and not 'FullTo' in p and not 'toPlanar' in p
+               and not 'evaluate' in p and not 'Wrapper' in p
+               and not 'toVertical' in p)]
+    rmpots= ['Potential','MWPotential','MWPotential2014',
+             'interpRZPotential', 'linearPotential', 'planarAxiPotential',
+             'planarPotential', 'verticalPotential','PotentialError',
+             'EllipsoidalPotential']
+    rmpots.append('DehnenBarPotential')
```
jobovy: Need to remove this to see whether ``DehnenBarPotential`` passes this test....

galpy/potential/DehnenBarPotential.py
```diff
@@ -153,26 +175,17 @@ def _evaluate(self,R,z,phi=0.,t=0.):
...
-                    *((r[indx]/self._rb)**3.-2.)*R[indx]**2./r2[indx]
+            out[indx]= ((r[indx]/self._rb)**3.-2.)*R[indx]**2./r2[indx]
             indx=numpy.invert(indx)
-            out[indx]= -self._af*smooth*numpy.cos(2.*(phi[indx]-self._omegab*t-
-                                                      self._barphi))\
-                                                      *(self._rb/r[indx])**3.\
-                                                      *R[indx]**2./r2[indx] 
+            out[indx]= -(self._rb/r[indx])**3.*R[indx]**2./r2[indx] 
+
+            out*=self._af*smooth[indx]*numpy.cos(2.*(phi[indx]-self._omegab*t[indx]-self._barphi))
```
jobovy: Think you need to drop the ``[indx]`` here

galpy/potential/DehnenBarPotential.py
```diff
@@ -329,28 +330,24 @@ def _zforce(self,R,z,phi=0.,t=0.):
...
                                         -R[indx]**2./r[indx]**6.*(3.*R[indx]**2.+2.*z[indx]**2.))\
                           +4.*z[indx]**2./r[indx]**6.*(4.*R[indx]**2.-r[indx]**2.))
             indx=numpy.invert(indx)
-            out[indx]= self._af*smooth*numpy.cos(2.*(phi[indx]-self._omegab*t-
-                                                     self._barphi))\
-                    *(self._rb/r[indx])**3./r[indx]**6.*((r[indx]**2.-7.*R[indx]**2.)*(3.*R[indx]**2.-2.*z[indx]**2.)\
+            out[indx]= (self._rb/r[indx])**3./r[indx]**6.*((r[indx]**2.-7.*R[indx]**2.)*(3.*R[indx]**2.-2.*z[indx]**2.)\
                                                  +6.*R[indx]**2.*r[indx]**2.)
+
+            out*=self._af*smooth[indx]*numpy.cos(2.*(phi[indx]-self._omegab*t[indx]-self._barphi))
```
jobovy: Drop the ``[indx]`` here

galpy/potential/DehnenBarPotential.py
```diff
@@ -201,26 +214,22 @@ def _Rforce(self,R,z,phi=0.,t=0.):
...
-            xi= 2.*deltat/(self._tsteady-self._tform)-1.
-            smooth= (3./16.*xi**5.-5./8*xi**3.+15./16.*xi+.5)
-        else: #bar is fully on
-            smooth= 1.
+        smooth=_smooth(t)
         r= numpy.sqrt(R**2.+z**2.)
         if isinstance(r,numpy.ndarray):
+            if not isinstance(R,np.ndarray):
+                R=np.repeat(len(r))
+            if not isinstance(z,,np.ndarray):
```
jobovy: Double comma

galpy/potential/DehnenBarPotential.py
```diff
@@ -249,26 +258,22 @@ def _phiforce(self,R,z,phi=0.,t=0.):
...
-            smooth= (3./16.*xi**5.-5./8*xi**3.+15./16.*xi+.5)
-        else: #bar is fully on
-            smooth= 1.
+        smooth=_smooth(t)
         r2= R**2.+z**2.
         r= numpy.sqrt(r2)
         if isinstance(r,numpy.ndarray):
+            if not isinstance(R,np.ndarray):
+                R=np.repeat(len(r))
+            if not isinstance(z,,np.ndarray):
```
jobovy: Double comma (and probably more of them...)

webbjj:
Fixed typos in DehnenBarPotential and made sure DehnenBarPotential is now called in test_potential_array_input

jobovy:
Fixed how ``numpy.repeat`` was used and added a few more tests to touch all different cases. Looks good now, so will merge!

PR #392, opened 2019-07-23
Add named objects to Orbit for easy initialization

turnergarrow:
Added globular clusters from Vasiliev 2019 and dwarf galaxies from Helmi 2018 and Fritz 2018.


jobovy:
Thanks @turnergarrow! There are two tests that fail:

* one because the LMC was part of a previous test that returned different values from the ones that are currently in the file. You should replace the values being tested against with the ones from the file for that I think.
* And (at least) one of the automated tests that the Orbit that gets setup has the same values as those in the JSON file fails. I'm not sure which or why (could be a problem with the test I wrote). You can try to figure out what's going wrong by running the test locally (and can add print statements to it locally, just don't commit them)

Check that the tests pass by running them locally before committing more changes.

One thing that I probably didn't mention before: it's only necessary to provide the ``ro``, ``vo``, ``zo``, and ``solarmotion`` entries if the object for some reason requires a specific value for those (for example, the GD-1 entry that I put in comes from a fit to data that assumes values for these). If the object doesn't need them (and I don't think any of the ones you added need them, because all values just come from direct, straightforward measurements), then you should delete them from the JSON file.

jobovy:
Thanks @turnergarrow for the updates, this is starting to look very close to done! I made a few additions to this PR, adding tab-completion in IPython/Jupyter for the names of the known objects and adding support for collections of objects (such as 'MW globular clusters') and synonyms. Could you:

* Check this: I assumed that all objects in the file except for GD1, Sgr, LMC, and SMC are globular clusters, so I put them in the 'MW globular clusters' collection. Please confirm that this is correct
* Similarly, I assume Sgr, LMC, and SMC are the only dwarf satellites, so I put them in the 'MW satellites' collection. Please confirm this as well
* Add the other MW dwarf galaxies from Fritz et al. 2018 and then add them also to the 'mwsatellites' collection.
* Add all of the globular cluster synonyms from Vasiliev (the name in parentheses in their Table C1, I already put a few in the `_synonyms` section of the `named_objects.json` file.

jobovy:
Thanks for the updates @turnergarrow! I slightly tweaked the names of the satellites to be more like the commonly-used names (long known galaxies such as Draco, Sculptor, or Fornax don't typically get the I attached to their names etc.).

In checking this, I noticed that I think you used the Galactocentric distance from Table 2 in Fritz et al. (2018) as the distance, but the distance in the file should be the *heliocentric* distance. This you can get from the distance modulus that is given in their Table 1. Please fix that by converting their distance modulus to the distance in kpc.

Could you also tell me where you got the right ascension and declination from? As far as I can tell, those aren't listed in the Fritz et al. (2018) paper, so it would be useful to note down here where they come from.

Thanks!

turnergarrow:
I've changed the coordinates to use the heliocentric instead of Galactocentric distance. 
The right ascension and declination were taken from Riley et al. 2018 (arxiv:1810.10645), who cited the Fritz et al. paper for these values.

jobovy:
Thanks! I've added Eridanus II and Phoenix I just for completeness (Crater was already in the file, because Vasiliev has it as a globular cluster and therefore I've also not included in the satellites collection).

I think all looks good now! @webbjj, would you be able to take a look at this to see whether it looks okay to you?

webbjj:
Looks good to me. I cloned Turner's fork and tested Orbit.from_name to see
if it matched up with my (ugly) version in my nbodypy code for analyzing
Nbody clusters. Looking at a few different clusters yielded the same orbit
instances. I loved the autocomplete!

Jeremy

jobovy:
Great, thanks all! I will merge now.

PR #393, opened 2019-07-23
Added a C implementation of the MovingObjectPotential

turnergarrow:


galpy/potential/potential_c_ext/MovingObjectPotential.c
```diff
@@ -6,7 +6,11 @@
 // MovingObjectPotential
 // 5 fixed arguments: amp, t0, tf, nsteps, ndim
 // Followed by <nsteps> time values and <ndim>*<nsteps> coordinates
-double MovingObjectPotentialRforce(double R,double z, double phi,
+void constrain_range(double * d) {
+  // Constrains index to be within interpolation range
+  if (*d < 0) *d = 0.0;
+  if (*d > 1) *d = 1.0;
+}double MovingObjectPotentialRforce(double R,double z, double phi,
```
jobovy: Oops, need to fix that!

jobovy:
@turnergarrow: the code failed to compile on Windows and I tried to fix it. In particular, Windows didn't like the time array definition in ``initSplines`` and ``initPlanarSplines``, so I switched it to an explicit ``malloc/free`` in 4414a071a9752dc3f773e36975aab25f29a553b5; my understanding of the GSL splines is that once you've initialized the spline, you can get rid of the time array, but please confirm and check that the code that I changed is correct.

turnergarrow:
@jobovy you are correct, the time array can be removed once the spline has been initialized. The other changes you made look good as well.

galpy/potential/MovingObjectPotential.py
```diff
@@ -154,8 +159,11 @@ def _phiforce(self,R,z,phi=0.,t=0.):
         #Cylindrical distance
         Rdist = _cylR(R,phi,self._orb.R(t),self._orb.phi(t))
         # Difference vector
-        (xd,yd,zd) = _cyldiff(self._orb.R(t), self._orb.phi(t), self._orb.z(t),
+        orbz = self._orb.z(t) if self._orb.dim() == 3 else 0
+        (xd,yd,zd) = _cyldiff(self._orb.R(t), self._orb.phi(t), orbz,
             R, phi, z)
+        x = self._orb.R(t)*nu.cos(self._orb.phi(t))
```
jobovy: @turnergarrow: are ``x`` and ``y`` as defined in this line used anywhere?

galpy/orbit/orbit_c_ext/integratePlanarOrbit.c
```diff
@@ -611,3 +618,35 @@ void evalPlanarRectDeriv_dxdv(double t, double *q, double *a,
...
+
+  gsl_spline_init(x_spline, t, x_arr, nPts);
+  gsl_spline_init(y_spline, t, y_arr, nPts);
+
+  potentialArgs->xSpline = x_spline;
+  potentialArgs->accx = x_accel_ptr;
+  potentialArgs->ySpline = y_spline;
+  potentialArgs->accy = y_accel_ptr;
+
+  potentialArgs->nargs= (int) 5+(1+*(*pot_args+4))*nPts;
```
jobovy: @turnergarrow: from looking at the code, it seems like after you've set up the splines in C, the only arguments that you need are ``amp``, ``t0``, and ``tf`` (none of the other arguments seem to be used in the ``MovingObjectPotential.c`` code, which would make sense). But currently, all of the arguments get copied into the ``potentialArgs-->args`` array. So the memory footprint is twice as large as necessary.

This is similar to what happens in ``interpRZPotential``. There, we deal with this by putting the arguments that need to be stored in ``potentialArgs-->args`` as the last ones in the argument array. Then you could increment the ``pot_args`` pointer until you get to ``amp`` (I think this is 
```
*(*pot_args)+=  (int) 2+(1+*(*pot_args+4))*nPts
```
 in this case, but it's a little complicated because ``pot_args`` is a double pointer, so maybe the syntax is wrong [haven't tested]) and then do
```
potentialArgs->nargs= 3;
```
because then only the last three arguments will get copied by the general code for copying ``pot_args`` to ``potentialArgs->args``.

galpy/potential/potential_c_ext/MovingObjectPotential.c
```diff
@@ -0,0 +1,158 @@
...
+				      double t,
+				      struct potentialArg * potentialArgs){
+  double * args= potentialArgs->args;
+  //Get args
+  double amp= *args;
+  double t0= *(args+1);
+  double tf= *(args+2);
+  int n_steps= (int) *(args+3);
```
jobovy: @turnergarrow: I don't think ``n_steps``, ``n_dim``, or ``o`` is used in this function (or in any of the other functions). See my other comment about not copying these to ``args``.

galpy/potential/MovingObjectPotential.py
turnergarrow: No, I think those lines are left over from testing. They can be removed.

jobovy:
@turnergarrow: Thanks, all looks great now! Will merge this in.

PR #398, opened 2019-08-12
Add Potential function to create an AMUSE representation of any galpy potential

webbjj:
Ok, think this is what you requested me to do.  Sorry if I am still being a gidiot.


jobovy:
Great, thanks for the final check! Will merge then.

PR #401, opened 2019-10-10
added Dehnen Spherical Potential

nstarman:
A simple implementation of the Dehnen 1993 spherical potential.
amp is adjusted to amp*(3-alpha)
Since there is not a c-implementation of TwoPowerSphericalPotential, I added the keyword `allow_evolve` to DehnenSphericalPotential, which if false (default) will allow the class to return a Hernquist / Jaffe potential if the alpha parameter is 1 / 2. This allows a Hernquist/Jaffe potential to be implemented by calling DehnenSphericalPotential with alpha=1 / 2 and a c-implemented Hernquist/JaffePotential will be returned. The `allow_evolve=True` option also means that a WrapperPotential could modify alpha.


nstarman:
That's strange. I ran the tests on my computer and they all passed. The error seems to be 
`AssertionError: Analytically computed pericenter radius does not agree with numerical estimate for potential DehnenSphericalPotential and integrator dopr54_c`

nstarman:
Don't merge yet, I want to check the amplitude normalization.

nstarman:
Ok, I'm really not sure what codecov is complaining about.
But the normalizations work well, especially with this new commit.
I am worried about the stability of the code for alpha=1+eps or alpha=2+eps since the general formulae for `._evaluate` and other methods have 1/(alpha-2)(alpha-3) factors.

PR #403, opened 2019-10-21
Added Dehnen Spherical Potential

nstarman:
Implementation of the Dehnen 1993 spherical potential.


galpy/potential/Potential.py
```diff
@@ -70,7 +70,7 @@ def __init__(self,amp=1.,ro=None,vo=None,amp_units=None):
         OUTPUT:
         HISTORY:
         """
-        Force.__init__(self,amp=amp,ro=ro,vo=vo,amp_units=amp_units)
+        super(Potential, self).__init__(amp=amp,ro=ro,vo=vo,amp_units=amp_units)
```
jobovy: Please don't change this line (nothing to do with this PR).

galpy/potential/interppotential_c_ext/interppotential_calc_potential.c
```diff
@@ -17,6 +17,7 @@
 #include <actionAngle.h>
 #include <integrateFullOrbit.h>
 #include <interp_2d.h>
+#include <sf_math.h>
```
jobovy: Doesn't seem necessary here (there is no other change in this file, so can't be that anything from this is now used).

galpy/potential/potential_c_ext/DehnenPotential.c
```diff
@@ -0,0 +1,113 @@
...
+  double a= *args++;
+  double alpha= *args;
+  // See if Hernquist or Jaffe
+  if (alpha == 1.) {
+    return HernquistPotentialEval(R, Z, phi, t, potentialArgs);
+  }
+  else if (alpha == 2.) {
+    return JaffePotentialEval(R, Z, phi, t, potentialArgs);
+  }
+  // else, calculate Rforce
```
jobovy: As long as this code is run through galpy's Python interface (which I think we should assume), I don't think the user could ever end up here, because Hernquist and Jaffe should just be caught at the potential's initialization. So I think we should remove these special cases in this file.

galpy/potential/potential_c_ext/DehnenPotential.c
```diff
@@ -0,0 +1,113 @@
...
+  double amp= *args++;
+  double a= *args++;
+  double alpha= *args;
+  // See if Hernquist or Jaffe
+  if (alpha == 1.) {
+    return HernquistPotentialRforce(R, Z, phi, t, potentialArgs);
+  }
+  else if (alpha == 2.) {
+    return JaffePotentialRforce(R, Z, phi, t, potentialArgs);
+  }
```
jobovy: Remove special cases

galpy/potential/potential_c_ext/DehnenPotential.c
```diff
@@ -0,0 +1,113 @@
...
+  double amp= *args++;
+  double a= *args++;
+  double alpha= *args;
+  // See if Hernquist or Jaffe
+  if (alpha == 1.) {
+    return HernquistPotentialPlanarRforce(R, phi, t, potentialArgs);
+  }
+  else if (alpha == 2.) {
+    return JaffePotentialPlanarRforce(R, phi, t, potentialArgs);
+  }
```
jobovy: Remove special cases

galpy/potential/potential_c_ext/DehnenPotential.c
```diff
@@ -0,0 +1,113 @@
...
+  double amp= *args++;
+  double a= *args++;
+  double alpha= *args;
+  // See if Hernquist or Jaffe
+  if (alpha == 1.) {
+    return HernquistPotentialzforce(R, Z, phi, t, potentialArgs);
+  }
+  else if (alpha == 2.) {
+   return JaffePotentialzforce(R, Z, phi, t, potentialArgs);
+  }
```
jobovy: Remove special cases

galpy/potential/potential_c_ext/DehnenPotential.c
```diff
@@ -0,0 +1,113 @@
...
+  double amp= *args++;
+  double a= *args++;
+  double alpha= *args;
+  // See if Hernquist or Jaffe
+  if (alpha == 1.) {
+    return HernquistPotentialPlanarR2deriv(R, phi, t, potentialArgs);
+  }
+  else if (alpha == 2.) {
+    return JaffePotentialPlanarR2deriv(R, phi, t, potentialArgs);
+  }
```
jobovy: Remove special cases

galpy/potential/potential_c_ext/galpy_potentials.h
```diff
@@ -8,6 +8,7 @@
 extern "C" {
 #endif
 #include <interp_2d.h>
+#include <sf_math.h>
```
jobovy: Don't think this is necessary here?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -8,11 +8,14 @@
 ###############################################################################
 import math as m
 import numpy
-from scipy import special, optimize, integrate
+from numpy import sqrt, pi
```
jobovy: ``galpy`` coding style is to always explicitly include ``numpy.`` in numpy calls (there is older code that uses ``nu.`` left, but now should be ``numpy.``)

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -8,11 +8,14 @@
...
+from scipy import optimize, integrate
+from scipy.special import gamma as Gamma, hyp2f1
```
jobovy: To remember where these functions come from, just import ``special`` and use these functions as ``special.``

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -21,8 +24,8 @@ class TwoPowerSphericalPotential(Potential):
 
         \\rho(r) = \\frac{\\mathrm{amp}}{4\\,\\pi\\,a^3}\\,\\frac{1}{(r/a)^\\alpha\\,(1+r/a)^{\\beta-\\alpha}}
     """
-    def __init__(self,amp=1.,a=5.,alpha=1.5,beta=3.5,normalize=False,
-                 ro=None,vo=None):
+    def __init__(self,amp=1.,a=5.,alpha=1.5,beta=3.5,
+                 normalize=False,ro=None,vo=None):
         """
```
jobovy: Please avoid these types of changes (whitespace, linebreaks) in PRs if they have nothing to do with the PR.

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -53,30 +56,63 @@ def __init__(self,amp=1.,a=5.,alpha=1.5,beta=3.5,normalize=False,
...
-        if alpha == round(alpha) and beta == round(beta):
-            Potential.__init__(self,amp=amp,ro=ro,vo=vo,amp_units='mass')
-            integerSelf= TwoPowerIntegerSphericalPotential(amp=1.,a=a,
-                                                           alpha=int(alpha),
-                                                           beta=int(beta),
-                                                           normalize=False)
-            self.integerSelf= integerSelf
+        # instantiate
+        super(TwoPowerSphericalPotential, self).__init__(amp=amp,ro=ro,vo=vo,amp_units='mass')
```
jobovy: I prefer ``Potential.__init__(self,...``

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -53,30 +56,63 @@ def __init__(self,amp=1.,a=5.,alpha=1.5,beta=3.5,normalize=False,
...
+    @beta.setter
+    def beta(self, beta):
+        self._beta = beta  # setting alpha
+        # update .integerSelf
+        if self._alpha == round(self._alpha) and beta == round(beta):
+            self.integerSelf= TwoPowerIntegerSphericalPotential(
+                amp=1.,a=self.a,alpha=round(self._alpha),beta=round(beta),
+                normalize=False)
+        else:
+            self.integerSelf= None
```
jobovy: I'm not sure I see the point of allowing alpha and beta to be adjusted like this, this isn't something that is typically allowed for galpy potentials, so introduces inconsistencies in how the parameters of potentials can be changed. I'd prefer to remove this part of the PR and perhaps re-think in a more general way in the future

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -93,25 +129,21 @@ def _evaluate(self,R,z,phi=0.,t=0.,_forceFloatEval=False):
...
+            return (1./a)\
+                *(r-a*(r/a)**(3.-alpha)/(3.-alpha)\
+                      *hyp2f1(3.-alpha, 2.-alpha, 4.-alpha, -r/a))/(alpha-2.)/r
+        # else:
+        a, alpha, beta = self.a, self._alpha, self._beta
+        r= sqrt(R**2.+z**2.)
+        return Gamma(beta-3.)\
+            *((r/a)**(3.-beta)/Gamma(beta-1.)\
+                  *hyp2f1(beta-3., beta-alpha, beta-1., -a/r)
+              -Gamma(3.-alpha)/Gamma(beta-alpha))/r
```
jobovy: From a glance through this, nothing seems to have changed; why did you re-write this?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -129,15 +161,13 @@ def _Rforce(self,R,z,phi=0.,t=0.,_forceFloatEval=False):
...
-            return -R/r**self.alpha*self.a**(self.alpha-3.)/(3.-self.alpha)\
-                *special.hyp2f1(3.-self.alpha,
-                                self.beta-self.alpha,
-                                4.-self.alpha,
-                                -r/self.a)
+        # else:
+        a, alpha, beta = self.a, self._alpha, self._beta
+        r= sqrt(R**2.+z**2.)
+        return -R/r**alpha*a**(alpha-3.)/(3.-alpha)\
+            *hyp2f1(3.-alpha, beta-alpha, 4.-alpha, -r/a)
```
jobovy: Why did you re-write this, nothing seems changed? It's good practice to show clearly that this is an ``else`` statement (and it's bad practice of mine to return in the middle of the function...).

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -155,15 +185,119 @@ def _zforce(self,R,z,phi=0.,t=0.,_forceFloatEval=False):
...
-            return -z/r**self.alpha*self.a**(self.alpha-3.)/(3.-self.alpha)\
-                *special.hyp2f1(3.-self.alpha,
-                                self.beta-self.alpha,
-                                4.-self.alpha,
-                                -r/self.a)
+        # else:
+        a, alpha, beta = self.a, self._alpha, self._beta
+        r= sqrt(R**2.+z**2.)
+        return -z/r**alpha*a**(alpha-3.)/(3.-alpha)\
+            *hyp2f1(3.-alpha, beta-alpha, 4.-alpha, -r/a)
```
jobovy: I think this is again an unnecessary change?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -155,15 +185,119 @@ def _zforce(self,R,z,phi=0.,t=0.,_forceFloatEval=False):
...
+    #              a**(-3 + beta) * (R**2 + z**2)**(1 - beta / 2.) *
+    #              ((z**2 * (a * (alpha - beta) - r * (beta - 2)) +
+    #                R**2 * (r * (beta - 2) *
+    #                        (beta - 1) + 2 * a * (beta - alpha))
+    #                ) * hyp2f1(-2 + beta, beta - alpha + 1, beta,
+    #                           -(a / r)) / Gamma(beta) +
+    #               (a * (R**2 * (2 + alpha * (-3 + beta)) - z**2 * (beta - 2)) +
+    #                r * (-z**2 + R**2 * (beta - 1)) * (beta - 2)) *
+    #               (-2 + beta) * hyp2f1(beta - 1, beta - alpha + 1, beta,
+    #                                    -(a / r)) / Gamma(beta))))
```
jobovy: Not sure that we should keep this if it's going to be commented out...

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -155,15 +185,119 @@ def _zforce(self,R,z,phi=0.,t=0.,_forceFloatEval=False):
...
+    #             ((-3 * Gamma(3 - alpha)) / Gamma(beta - alpha) +
+    #              a**(beta - 3) * r**(2 - beta) *
+    #              ((r * (beta - 2) * beta + 3 * a * (-alpha + beta)) *
+    #               hyp2f1(beta - 2, beta - alpha + 1, beta,
+    #                      -(a / r)) / Gamma(beta) +
+    #               (-2 + beta) * (r * (-2 + beta) * beta +
+    #                              a * (alpha * (-3 + beta) + beta)) *
+    #               hyp2f1(beta - 1, beta - alpha + 1, beta,
+    #                      -(a / r)) / Gamma(beta))))
+
```
jobovy: Not sure that we should keep this if it's going to be commented out

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
         HISTORY:
            2010-08-08 - Written - Bovy (NYU)
         """
-        r= numpy.sqrt(R**2.+z**2.)
-        return (self.a/r)**self.alpha/(1.+r/self.a)**(self.beta-self.alpha)/4./m.pi/self.a**3.
+        a, alpha, beta = self.a, self._alpha, self._beta
+        r= sqrt(R**2.+z**2.)
+        return (a/r)**alpha/(1.+r/a)**(beta-alpha)/4./m.pi/a**3.
```
jobovy: Unnecessary change?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
...
+        HISTORY:
+           2014-04-01 - Written - Erkal (IoA)
+        """
+        a, alpha, beta = self.a, self._alpha, self._beta
+        r= R if z is None else sqrt(R**2.+z**2.)
+        # if z is None:
+        #     r= R
+        # else:
+        #     r= sqrt(R**2.+z**2.)
+        return (r/a)**(3.-alpha)/(3.-alpha)*hyp2f1(3.-alpha,-alpha+beta,4.-alpha,-r/a)
```
jobovy: Unnecessary change?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
...
+
+
+class DehnenSphericalPotential(TwoPowerSphericalPotential):
+    """Class that implements the Dehnen Spherical Potential from Dehnen 1993
```
jobovy: Add hyperlink to paper

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
...
+
+        INPUT:
+
+           amp - amplitude to be applied to the potential (default: 1); can be a Quantity with units of mass or Gxmass will be rescaled to amp*(3-alpha) (except for Jaffe, which is already scaled)
+
+           a - scale radius (can be Quantity)
+
+           alpha - inner power, restricted to [0, 3)
+
+           dehnen_amp - if True, use dehnen normalization, amp -> amp * (3 - alpha)
```
jobovy: Not sure we should support this, it's pretty trivial, but just adds a use-case to test...

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
...
+            (alpha == round(alpha))):
+            self.integerSelf= TwoPowerIntegerSphericalPotential(
+                amp=1.,a=a,alpha=round(alpha),beta=4,
+                normalize=False,ro=None,vo=None)
+        else:
+            self.integerSelf= None
+        # set properties
+        self.hasC= True
+        self.hasC_dxdv= True
+        self._nemo_accname= 'Dehnen'
```
jobovy: Did you add a test to ``test_nemo``?

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -181,8 +315,143 @@ def _dens(self,R,z,phi=0.,t=0.):
...
+                normalize=False,)
+        else:
+            self.integerSelf= None
+
+    @property
+    def beta(self):
+        return self._beta
+    @beta.setter
+    def beta(self, value):
+        raise Exception('cannot modify beta')
```
jobovy: Isn't this handled by the super-class? Anyway, see comment about that above.

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -198,88 +467,140 @@ def _z2deriv(self,R,z,phi=0.,t=0.):
...
-           2012-07-26 - Written - Bovy (IAS@MPIA)
+           2019-10-20 - Written - Starkman (UofT)
         """
-        return self._R2deriv(numpy.fabs(z),R) #Spherical potential
+        a, alpha = self.a, self._alpha
+        r = sqrt(R**2. + z**2.)
+        return (((1 + a / r)**alpha *
+                 (R**4 - 2 * z**4 + R**2 * (-z**2 + a * r) -
+                  a * z**2 * r * (-1 + alpha))) /
+                (r**3 * (a + r)**4 * (3 - alpha)))
```
jobovy: Use ``return self._R2deriv(numpy.fabs(z),R) #Spherical potential``

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -198,88 +467,140 @@ def _z2deriv(self,R,z,phi=0.,t=0.):
...
+        return p / q
+
 
 class TwoPowerIntegerSphericalPotential(TwoPowerSphericalPotential):
     """Class that implements the two-power-density spherical potentials in 
     the case of integer powers"""
-    def __init__(self,amp=1.,a=1.,alpha=1,beta=3,normalize=False,
-                 ro=None,vo=None):
+    def __init__(self,amp=1.,a=1.,alpha=1,beta=3,
+                 normalize=False,ro=None,vo=None):
```
jobovy: Unnecessary change

nstarman: reverted.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done

galpy/potential/Potential.py
nstarman: reverted.

galpy/potential/interppotential_c_ext/interppotential_calc_potential.c
nstarman: deleted.

galpy/potential/potential_c_ext/DehnenPotential.c
nstarman: first just commenting them out to test.

galpy/potential/potential_c_ext/DehnenPotential.c
nstarman: first just commenting them out to test.

galpy/potential/potential_c_ext/DehnenPotential.c
nstarman: first just commenting them out to test.

galpy/potential/potential_c_ext/DehnenPotential.c
nstarman: first just commenting them out to test.

galpy/potential/potential_c_ext/DehnenPotential.c
nstarman: first just commenting them out to test.

galpy/potential/potential_c_ext/galpy_potentials.h
nstarman: I started by adding <sf_math.h> wherever interp_2d appeared to ensure that the addition worked.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: reverted.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: for the SciPy special function spaces.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: I've solved the special cases in Mathematica, but they involve partial derivatives of the hypergeometric function, which is not in scipy special functions.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: I've solved the special cases in Mathematica, but they involve partial derivatives of the hypergeometric function, which is not in scipy special functions.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: yes.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: the `r= R if z is None else sqrt(R**2.+z**2.)` should be infinitesimally faster

galpy/potential/TwoPowerSphericalPotential.py
nstarman: I put in in there so that the default `DehnenSphericalPotential` would match `HernquistPotential` when initialized with the same arguments. switching `dehnen_amp` to True makes it match the Dehnen 1993 paper. This is also useful if `alpha` is ever permitted to vary.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: added now.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: I've commented this out.
This version doesn't permit beta to vary.

nstarman:
> I don't like the idea of adding a bunch of C code just to get the hypergeometric functions (I think?). The latest version of the GSL seems to have hypergeometric functions , so that would be preferable. Of course, if it's only in a recent version, then we'll have to sort out how to deal with older versions of the GSL..

I tried using the GSL hypergeometric functions, except that they do not cover x < -1 in hyp2f1(a, b, c, x) and for x = - r / a this is a big problem. All the C code is for scipy's hyp2f1, which covers the required domain. The scipy functions have been modified to use GSL functions rather than the entire Cephes numerical library.

nstarman:
> I'm not a fan of allowing alpha and beta to be changed and the amount of code this requires
I've commented all of this out. Will delete before final commit.

galpy/potential/potential_c_ext/galpy_potentials.h
nstarman: re-adding it to test why the build errored.

galpy/potential/interppotential_c_ext/interppotential_calc_potential.c
nstarman: readied for test.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done.

galpy/potential/TwoPowerSphericalPotential.py
jobovy: So maybe we should leave them out yet, if we ever wanted to add them it would be easy enough to re-do the Mathematica calc. (also, we can get 2nd derivatives with the ``NumericalDerivativesMixin``)

galpy/potential/TwoPowerSphericalPotential.py
jobovy: This is gone?

galpy/potential/TwoPowerSphericalPotential.py
jobovy: Ah, cool, then we should keep it; but remove the commented out lines

galpy/potential/TwoPowerSphericalPotential.py
jobovy: I would prefer just keeping a single amp definition and just putting a note in the docs that this definition differs from the Dehnen paper. It will just be less confusing and require less testing of different cases...

galpy/potential/interppotential_c_ext/interppotential_calc_potential.c
jobovy: So this is necessary? Seems weird if it's not used...

jobovy:
Thanks for the changes so far. I looked into this again and I'm a little confused, because it seems to me like you don't need the hypergeometric functions to implement the Dehnen models. Dehnen's paper gives expressions for the potential and enclosed mass (--> force ~ M /r^2) as simple functions. I think you've implemented the general two-power case here, which I think we don't want to do, because I don't want to include so much additional C code (that would have to be supported into the future) unless it's absolutely necessary.

So I would suggest:

* Implement the specific Dehnen potential, forces, 2nd derivs in python in the Dehnen class
* Implement that in C
* Remove all hypergeometric code

Another thing you'll have to add is a documentation entry for the Dehnen models, you can use ``HernquistPotential` as a guide.

I've made some changes to the travis setup in ``master`` that should make the tests run faster and avoid the FTP time-out error that we often get, so merge in the latest ``master`` before pushing again.

galpy/potential/interppotential_c_ext/interppotential_calc_potential.c
nstarman: not sure. Have to test without it. I just re-added many of the things I deleted when it stopped working.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: deleted. Will (not) show up in next push.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: deleted. Will (not) show up in next push.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: Ok. I've made dehnen_amp=False the norm. This will show up in the next push.

jobovy: I think you should remove the ``dehnen_amp`` option altogether, not just change the default.

nstarman:
> Thanks for the changes so far. I looked into this again and I'm a little confused, because it seems to me like you don't need the hypergeometric functions to implement the Dehnen models. Dehnen's paper gives expressions for the potential and enclosed mass (--> force ~ M /r^2) as simple functions. I think you've implemented the general two-power case here, which I think we don't want to do, because I don't want to include so much additional C code (that would have to be supported into the future) unless it's absolutely necessary.
> 
> So I would suggest:
> 
> * Implement the specific Dehnen potential, forces, 2nd derivs in python in the Dehnen class
> * Implement that in C
> * Remove all hypergeometric code
> 
> Another thing you'll have to add is a documentation entry for the Dehnen models, you can use ``HernquistPotential` as a guide.
> 
> I've made some changes to the travis setup in `master` that should make the tests run faster and avoid the FTP time-out error that we often get, so merge in the latest `master` before pushing again.

- [x] Implement the specific Dehnen potential, forces, 2nd derivs in python in the Dehnen class
- [x] Implement that in C
- [x] Remove all hypergeometric code

I added .rst files for the Dehnen code.
I also added a DehnenCoredSphericalPotential with alpha=0,  beta=4 to complete the complement of Hernquist and Jaffe. This potential is also nicely linked through TwoPowerIntegerSphericalPotential and solves the problem of integer parameters in Dehnen not having a counterpart in TwoPowerIntegerSphericalPotential

galpy/potential/TwoPowerSphericalPotential.py
nstarman: removed.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done. and changed them all.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: done.

galpy/potential/TwoPowerSphericalPotential.py
```diff
@@ -217,10 +219,244 @@ def _mass(self,R,z=0.,t=0.):
...
+            (alpha == round(alpha))):
+            self.integerSelf= TwoPowerIntegerSphericalPotential(
+                amp=1.,a=a,alpha=round(alpha),beta=4,
+                normalize=False,ro=None,vo=None)
+        else:
+            self.integerSelf= None
+        # set properties
+        self.hasC= True
+        self.hasC_dxdv= True
+        # self._nemo_accname= 'Dehnen'
```
nstarman: not sure if it Nemo compatible. How should I check?

jobovy: You should add a test to ``tests/test_nemo.py`` similar to the tests that are already in that file. These don't run on travis (like the AMUSE ones, because NEMO is too difficult to install on travis), but you should be able to run them on the server if you get NEMO to work there using [these instructions](http://astro.utoronto.ca/~bovy/group/compute_servers.html#nemo)

jobovy:
Thanks for the changes, I think it's now getting very close! I haven't had time to conduct a full review of the new changes, but a few things:
 - [ ] Because DehnenCored and Hernquist inherit from DehnenSpherical, in the ``integrateFullOrbit.py`` and ``integratePlanarOrbit.py`` code, DehnenCored and Hernquist don't get passed to C as themselves, but rather as a DehnenSpherical instance. Thus, the C code for DehnenCored and Hernquist doesn't get touched. I think you can simply fix this by re-ordering the ``if`` statements there, so that you first test Hernquist and DehnenCored and only then the more general DehnenSpherical (which should be the case at least for Hernquist, because new potentials should be added to the bottom of the list). Maybe better would be to test for DehnenSpherical and NOT DehnenCored or Hernquist
 - [ ] Is the DehnenCore model a special model that's used a lot? Just wondering whether there is much point to introducing another new potential...
 - [ ] Looks like there are still some new lines in ``TwoPowerSpherical`` that aren't touched by the tests (don't use [skip travis] unless there's a reason, because it's now hard to assess the report, because the travis tests didn't run for the latest changes I think).

nstarman:
> Thanks for the changes, I think it's now getting very close! I haven't had time to conduct a full review of the new changes, but a few things:
> 
> * [ ]  Because DehnenCored and Hernquist inherit from DehnenSpherical, in the `integrateFullOrbit.py` and `integratePlanarOrbit.py` code, DehnenCored and Hernquist don't get passed to C as themselves, but rather as a DehnenSpherical instance. Thus, the C code for DehnenCored and Hernquist doesn't get touched. I think you can simply fix this by re-ordering the `if` statements there, so that you first test Hernquist and DehnenCored and only then the more general DehnenSpherical (which should be the case at least for Hernquist, because new potentials should be added to the bottom of the list). Maybe better would be to test for DehnenSpherical and NOT DehnenCored or Hernquist
> * [ ]  Is the DehnenCore model a special model that's used a lot? Just wondering whether there is much point to introducing another new potential...
> * [ ]  Looks like there are still some new lines in `TwoPowerSpherical` that aren't touched by the tests (don't use [skip travis] unless there's a reason, because it's now hard to assess the report, because the travis tests didn't run for the latest changes I think).

- [x] re-ordered the `if` statements. I was hesitant to add the new potentials to the bottom of the list since they are closely associated with the Hernquist and Jaffe. Ideally the cases would be recorded to slot the Dehnen(Core) with the Hernquist and Jaffe, but I didn't because it would be very easy to make a mistake. I can put them at the end. I did reorder the  Dehnen(Core) case numbers so that the more general Dehnen naturally comes later.

- [x ]  introducing DehnenCore. I did it for 2 reasons. 1) there isn't a cored NFW-like potential in Galpy and the cored Dehnen can be used as a substitute and was a lot easier to program. 2) the way the code is structured, with `TwoPowerIntegerSphericalPotential` requires either a number of `if` statements to catch the DehnenCore case, or to just add the potential. I found it cleaner and more transparent to have dedicated potentials for the integer-values of alpha and beta. Also, the code should be more resilient to $R\rightarrow0$ since I can analytically cancel these terms from the denominator

- [x] I'll run it with Travis. I skipped had travis since the lines added were just copy-paste replacements for a line that had been previously tested.

galpy/potential/TwoPowerSphericalPotential.py
nstarman: I followed the instructions but all the Nemo tests (for all the potentials, not only dehnen) are failing. I imagine that means my NEMO is not working correctly.

jobovy:
To test the C implementation of the potential evaluation (currently not hit by the tests), you should add a test to ``tests/test_actionAngle.py#test_actionAngleStaeckel_wSpherical_conserved_actions_c``, similar to the ones already there (just add the potentials DehnenCore and a Dehnen case that isn't a special case).

nstarman:
I cannot find the test for DehnenSphericalPotentialEval and DehnenCoreSphericalPotentialEval. That's what's preventing the 99.8% coverage.

jobovy:
See my comment above, just add these two potentials in that ``actionAngle`` test.

jobovy:
I think the issue with the remaining untested C line is that the special case alpha=2 is the JaffePotential, so never gets setup in galpy. That means that we should probably remove the special case.

Was also wondering whether we should get rid of ``TwoPowerIntegerSphericalPotential``, because as written I don't think it's actually doing anything? I think I originally introduced it because in principle there are quite some simplifications for the case of integer powers, but I never implemented them, and given that almost ten years have gone by without needing that, maybe that means it's not really useful. Then ``TwoPowerSphericalPotential`` could just have a ``_specialSelf`` member for Hernquist, Dehnen, etc. and might allow us to get rid of all ``forceFloatEval`` etc?

nstarman:
- [x] removed the special case in DehnenSpherical.c
- [x] Got rid of TwoPowerIntegerSphericalPotential
- [x] renamed `integerSelf` to `_specialSelf`
- [x] got rid of `forceFloatEval`

jobovy:
Looks like you got 100% diff coverage, impressive 🎉 🎉! Will review fully soon then.

PR #404, opened 2019-10-30
add time dependences to Potential and related functions

nstarman:
there are a lot of functions, like vcirc, that call time-dependent functions but do not themselves have a time parameter. I just went and found as many time-dependent functions as I could. in `galpy.potential` (and all functions that called those functions) and gave them time parameters.

### DIFF
--- a/galpy/potential/Potential.py
+++ b/galpy/potential/Potential.py
@@ -411,17 +411,17 @@ class Potential(Force):
                               ", specify z)",galpyWarning)
                 return 4.*nu.pi\
                     *integrate.quad(lambda x: x**2.\
-                                        *self.dens(x,0.,
+                                        *self.dens(x,0.,t=t,
                                                   use_physical=False),
                                     0.,R)[0]
             else:
                 return 4.*nu.pi\
                     *integrate.dblquad(lambda y,x: x\
-                                           *self.dens(x,y,use_physical=False),
+                                           *self.dens(x,y,t=t,use_physical=False),
                                        0.,R,lambda x: 0., lambda x: z)[0]
 
     @physical_conversion('mass',pop=False)
-    def mvir(self,H=70.,Om=0.3,overdens=200.,wrtcrit=False,
+    def mvir(self,H=70.,Om=0.3,t=0.,overdens=200.,wrtcrit=False,
              forceint=False,ro=None,vo=None,
              use_physical=False): # use_physical necessary bc of pop=False, does nothing inside
         """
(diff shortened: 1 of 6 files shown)


galpy/potential/Potential.py
```diff
@@ -578,7 +578,7 @@ def Rzderiv(self,R,Z,phi=0.,t=0.):
         except AttributeError: #pragma: no cover
             raise PotentialError("'_Rzderiv' function not implemented for this potential")      
 
-    def normalize(self,norm,t=0.):
+    def normalize(self,norm):
```
jobovy: Why did you remove the time dependence here? I don't think it's used ever, that's true...

galpy/potential/plotEscapecurve.py
```diff
@@ -145,14 +148,25 @@ def calcEscapecurve(Pot,Rs):
     except TypeError:
         grid=1
         Rs= nu.array([Rs])
+    try:
```
jobovy: I don't think it makes sense to allow an array for ``t`` here, because the point of the function is to compute the escape curve at a given time. So just change this to ``t`` being a number.

galpy/potential/plotEscapecurve.py
```diff
@@ -120,7 +120,7 @@ def plotEscapecurve(Pot,*args,**kwargs):
     return plot.bovy_plot(Rs,esccurve,*args,
                           **kwargs)
 
-def calcEscapecurve(Pot,Rs):
+def calcEscapecurve(Pot,Rs,ts=0.):
```
jobovy: Docs say ``t``, for consistency with other methods, use ``t``.

galpy/potential/plotRotcurve.py
```diff
@@ -158,14 +160,25 @@ def calcRotcurve(Pot,Rs,phi=None):
     except TypeError:
         grid=1
         Rs= nu.array([Rs])
+    try:
```
jobovy: Like for the escape curve, only allow a scalar time here.

galpy/potential/plotRotcurve.py
```diff
@@ -123,7 +123,7 @@ def plotRotcurve(Pot,*args,**kwargs):
     return plot.bovy_plot(Rs,rotcurve,*args,
                           **kwargs)
 
-def calcRotcurve(Pot,Rs,phi=None):
+def calcRotcurve(Pot,Rs,phi=None,ts=0.):
```
jobovy: Should be ``t`` for consistency.

galpy/potential/Potential.py
nstarman: yeah, 1) because I could never find a time it's used and 2) because I couldn't think of a time when a potential would be normalized except at instantiation.

nstarman: also, `normalize` wasn't passing the parameter to `Rforce`.

PR #405, opened 2019-11-06
fixed amuse compatibility with Potentials instantiated with astropy quantities

nstarman:
A Potential instantiated with astropy quantitities has the `physical` flag set to true. The `galpy_profile` wrapper for amuse compatibility does all calculations without physical units. Normally this is not a problem as the galpy output is piped directly from galpy to. AMUSE,  but for `galpy_profile.get_gravity_at_point` the galpy output is first converted to cartesian coordinates and the units are adjusted. To prevent double conversion in the instances that galpy uses physical units, the `evaluateRforces`, `evaluatephiforces`, and `evaluatezforces` methods need to pass `use_physical=False`.

### DIFF
--- a/galpy/potential/amuse.py
+++ b/galpy/potential/amuse.py
@@ -102,7 +102,6 @@ class galpy_profile(LiteratureReferencesMixIn):
         R= numpy.sqrt(x.value_in(units.kpc)**2.+y.value_in(units.kpc)**2.)
         zed= z.value_in(units.kpc)
         phi= numpy.arctan2(y.value_in(units.kpc),x.value_in(units.kpc))
-
         res= potential.evaluatePotentials(self.pot,R/self.ro,zed/self.ro,
                                           phi=phi,t=self.tgalpy,
                                           ro=self.ro,vo=self.vo,
@@ -174,7 +173,7 @@ class galpy_profile(LiteratureReferencesMixIn):
                                           ro=self.ro,vo=self.vo,
                                           use_physical=False) *
               bovy_conversion.dens_in_msolpc3(self.vo,self.ro))
-        return res | units.MSun / units.parsec**3.
+        return res | units.MSun / units.parsec**3
 
     def circular_velocity(self,r):
         """
@@ -212,7 +211,8 @@ class galpy_profile(LiteratureReferencesMixIn):
         vc= potential.vcirc(self.pot,r.value_in(units.kpc)/self.ro,phi=0,
                             t=self.tgalpy,ro=self.ro,vo=self.vo,
                             use_physical=False) * self.vo
-        return (vc**2.)*r.value_in(units.parsec)/bovy_conversion._G | units.MSun
+        return (vc**2.)*r.value_in(units.parsec)/bovy_conversion._G \
+            | units.MSun
 
     def stop(self):
         """
--- a/galpy/potential/amuse.py
+++ b/galpy/potential/amuse.py
@@ -1,10 +1,13 @@
 # galpy.potential.amuse: AMUSE representation of galpy potentials
 import numpy
-from amuse.units import units
+from amuse.units import units as u
 from amuse.units.quantities import ScalarQuantity
 from amuse.support.literature import LiteratureReferencesMixIn
 from .. import potential
 from ..util import bovy_conversion
+from .Force import _APY_LOADED
+if _APY_LOADED:
+    from astropy import units as apy_u
 class galpy_profile(LiteratureReferencesMixIn):
     """
     User-defined potential from galpy
@@ -54,10 +57,10 @@ class galpy_profile(LiteratureReferencesMixIn):
         else:
             self.model_time= \
                 t*bovy_conversion.time_in_Gyr(ro=self.ro,vo=self.vo) \
-                | units.Gyr
+                | u.Gyr
         #Initialize galpy time
         if isinstance(tgalpy,ScalarQuantity):
-            self.tgalpy= tgalpy.value_in(units.Gyr)\
+            self.tgalpy= tgalpy.value_in(u.Gyr)\
                 /bovy_conversion.time_in_Gyr(ro=self.ro,vo=self.vo)
         else:
             self.tgalpy= tgalpy
@@ -78,10 +81,10 @@ class galpy_profile(LiteratureReferencesMixIn):
         dt= time-self.model_time
         self.model_time= time  
         if self.reverse:
-            self.tgalpy-= dt.value_in(units.Gyr)\
+            self.tgalpy-= dt.value_in(u.Gyr)\
                 /bovy_conversion.time_in_Gyr(ro=self.ro,vo=self.vo)
         else:
-            self.tgalpy+= dt.value_in(units.Gyr)\
+            self.tgalpy+= dt.value_in(u.Gyr)\
                 /bovy_conversion.time_in_Gyr(ro=self.ro,vo=self.vo)
 
     def get_potential_at_point(self,eps,x,y,z):
@@ -98,13 +101,14 @@ class galpy_profile(LiteratureReferencesMixIn):
         HISTORY:
            2019-08-12 - Written - Webb (UofT)
         """
-        R= numpy.sqrt(x.value_in(units.kpc)**2.+y.value_in(units.kpc)**2.)
-        zed= z.value_in(units.kpc)
-        phi= numpy.arctan2(y.value_in(units.kpc),x.value_in(units.kpc))
+        R= numpy.sqrt(x.value_in(u.kpc)**2.+y.value_in(u.kpc)**2.)
+        zed= z.value_in(u.kpc)
+        phi= numpy.arctan2(y.value_in(u.kpc),x.value_in(u.kpc))
         return potential.evaluatePotentials(self.pot,R/self.ro,zed/self.ro,
                                             phi=phi,t=self.tgalpy,
-                                            ro=self.ro,vo=self.vo) \
-                                            | units.km**2*units.s**-2
+                                            ro=self.ro,vo=self.vo,
+                                            use_physical=False
+                                            ) | u.km**2*u.s**-2
 
     def get_gravity_at_point(self,eps,x,y,z):
         """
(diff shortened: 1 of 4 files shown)


jobovy:
Good point! Could you replace all occurrences of ``evaluateX`` with the correct call? Also, please don't change the ``LiteratureReferencesMixin.__init__`` call; that's unrelated to the topic of this PR and it's not so clear that calling ``super`` is always better in a case like this.

jobovy:
I think there are still some instances of ``evaluateX`` left, also the vcirc calls should have ``use_physical=False`` (basically all ``potential.X`` calls should be changed).

Don't force push! (makes it hard to keep track of changes).

nstarman:
I was leaving instances of `potential.X(self.pot, ...)` instead of switching to `self.pot.X` since these methods can handle when `self.pot` is a list of potentials, rather than a single potential.

jobovy:
Yes, that's good, I just meant that there are uses of ``potential.X`` functions where you haven't added ``use_physical=False`` (like in ``get_potential_at_point`` and the vcirc and mass functions I think).

nstarman:
I see. I've been testing this with `astropy-units=False` in my `.galpyrc`. In this case the only function was `get_gravity_at_point` since there is a  conversion from spherical to cartesian  coordinates at the end.  If  `use_physical` is not `False` and the galpy potential was instantiated with units (fine even if `astropy-units=False`) then the internal `physical` flag is turned on.  Then, `get_gravity_at_point` converts quantities a second time (1st by the potential, second by `galpy_profile`). By passing the use_physical=False, it switches it to a no-astropy state and avoids the double conversion.
I'll test if this is the case if `astropy-units=True` in my `.galpyrc`

nstarman:
Added some `if hasattribute(x, 'unit')` checks to the functions and then strip the astropy units. Now it should be compatible with `astropy-units=True`

jobovy:
Thanks, but I'm confused by all these latest changes, why can't you just add ``use_physical=False`` to the calls  of ``evaluatePotentials`` and ``vcirc`` to avoid getting a quantity? Should really just be a pretty atomic change of the existing code I think.

nstarman:
How do I collapse the merge into the previous commit? I was trying to keep the commit history as short as possible.

nstarman:
hmm. The test that I wrote does not actually capture some of the behavior. I implemented the suggestion to ``` just add use_physical=False to the calls of evaluatePotentials and vcirc to avoid getting a quantity```, but this does not actually work. For `circular_velocity`, the returned quantity may be different in galpy versus amuse.

```python
from galpy.util import config
from astropy import units as any_u
from galpy import potential
from galpy.potential import to_amuse
from amuse.units import units as u
print('astropy: ', config.__config__.getboolean('astropy', 'astropy-units'))

amp_u= 1e8 * apy_u.solMass
a_u= 0.8 * apy_u.kpc
ro_u, vo_u= 8.* apy_u.kpc, 220.* apy_u.km / apy_u.s
pot2= potential.TwoPowerSphericalPotential(amp=amp_u, a=a_u, ro=ro_u, vo=vo_u)
print('galpy: ', pot2.vcirc(1*apy_u.kpc))
amuse_pot2= to_amuse(pot2)
v2= amuse_pot2.circular_velocity(1|u.kpc)
print('amuse: ', v2)
```
results in
```
astropy:  False
galpy:  12.1664448254
amuse:  0.0553020219337 kms
```

If I switch all the flags from `False` to `True` (except in `set_gravity_at_point`) in `galpy_profile` then it works as advertised. Working on it now.

jobovy:
Don't worry about it, I will just squash them all into 1 commit at merge
(like for the previous PR)..

galpy/potential/amuse.py
```diff
@@ -179,9 +188,10 @@ def circular_velocity(self,r):
         HISTORY:
            2019-08-12 - Written - Webb (UofT)
         """
-        return potential.vcirc(self.pot,r.value_in(units.kpc)/self.ro,phi=0,
-                               t=self.tgalpy,ro=self.ro,vo=self.vo) | units.kms
-        
+        return potential.vcirc(self.pot,r.value_in(u.kpc)/self.ro,phi=0,
```
jobovy: I think this needs to be multiplied by ``self.vo`` to work properly (right now the galpy output is in internal units and is then given AMUSE kms units)

galpy/potential/amuse.py
```diff
@@ -195,9 +205,10 @@ def enclosed_mass(self,r):
         HISTORY:
            2019-08-12 - Written - Webb (UofT)
         """
-        vc2= potential.vcirc(self.pot,r.value_in(units.kpc)/self.ro,phi=0,
-                             t=self.tgalpy,ro=self.ro,vo=self.vo)**2.
-        return vc2*r.value_in(units.kpc)/bovy_conversion._G*1000. | units.MSun
+        vc2= potential.vcirc(self.pot,r.value_in(u.kpc)/self.ro,phi=0,
+                            t=self.tgalpy,ro=self.ro,vo=self.vo,
+                            use_physical=False)**2.
+        return vc2*r.value_in(u.kpc)/bovy_conversion._G*1000.|u.MSun
```
jobovy: Should probably check the units here, not sure this is correct from looking at it (probably not)

galpy/potential/amuse.py
```diff
@@ -158,13 +166,14 @@ def mass_density(self,x,y,z):
...
+        R= numpy.sqrt(x.value_in(u.kpc)**2.+y.value_in(u.kpc)**2.)
+        zed= z.value_in(u.kpc)
+        phi= numpy.arctan2(y.value_in(u.kpc),x.value_in(u.kpc))
         return potential.evaluateDensities(self.pot,R/self.ro,zed/self.ro,
                                            phi=phi,t=self.tgalpy,
-                                           ro=self.ro,vo=self.vo) \
-                                           | units.MSun/(units.parsec**3.)
+                                           ro=self.ro,vo=self.vo,
+                                           use_physical=False
+                                           ) | u.MSun/(u.parsec**3.)
```
jobovy: This also needs a ``bovy_conversion.dens_in_msolpc3(self.vo,self.ro)``

galpy/potential/amuse.py
```diff
@@ -98,13 +101,14 @@ def get_potential_at_point(self,eps,x,y,z):
...
+        R= numpy.sqrt(x.value_in(u.kpc)**2.+y.value_in(u.kpc)**2.)
+        zed= z.value_in(u.kpc)
+        phi= numpy.arctan2(y.value_in(u.kpc),x.value_in(u.kpc))
         return potential.evaluatePotentials(self.pot,R/self.ro,zed/self.ro,
                                             phi=phi,t=self.tgalpy,
-                                            ro=self.ro,vo=self.vo) \
-                                            | units.km**2*units.s**-2
+                                            ro=self.ro,vo=self.vo,
+                                            use_physical=False
+                                            ) | u.km**2*u.s**-2
```
jobovy: Think this needs a ``*self.vo**2`` to get the right units

nstarman:
Ok. changing the flags doesn't do the trick. Having `use_physical=False` results in galpy and amuse disagreeing if quantities with units are passed to galpy or amuse, regardless of whether `astropy-units` is true or false. The only viable solution I have found is to let the potential look up its own `physical` flag and then check for units on the output.

galpy/potential/amuse.py
nstarman: yes. having `use_physical=False` and multiplying by `self.vo` appears to work.

galpy/potential/amuse.py
nstarman: That also works.

galpy/potential/amuse.py
nstarman: done

galpy/potential/amuse.py
nstarman: All it needs is to change `vc2= potential.vcirc(self.pot,r.value_in(u.kpc)/self.ro,phi=0,
                            t=self.tgalpy,ro=self.ro,vo=self.vo,
                            use_physical=False)**2.` to `self.circular_velocity(r)**2.` Then the corrections to those units sort everyone else out.

nstarman:
I implemented the `use_physical=False` and then corrected for all the galpy internal unit conversions. I tested each function with a function like this and galpy and AMUSE now agree. The test function now checks `get_potential_at_point`, `get_gravity_at_point`, `mass_density`, `circular_velocity`, and `enclosed_mass` for potentials instantiated with and without units and compares the galpy & AMUSE outputs. I don't set `astropy-units` as I think the Travis test matrix has a test-version for both `True` and `False`.

Screen Shot 2019-11-08 at 10 02 03

nstarman:
Also, the Travis build failed because `wget ftp://ftp.gnu.org/gnu/gsl/gsl-2.4.tar.gz; fi"` timed out. I can't restart a single test. Can you?

galpy/potential/amuse.py
nstarman: I've implemented this in a new commit.

galpy/potential/amuse.py
nstarman: I've implemented this in a new commit.

galpy/potential/amuse.py
nstarman: I've implemented this in a new commit.

galpy/potential/amuse.py
nstarman: I've implemented this in a new commit.

galpy/potential/amuse.py
```diff
@@ -1,10 +1,13 @@
...
 import numpy
-from amuse.units import units
+from amuse.units import units as u
 from amuse.units.quantities import ScalarQuantity
 from amuse.support.literature import LiteratureReferencesMixIn
 from .. import potential
 from ..util import bovy_conversion
+from .Force import _APY_LOADED
+if _APY_LOADED:
+    from astropy import units as apy_u
```
jobovy: I think this now all looks good, but I don't think ``astropy.units`` is necessary anymore, so could you please switch back to the old form (where ``amuse.units`` is imported as ``units``)?

Also, the AMUSE tests aren't actually run on travis, because it's too difficult to install AMUSE on travis, so we just run these locally from time to time to make sure that they pass.

nstarman: done

jobovy:
All looks okay to me now, but I tried running
```
pytest -v tests/test_amuse.py -xs
```
on the server and it has a lot of errors (about ``np`` as ``numpy`` not existing, the units not being imported, and it also seems like you use ``units.pc`` in AMUSE both in the test and in the actual ``amuse.py`` code, which doesn't seem to exist). Could you make sure that the test runs on the server before I merge this?

nstarman:
Now it should work. Ran the tests on the server and on my different AMUSE installs.

jobovy:
Did you push any updates?

jobovy:
Found a few more ``units.pc`` in the tests that I changed to ``units.parsec``; now all looks good!

PR #406, opened 2019-11-10
Copyable potentials

nstarman:
adds an attribute `init_args` to every potential which holds the arguments used to initialize a potential.  This allows for easy copying of a potential (added. as `.copy()`) as well as initializing potentials with attributes modified from another potential. For example, taking a potential and making a new potential which is an exact copy except 3x the amplitude.


nstarman:
my commits are not triggering new travis builds

PR #414, opened 2020-01-20
Improvement orbits animation performance

henrysky:
I refactored the JavaScript to improve orbits animation performance. Doing extend traces and restyling for all traces at once seems to improve quite a lot! At least animating all solar system planets together is smooth now!

Not sure what ``update_trace34``, ``update_trace56`` are doing but I updated them the same way as ``update_trace12`` and hope they work correctly, they are probably some way of animating orbits I don't know?

### DIFF
--- a/galpy/orbit/Orbits.py
+++ b/galpy/orbit/Orbits.py
@@ -5036,34 +5036,26 @@ class Orbit(object):
 """.format(trace_indx=str(ii),trace_num_1=str(2*ii+1),trace_num_2=str(2*ii+2),
            line_color=line_colors[ii])
             traces_cumul+= """,trace{trace_num_1},trace{trace_num_2}""".format(trace_num_1=str(2*ii+1),trace_num_2=str(2*ii+2))
-        update_trace12= """
-      Plotly.extendTraces('{divid}', {{
-        x: [data.x1_0.slice(trace_slice_begin,trace_slice_end)],
-        y: [data.y1_0.slice(trace_slice_begin,trace_slice_end)],
-      }}, [0]);
-      trace_slice_begin-= trace_slice_len;
-      trace2= {{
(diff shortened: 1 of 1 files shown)


jobovy:
Cool, thanks! I can believe that this would improve performance, I only recently learned how to update traces simultaneously... Will have to look at this more closely later to make sure I understand the changes, but the ``update_trace34`` and ``update_trace56`` are for when you have multiple panels, so when you do ``o.animate(d1=['R','x'],d2=['z','y'])`` (for two panels) or ``o.animate(d1=['R','x','R'],d2=['z','y','R'])`` (for three panels). Could you test those cases (automated tests don't test the animation code).

henrysky:
Okay I updated so doing all panels together at once, it seems fine to me!

galpy/orbit/Orbits.py
```diff
@@ -5037,34 +5036,30 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+                trace_num_10_list += """{trace_num_10}, """.format(trace_num_10=str(2*jj*self.size + 2 * ii + 1 - 1))
+                trace_num_20_list += """{trace_num_20}, """.format(trace_num_20=str(2*jj*self.size + 2 * ii + 2 - 1))
+
+        master_traces = """
+        traces = {{x: [{x_data_list}], y: [{y_data_list}]
+        }}
+        """.format(x_data_list=x_data_list, y_data_list=y_data_list)
+        update_trace12 = """
+        {master_traces}
+        trace_slice_begin+= trace_slice_len;
```
jobovy: I don't think this should be here (because in the original, this was as part of the second trace to extend, to make up for the ``trace_slice_begin-= trace_slice_len;`` just before the ``Plotly.restyle`` of the first trace. Not sure how the code actually works in this case though, because I think the trace you are extending should have zero length...

PR #430, opened 2020-09-03
Replace Quantity parsing through the package as much as possible with a set of centralized parsers

jobovy:
Probably makes #428 unnecessary...


jobovy:
Net difference of -443 lines 🎉 🎉

PR #433, opened 2020-10-02
Adds starting index to range of cml_pvr used to make interpolator

jamesmlane:
Uses start_indx to set minimum value of cml_pvr used to make 1D cumulative p(v|r) -> v_vesc interpolators. Ensures when the DF is negative or 0 for a non-zero v_vesc that the interpolator can properly initialize and be used to sample v_vesc.


jobovy:
Looks good, thanks! I may add a warning that the DF appears to have negative parts before merging...

PR #434, opened 2020-11-03
Add implementation of einsum into covariance matrix propagation

jmackereth:
This should add the necessary functions and tests to speed up the covariance matrix propagation using Einstein summation implemented in numpy. Since I left this to go stale for so long, I've had to incorporate updates to the master branch a bunch of times, so hopefully nothing broke in that process...


jmackereth:
I've had this branch open for longer than 2 years - whoops...

jobovy:
Thanks for the PR! I have to look in more detail, but two quick comments to start:

- [ ] There's a test that fails (e.g., [here](https://travis-ci.org/github/jobovy/galpy/jobs/741215554#L3288)). To debug, you should be able to do `pytest -v test/test_coords.py -k test_cov_pmradec_to_pmllbb` to just run this test
- [ ] Is there any reason *not* to use `einsum`? That is, is there any point to the `no_einsum` option? If it works and is faster, then we should probably just implement that.

jmackereth:
Ok - I have fixed those first two issues. Something that may be an issue is that the commit history got a bit messy because I had to re-base quite a few times (and things got confusing when we went from bovy_coords -> coords). 

The tests seem to pass ok for now, though!

galpy/util/coords.py
```diff
@@ -742,20 +742,70 @@ def cov_pmrapmdec_to_pmllpmbb(cov_pmradec,ra,dec,degree=False,epoch=2000.0):
...
-                                                      degree,epoch)
+        out = cov_pmradec_to_pmllbb_array(cov_pmradec,
+                                          ra, dec, lb[:,1],
+                                          degree, epoch)
         return out
     else:
-        l,b = radec_to_lb(ra,dec,degree=degree,epoch=epoch)
-        return cov_pmradec_to_pmllbb_single(cov_pmradec,ra,dec,b,degree,epoch)
+        lb = radec_to_lb(ra,dec,degree=degree,epoch=epoch)
+        return cov_pmradec_to_pmllbb_single(cov_pmradec,ra,dec,lb[1],degree,epoch)
```
jobovy, line 756: Could these be combined into a single case (i.e., use the same `_array` function for both a single object and multiple objects)? That would also help with the test coverage, because right now the radian input case for arrays is not tested.

galpy/util/coords.py
```diff
@@ -856,16 +907,56 @@ def cov_dvrpmllbb_to_vxyz(d,e_d,e_vr,pmll,pmbb,cov_pmllbb,l,b,
...
     else:
-        ndata= len(d)
-        out= numpy.zeros((ndata,3,3))
-        for ii in range(ndata):
-            out[ii,:,:]= cov_dvrpmllbb_to_vxyz_single(d[ii],e_d[ii],e_vr[ii],
-                                                      pmll[ii],pmbb[ii],
-                                                      cov_pmllbb[ii,:,:],
-                                                      l[ii],b[ii])
-
+        out = cov_dvrpmllbb_to_vxyz_array(d,e_d,e_vr,pmll,pmbb,cov_pmllbb,l,b)
```
jobovy, line 910: Again, could the array case just handle the single case as well?

galpy/util/coords.py
```diff
@@ -900,6 +991,119 @@ def cov_dvrpmllbb_to_vxyz_single(d,e_d,e_vr,pmll,pmbb,cov_pmllbb,l,b):
...
+       propagate uncertainties in vxyz to galactocentric cylindrical coordinates
+    INPUT:
+       cov_vxyz - uncertainty covariance in U,V,W
+       phi - angular position of star in galactocentric cylindrical coords
+    OUTPUT:
+       cov(vR,vT,vz) [3,3]
+    HISTORY:
+       2018-03-22 - Written - Mackereth (LJMU)
+       2020-09-21- Moved to coords.py - Mackereth (UofT)
+    """
```
jobovy, line 1008: This is a new function? If so, could you add it to the API docs in `docs/source/reference/coords.rst` similar to other functions?
None of these following new functions are hit by any tests, see the report [here](https://codecov.io/gh/jobovy/galpy/pull/434/src/galpy/util/coords.py?before=galpy/util/coords.py#L1009).
Also same comment re:single vs. array as before...

jobovy (changes requested):
Looks good, could you summarize what functions you changed and what functions you added? I also have a bunch of comments on the implementation, testing, and documentation...

If you think the commit history is too messy, we could also just squash all commits into a single one (that's what I normally do when it's truly too messy). It looks fine wrt `bovy_coords` vs. `coords` to me...

jobovy (approved):

jobovy:
All looks good now 🎉 🎉

PR #436, opened 2020-11-12
Don't flatten Potential instances

nstarman:
first check its a Potential

### DIFF
--- a/galpy/potential/Potential.py
+++ b/galpy/potential/Potential.py
@@ -3257,7 +3257,9 @@ def flatten(Pot):
         2018-03-14 - Written - Bovy (UofT)
     
     """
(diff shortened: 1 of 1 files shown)


nstarman:
I also got CompositePotential to completely work. It operates completely like a list. should I push in different PR?

nstarman:
See https://github.com/nstarman/galpy/tree/composite_potentials

jobovy:
Thanks, this is a very straightforward way to fix the issue you're having while not affecting any of the other code!

Let's hold off for now on the `CompositePotential`, I need to think more about how that would mesh with the rest of the code.

With travis slow these days, it will be a while until the tests all run...

PR #443, opened 2021-01-10
pass random state to sphericaldf.sample

nstarman:
Allow the random generator to be specified when sampling.

### DIFF
--- a/galpy/df/sphericaldf.py
+++ b/galpy/df/sphericaldf.py
@@ -33,6 +33,9 @@ from ..util.conversion import physical_conversion
 if conversion._APY_LOADED:
     from astropy import units
 
+
+_default_random = numpy.random.RandomState(seed=None)  # random seed
+
 class sphericaldf(df):
     """Superclass for spherical distribution functions"""
     def __init__(self,pot=None,scale=None,ro=None,vo=None):
@@ -280,7 +283,7 @@ class sphericaldf(df):
         return 1.-self._vmomentdensity(r,0,2)/2./self._vmomentdensity(r,2,0)
     
 ############################### SAMPLING THE DF################################
-    def sample(self,R=None,z=None,phi=None,n=1,return_orbit=True):
+    def sample(self,R=None,z=None,phi=None,n=1,return_orbit=True, random=None):
         """
         NAME:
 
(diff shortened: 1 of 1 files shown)


jobovy:
Bunch of the tests fail, mostly I think because they rely on a different way of setting the random state. Why not just set the random state just before calling the sampling function like in the tests (there with `numpy.random.seed(..)`)?

nstarman:
I do that currently, I just thought it might be more user-facing to allow passing the random state. For instance, with the new numpy ``default_rng``  and  ``Generator`` implementation a lot of the stuff in ``numpy.random`` is deprecated or considered not-advised

```
Many of the RandomState methods above are exported as functions in numpy.random <https://numpy.org/doc/stable/reference/random/index.html#module-numpy.random> This usage is discouraged, as it is implemented via a global RandomState <https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState> instance which is not advised on two counts:

It uses global state, which means results will change as the code changes

It uses a RandomState <https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState> rather than the more modern Generator <https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator>.

For backward compatible legacy reasons, we cannot change this. See random-quick-start.

```

nstarman:
I do that currently, I just thought it might be more user-facing to allow passing the random state. For instance, with the new numpy ``default_rng``  and  ``Generator`` implementation a lot of the stuff in ``numpy.random`` is deprecated or considered not-advised

> Many of the RandomState methods above are exported as functions in numpy.random <https://numpy.org/doc/stable/reference/random/index.html#module-numpy.random> This usage is discouraged, as it is implemented via a global RandomState <https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState> instance which is not advised on two counts:
> 
> It uses global state, which means results will change as the code changes
> 
> It uses a RandomState <https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState> rather than the more modern Generator <https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator>.
> 
> For backward compatible legacy reasons, we cannot change this. See random-quick-start.

jobovy:
Interesting, I didn't know about that change! I think we should keep this on the to-do list as part of a more general change of the use of `numpy.random` throughout the code (`sphericaldf` isn't the only place random numbers are used). Also, this was only introduced in numpy v1.17 and we can only drop 1.16 support t[omorrow](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) 😄

PR #447, opened 2021-03-10
Added the Cautun20 potential

MariusCautun:
Added the Cautun20 potential using the McMillan17 example. The potential is based on Cautun et al (2020) https://ui.adsabs.harvard.edu/abs/2020MNRAS.494.4291C/abstract.

Added the potential in the list of MW potentials, a test function, and also in the .rst file that gives the list of MW potentials.


jobovy:
Thanks Marius, this looks great! I'll take a close look sometime in the next few days, but given that the tests pass and all of the new code is covered by the tests, I don't anticipate any major hiccups.

MariusCautun:
Jo, that is great! Let me know if you have questions or would like me to make any changes.

jobovy:
Hi Marius,

I took a closer look at this and here are some comments/questions:

* For the CGM density, I think you just use a spherical power-law with a cut-off at R200? If that's the case, I think you could just use a [`PowerSphericalPotentialwCutoff`](https://docs.galpy.org/en/latest/reference/potentialpowerspherwcut.html). That would give a small speed-up
* I'm a little confused about your exponential cut-offs (CGM and also DM): why do you use e^(1-r/r200) [or e^(1-[r/r200]^2) for the DM]? That increases the density by e at r200 before dropping exponentially.
* I'm wondering whether we could easily implement your adiabatic-contraction prescription more generally, so I have a few questions about how it's computed:
  * For axisymmetric densities, you compute M(<r) as the actual integral over the spherical region rather than, say, do M(<R)? I'm asking because the latter would be a bit easier to implement using the existing `Potential.mass` method.
  * I think you use your eqn. (11), but would it be better to directly use eqn. (10)? I don't think that would be harder.
  * For a more general implementation, I think I would assume users have just accounted for fbar in the input densities. Assuming that, are the formulae valid just for taking a DM halo with M_DM(<r) and growing a baryonic distribution M_bar(<r) and then M_DM_contract(<r) is given by applying eqn. (10/11)? That is, without having to worry about fbar? 
  * I think a better way to implement the contracted halo would be to use a [`interpSphericalPotential`](https://docs.galpy.org/en/latest/reference/potentialinterpsphere.html), then you can directly use the enclosed mass rather than having to compute the density. And it will be much faster than using SCF. 

Let me know what you think. If we're going to do the more general implementation, I'm happy to do some of the heavy lifting (but it may be useful for you to provide some tests of just the contraction).

MariusCautun:
Hi Jo,

Thank you very much for going carefully over the code and for the suggestions! I revised the code accordingly and here are my thoughts on your comments:

1. Changed the CGM potential to  PowerSphericalPotentialwCutoff.

2. You are right, there was an error in CGM and halo cut-offs. Nice catch! Changed it now to e^(-[r/2*r200]^2) for CGM and there is no need for a halo cut-off using the new potential implementation.

3. Changed the DM halo to the interpSphericalPotential class. It wasn't fully clear to me what were the units for the radial force, so I had to implement a workaround: first represented an NFW halo using this class to find the normalization factor. Maybe you could make this aspect clearer in the documentation.

4. I think would be great to implement the adiabatic contraction, such as my formula, as well as the older Blumenthal 1986 and Gnedin 2004. For all of the prescriptions, you need the spherically averaged baryonic enclosed mass since the effect depends on the change in potential assuming spherical symmetry. From what I know, nobody looked into how the DM halo contraction changes as a function of (R,z). The only exception is when you assume adiabatic conservation of actions, which you can calculate in the axisymmetric potential of the baryons. However, the actions of DM particles are not very well preserved between the DM-only and the hydrodynamics cases, for example, see [Callingham et al (2020)](https://ui.adsabs.harvard.edu/abs/2020MNRAS.495...12C/abstract), and can result in ~10% errors in the predicted DM halo profile.

5. My Eq. (11) reproduces surprisingly well my  Eq. (10) -- if I remember correctly to better than 1% accuracy or even higher (I was also very surprised by that) -- so really it does not matter which one you end up using. I prefer eq. 11 since it is easier to implement and was much faster when calculating the MCMC chains.

6. You do need fbar to calculate the adiabatic contraction. This is because the halo might contain a different fraction of baryons than the universal value. One needs to calculate the change in potential between the DM case, where the baryons account for (1-fbar) of the mass and are modelled as following the same radial profile as the DM, and the hydro case, where the baryonic fraction of the halo is different and moreover their spatial distribution is more concentrated. 

7. One could implement the contraction by either assuming a default fbar (e.g. the latest Planck value) -- I think this value is rather robust nowadays, at least for Galactic dynamic purposes -- or, alternatively, calculate fbar from the input baryonic components by assuming that all baryons are kept in the halo.  

8. If you want, I have python implementations calculating the Blumenthal 1986, Gnedin 2004, and Cautun 2020 baryonic contraction expressions (they return the fractional change in enclosed DM mass as a function of radius) that could then be implemented in galpy using, for example, the interpSphericalPotential potential. Alternatively, we could augment the NFW_potential (as well as other classes) to multiply the force, enclosed mass, densities, and so on by a "halo contraction factor". This should be straightforward for the mass and force, and would need some extra numerical calculations for the potential and densities.

jobovy:
Okay, thanks! I will try to implement a simple version of the adiabatic contraction that we can iterate on and add the Gnedin/Blumenthal prescriptions to. 

The units for the radial force in `interpSphericalPotential` should be internal units, so you'd have to divide the force in physical units using one of the force conversion routines in `galpy.util.conversion`. I can fix that later.

Don't worry about the failing test, I think that's just related to a minor issue in the way the enclosed mass is computed in `interpSphericalPotential`, which I think I've already fixed in `master`.

jobovy:
Hi Marius,

Okay, I quickly implemented a first version of the general adiabatic-contraction potential, see here:

https://github.com/jobovy/galpy/compare/adiabatic-contraction

You can use it as follows, e.g.,
```
dm= AdiabaticContractionWrapperPotential(pot=MWPotential2014[2],baryonpot=MWPotential2014[:2])
```
to contract the DM halo in MWPotential2014.

Comparing it to your implementation in this PR, they seem to agree well (at least when I up the number of interpolation points used in your implementation, which is rather low), to better than a percent at r > 0.2 (they are a few percent off at r ~ 0.1; I'm not sure why, probably due to the interpolation).

Because your `master` is now ahead of galpy's main `master` branch, it's probably difficult for you to get it in your fork, but you could just download the zip file at https://github.com/jobovy/galpy/tree/adiabatic-contraction and install that. Or you could branch off of your master using
```
git checkout -b adiabatic-contraction-cautun
git merge https://github.com/jobovy/galpy.git adiabatic-contraction
```
which would merge this PR with the new adiabatic-contraction branch without changing this PR immediately. Then you could easily test the original implementation and the new one.

Are the Blumenthal and Gnedin formulations similar? That is, could we just implement them using a variation of the eta_bar factor? If so, could you send me the relevant code?

If you think this looks good, I can try to merge the adiabatic-contraction wrapper first and then edit this PR to use that wrapper.

jobovy:
Hi Marius,

Because the adiabatic-contraction wrapper seems to be working, I've gone ahead and re-written your `Cautun20` potential using it and pushed this to this PR (if you want to make any further changes, make sure to pull changes from your master branch, `git pull origin master`). Because my adiabatic-contraction implementation does a better job at the mass interpolation, the potential is slightly changed, so I tweaked the input mass and concentration of the halo a bit to make the tests largely pass (I can't get the local DM density one to pass, that density always comes out about a percent lower, but I think that's probably alright).

If you could take a look that would be useful. To finish this, I'd like to merge the adiabatic-contraction wrapper first and ideally we could add the Blumenthal and Gnedin prescriptions to that first. If that's not too difficult.

MariusCautun:
Hi Jo,

Thank you for your fast reply, all the changes, and the help! Sorry for not replying faster, but it seems that always something else fights for my attention.

I added the Blumenthal and the Gnedin adiabatic contraction methods in the  [AdiabaticContractionWrapperPotential file](galpy/potential/AdiabaticContractionWrapperPotential.py) and modified the previous implementation of my contraction to Eq. 10 in my paper. The latter was motivated by recent tests, during which I found out that while the previous implementation, i.e. the fit in my Eq. 11, works very well for the Planck (2015) baryonic fraction, gives bad approximations when changing the universal value of the baryonic fraction. 

The three halo contraction methods involve several lines of code, so they are implemented as self-standing functions inside the [AdiabaticContractionWrapperPotential file](galpy/potential/AdiabaticContractionWrapperPotential.py). I am not sure which format you like for that, so feel free to change the format as you like.

Also, I made a few changes to the numerical integration code to find 'Phi0':
* dealt with some numerical convergence warning by numerically integrating only up to 'rmax'
* used the 'scipy.interpolate.interp1d' function which I think should be faster than the numpy version (I think numpy recalculates the interpolation function each time you call it)
* changed the innermost point from 'rgrid[0]=0' to 'rgrid[0]=1.e-4' since the previous value raised 'Nan' that affected several array elements for the 'blumenthal' and 'gnedin' methods.

**One thing that is missing is a warning to users when they insert an invalid value for the 'method' argument in the 'AdiabaticContractionWrapperPotential' class.**

I also added some extra test conditions in the 'test_AdiabaticContractionWrapper' function in the [test_potential file](tests/test_potential.py). For example, by comparing the results for the 'dm1' and 'dm2' potentials, you can see the effect of changing the universal baryonic fraction. Comparing 'dm2', 'dm3' and 'dm4', you can see the effect of different contraction methods. The comparison should be easy since I inserted the enclosed mass increase factor for all 4 cases (i.e. 'dm1' to 'dm4') at a few different radii. 

From a physical point, I think the default value for 'f_bar' in 'AdiabaticContractionWrapperPotential' should be set to one the recent measurements, such as the Planck 2015 value of 'f_bar=0.157' -- this philosophy corresponds to the methodology used to derive the Gnedin 2004 and Cautun 2020 expressions for the contraction. Calculating 'f_bar' using the mass enclosed within 'rmax' results, typically, in significantly lower values since most models do not include a CGM component, which accounts for a large share of the baryonic mass at large distances. For example, for the 'MWPotential2014' with 'rmax=50.' one obtains 'f_bar~0.06' (don't remember exactly the value), which is ~2.5 times lower than what we think is the universal baryon fraction (e.g. Planck 2015 value of 0.157). When using the lower  'f_bar~0.06', one tends to overestimate the contraction of the DM halo in the inner parts where the baryons are dominant -- e.g. compare the 'dm1' and 'dm2' potentials in the 'test_AdiabaticContractionWrapper' function in the [test_potential file](tests/test_potential.py).

jobovy:
Apologies for the delay, I'll hopefully be able to take another close look in the following days, but on the whole it now looks pretty straightforward, thanks!

PR #450, opened 2021-03-31
Including a wrapper to allow the z-axis vector of a potential to be adjusted (e.g. inclined discs, tilted bar)

jmackereth:
I started having a go at implementing this wrapper potential which allows an adjustment to the z-axis of a potential. This will allow for potentials to be tilted around (in a similar way to the transformations currently allowed when using `EllipsoidalPotential` type potentials). Starting the PR to keep my momentum up!

The python implementation of the potential seems to be working as expected, but including the C code has given me some hiccups (which I think will cause the tests to fail!). I can't see why - but orbit integration in C doesn't seem to want to work. This is my first foray into C, so that might explain it.

I need to add testing of all the added things - once that's done, hopefully this should merge in easily...

### DIFF
--- a/galpy/orbit/orbit_c_ext/integrateFullOrbit.c
+++ b/galpy/orbit/orbit_c_ext/integrateFullOrbit.c
@@ -507,7 +507,7 @@ void parse_leapFuncArgs_Full(int npot,
       potentialArgs->Rforce= &RotateAndTiltWrapperPotentialRforce;
       potentialArgs->zforce= &RotateAndTiltWrapperPotentialzforce;
       potentialArgs->phiforce= &RotateAndTiltWrapperPotentialphiforce;
-      potentialArgs->nargs= (int) 10;
+      potentialArgs->nargs= (int) 16;
       potentialArgs->requiresVelocity= false;
       break;
     }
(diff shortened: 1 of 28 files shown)


galpy/orbit/orbit_c_ext/integrateFullOrbit.c
```diff
@@ -507,7 +507,7 @@ void parse_leapFuncArgs_Full(int npot,
       potentialArgs->Rforce= &RotateAndTiltWrapperPotentialRforce;
       potentialArgs->zforce= &RotateAndTiltWrapperPotentialzforce;
       potentialArgs->phiforce= &RotateAndTiltWrapperPotentialphiforce;
-      potentialArgs->nargs= (int) 2;
+      potentialArgs->nargs= (int) 10;
```
jmackereth, line 510: I think it's possible that the reason that the C implementation is still not working (orbit integration hangs, or takes a longer time than my patience limit...) is something to do with this... its the only thing I can think of.

The compilation seems to work fine now.

jmackereth:
I've put together a gist [here](https://gist.github.com/jmackereth/d431c078555c436b4560a146f851d46c) running through some tests - looks like things are doing what I expect, at least for the python implementation.

Still to do:

- [ ] consider just adding the x,y,z force computations to Force.py etc.
- [ ] same as above for C code...
- [ ] tests

galpy/orbit/orbit_c_ext/integrateFullOrbit.c
jobovy: This should be the number of parameters that you're passing that are specific to the wrapper, which from `integrateFullOrbit.py` looks to be 10

PR #454, opened 2021-04-27
expand Orbit init to accept tuples

nstarman:
Hi @jobovy, just a small expansion to allowed input types.


jobovy:
Hi @nstarman--Okay, I think should work fine, but I'm not sure about the rationale for not just using a list? Also, it would be better to include a test of the new setup to make sure it keeps working.

nstarman:
It was just a Postel's law thing. I accidentally encountered the invalid input type when doing a sanity test of an orbit. I'll write a quick test.

PR #469, opened 2022-03-15
Include an offset option in RotateAndTiltWrapperPotential

jmackereth:
adds an `offset` option that allows the RotateAndTiltWrapper potential to also be offset in Cartesian coordinates, after the rotation and tilt is applied. I think the test should cover all the added code - not sure if there is sufficient testing of the offset as added to the C code, though. Will add that if necessary!

### DIFF
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -311,7 +311,7 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
             pot_args.extend([p._amp])
             pot_args.extend([0.,0.,0.,0.,0.,0.]) # for caching
             pot_args.extend(list(p._rot.flatten()))
-            pot_args.append(not (p._rot == numpy.eye(3)).all())
+            pot_args.append(not p._norot)
             pot_args.append(not p._offset is None)
             pot_args.extend(list(p._offset) if not p._offset is None else [0.,0.,0.])
     pot_type= numpy.array(pot_type,dtype=numpy.int32,order='C')
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -311,6 +311,7 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
             pot_args.extend([p._amp])
             pot_args.extend([0.,0.,0.,0.,0.,0.]) # for caching
             pot_args.extend(list(p._rot.flatten()))
+            pot_args.append(not (p._rot == numpy.eye(3)).all())
             pot_args.append(not p._offset is None)
             pot_args.extend(list(p._offset) if not p._offset is None else [0.,0.,0.])
     pot_type= numpy.array(pot_type,dtype=numpy.int32,order='C')
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -21,6 +21,14 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
     #Figure out what's in pot
     if not isinstance(pot,list):
         pot= [pot]
+    if (potforactions or potfortorus) \
+        and ( (len(pot) == 1 and isinstance(pot[0],potential.NullPotential)) 
+             or numpy.all([isinstance(p,potential.NullPotential) for p in pot]) ):
+        raise NotImplementedError("Evaluating actions using the C backend is not supported for NullPotential instances")
+    # Remove NullPotentials from list of Potentials containing other potentials
+    purged_pot= [p for p in pot if not isinstance(p,potential.NullPotential)]
+    if len(purged_pot) > 0:
+        pot= purged_pot
     #Initialize everything
     pot_type= []
     pot_args= []
@@ -216,6 +224,9 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
                              p._Phi0,p._Phimax])
         # 37: TriaxialGaussianPotential, done with others above
         # 38: PowerTriaxialPotential, done with others above
+        elif isinstance(p,potential.NullPotential):
+            pot_type.append(40)
+            # No arguments, zero forces
         ############################## WRAPPERS ###############################
         elif isinstance(p,potential.DehnenSmoothWrapperPotential):
             pot_type.append(-1)
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -300,6 +300,7 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
             pot_args.extend([p._amp])
             pot_args.extend([0.,0.,0.,0.,0.,0.]) # for caching
             pot_args.extend(list(p._rot.flatten()))
+            pot_args.extend(list(p._offset))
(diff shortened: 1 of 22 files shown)


galpy/potential/RotateAndTiltWrapperPotential.py
```diff
@@ -72,6 +77,7 @@ def __init__(self,amp=1.,inclination=None,galaxy_pa=None,sky_pa=None,
         galaxy_pa= conversion.parse_angle(galaxy_pa)
         zvec, galaxy_pa= self._parse_inclination(inclination,sky_pa,
                                                  zvec,galaxy_pa)
+        self._offset= self._setup_offset(offset)
```
jobovy, line 80:
```suggestion
        self._offset= conversion.parse_length(offset,ro=self._ro)
```

jobovy (changes requested):

galpy/orbit/integrateFullOrbit.py
```diff
@@ -300,6 +300,8 @@ def _parse_pot(pot,potforactions=False,potfortorus=False):
             pot_args.extend([p._amp])
             pot_args.extend([0.,0.,0.,0.,0.,0.]) # for caching
             pot_args.extend(list(p._rot.flatten()))
+            pot_args.append(not p._offset is None)
+            pot_args.extend(list(p._offset))
```
jobovy, line 304: I think because of my other suggestions, this now has to be

```suggestion
            pot_args.extend(list(p._offset) if not p._offset is None else [0.,0.,0.])
```

which I think should be valid Python? Probably want to check...

tests/test_potential.py
```diff
@@ -3640,6 +3640,11 @@ def test_RotateAndTiltWrapper():
     NFW_wrapped= potential.RotateAndTiltWrapperPotential(galaxy_pa=galaxy_pa, pot=potential.TriaxialNFWPotential(amp=1.,b=0.7,c=0.5))
     NFW_rot= potential.TriaxialNFWPotential(amp=1., zvec=[0.,0.,1.],pa=galaxy_pa,b=0.7,c=0.5)
     assert (evaluatePotentials(NFW_wrapped, Rphiz_test[0], Rphiz_test[2], phi=Rphiz_test[1])-evaluatePotentials(NFW_rot, Rphiz_test[0], Rphiz_test[2], phi=Rphiz_test[1])) < 1e-6, 'Wrapped and Internally rotated NFW potentials do not match when evaluated at the same point'
+    #make sure the offset works as intended
+    # triaxial NFW at x,y,z = [20.,0.,3.]
+    NFW_wrapped= potential.RotateAndTiltWrapperPotential(zvec=zvec, galaxy_pa=galaxy_pa, offset=[20.,0.,3.], pot=potential.TriaxialNFWPotential(amp=1.,b=0.7,c=0.5))
+    NFW_rot= potential.TriaxialNFWPotential(amp=1., zvec=zvec, pa=galaxy_pa,b=0.7,c=0.5)
+    assert (evaluatePotentials(NFW_wrapped, 0., 0., phi=0.)-evaluatePotentials(NFW_rot, 20., - 3., phi=numpy.pi)) < 1e-6, 'Wrapped + Offset and Internally rotated NFW potentials do not match when evaluated at the same point'
```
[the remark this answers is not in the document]
jobovy: Maybe in `test_potential.py`? Should probably run it both in C and Python.

galpy/orbit/integrateFullOrbit.py
jmackereth: included!

jmackereth:
Ok, the C code is indeed failing the test, so just working on that now!

tests/test_potential.py
```diff
@@ -3645,6 +3646,62 @@ def test_RotateAndTiltWrapper():
     NFW_wrapped= potential.RotateAndTiltWrapperPotential(zvec=zvec, galaxy_pa=galaxy_pa, offset=[20.,0.,3.], pot=potential.TriaxialNFWPotential(amp=1.,b=0.7,c=0.5))
     NFW_rot= potential.TriaxialNFWPotential(amp=1., zvec=zvec, pa=galaxy_pa,b=0.7,c=0.5)
     assert (evaluatePotentials(NFW_wrapped, 0., 0., phi=0.)-evaluatePotentials(NFW_rot, 20., - 3., phi=numpy.pi)) < 1e-6, 'Wrapped + Offset and Internally rotated NFW potentials do not match when evaluated at the same point'
+    #test a quick orbit integration to hit the C code (also test pure python)
```
jobovy, line 3649: Let’s  split this off into a separate test

jmackereth: ok, done - the problem seems to be that the offset is not being applied at all in the C code... I can't quite see why though? If I integrate the same orbit in a regular and offset potential, I get the same result when using the C integrators, but not when using the pure python (note that the test passes for pure python).

I initially thought this was a problem to do with the caching, but after doing some tests I can't see that that is the issue. It's probably blindingly obvious 😬

jmackereth: I think the currently implemented C code is doing things a little wrongly, since it applies the offset _only_ to the cached values. Just pushing up my current version, which fixes that...

jmackereth: here's a quick gist showing the problem (still persisting in the code I just uploaded): https://gist.github.com/jmackereth/94dd533eb91c1ddc24931489573a2d69

jobovy: Thanks for the notebook! I just ran this and it works for me, tests also pass... Are you sure you re-compiled?

galpy/potential/potential_c_ext/RotateAndTiltWrapperPotential.c
```diff
@@ -22,11 +23,23 @@ static inline void rotate_force(double *Fx, double *Fy, double *Fz,
...
+static inline void apply_offset(double *x, double *y, double *z, bool offsetSet, double *offset){
+  double xp,yp,zp;
+   xp = *x + *(offset);
+   yp = *y + *(offset+1);
+   zp = *z + *(offset+2);
+   *x = xp;
+   *y = yp;
+   *z = zp;
+  
+}
```
jobovy, line 35:
```suggestion
```

galpy/potential/potential_c_ext/RotateAndTiltWrapperPotential.c
```diff
@@ -36,6 +49,7 @@ void RotateAndTiltWrapperPotentialxyzforces(double R, double z, double phi,
     *(args + 3)= z;
     //now get the forces in R, phi, z in the aligned frame
     rotate(&x,&y,&z,rot);
+    apply_offset(&x,&y,&z,offsetSet,offset);
```
jobovy, line 52: I think we can just inline this here, in another branch I've also remove the `rotate` functions from this file:

```suggestion
    if ( offsetSet ) {
        x+= *offset;
        y+= *(offset+1);
        z+= *(offset+2);
    }
```

jobovy (changes requested):
Think it all looks good now, just some small suggestions in the C code.

For efficiency, maybe we should also track whether the rotation is set and only apply it if it is (so that if you just use an offset, the coordinates aren't constantly rotated). But we can do that in a follow-up PR as well...

jobovy:
Thanks, all good now! I just rebased to remove the commits with the original `OffsetWrapperPotential`, but once the tests pass I will merge. Thanks again!

PR #485, opened 2022-06-06
Updated named objects to includes GCs and errors

anyadovgal:
As I mentioned in my slack message, I included the GCs from https://github.com/GalacticDynamics-Oxford/GaiaTools/blob/master/input.txt , as well as updated the satellite galaxies to include errors where available from Dr. Pace's paper.


jobovy:
All looks good, great work again @anyadovgal, thanks!

PR #531, opened 2022-12-09
OrbitDistribution class

nstarman:
Signed-off-by: nstarman <nstarkman@protonmail.com>

TODO:

- [x] work if astropy not installed
- other convenience methods
- from_fit
- make sure properties like shape, and methods like reshape work as expected.
- work with orbits (plural): Probably want to interpret so that the last axis of the orbit shape is the sampling axis.
- shared base class Protocol for orbit-like things?

### DIFF
--- a/galpy/orbit/OrbitDistribution.py
+++ b/galpy/orbit/OrbitDistribution.py
@@ -1,0 +1,86 @@
+from dataclasses import dataclass
+
+import astropy.units as u
+import numpy as np
+from astropy.uncertainty import Distribution
+
+from galpy.orbit.Orbits import Orbit
+
+
+@dataclass(frozen=True)
+class OrbitDistribution:
+    orbit: Orbit
+
+    def __getattr__(self, name):
+        out = getattr(self.orbit, name)
+
+        if isinstance(out, (np.ndarray, u.Quantity)):
+            return Distribution(out)
+
+        elif callable(out):
+
+            def wrapped(*args, **kwargs):
+                _out = out(*args, **kwargs)
+                if isinstance(_out, (np.ndarray, u.Quantity)):
+                    return Distribution(_out)
+                return _out
+            return wrapped
+        return out
+
+    @classmethod
+    def from_orbit(cls, orbit):
+        return cls(orbit)
+
+    @classmethod
+    def from_obs_covariance(cls, orbit, covariance, *, units=(u.deg, u.deg, u.kpc, u.mas/u.yr, u.mas/u.yr, u.km/u.s), n_samples=10):
+        """From observational covariance matrix.
+
+        Parameters
+        ----------
+        orbit : Orbit
+            The orbit instance.
+        covariance : ndarray
+            Columns / rows are [ra, dec, distance, pm_ra_cosdec, pm_dec, vr]
+        units : tuple, optional
+            The units of the covariance matrix, by default (u.deg, u.deg, u.kpc,
+            u.mas/u.yr, u.mas/u.yr, u.km/u.s)s
+        n_samples : int, optional
+            _description_, by default 10
+
+        Returns
+        -------
+        _type_
+            _description_
+
+        Raises
+        ------
+        ValueError
+            _description_
+        """
+        if orbit.shape != ():
+            raise ValueError("Orbit must be a single orbit!")
+        # TODO! orbit cannot be integrated
+
+        from scipy.stats import multivariate_normal
+
+        mean = np.c_[
+            tuple(
+                getattr(orbit, n)(quantity=True).to_value(unit)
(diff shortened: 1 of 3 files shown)


jobovy:
Thanks, this looks great! I think this is an interesting enough addition that we can just try to merge a basic version and worry about improvements in follow-up PRs (we could open issues for any that we think _need_ to happen). But some design decisions have to be made now because they will be difficult to change later. Here are some questions/comments:
* What do we imagine supporting eventually? Initializing from 
  - [ ] Given samples
  - [ ] a mean + covariance in ICRS coordinates
  - [ ] a mean + covariance in Galactocentric coordinates
  - [ ] `from_fit`? --> follow-up PR, remove here
  - [ ] `from_name` using the listed uncertainties in the JSON file --> follow-up PR?
* Can't say I'm sold on having classmethods for all initialization, that is also different from how `Orbit` works. I think it would be best to mimic `Orbit` as much as possible.
* Would be good to have a member that is the 'best-fit' Orbit, so it could be, for example, plotted surrounded samples
* Do we eventually want to support multiple objects in the same `OrbitDistribution`? I think for now we should just do a single one.

> Looks like the slots option is py3.10+, so I'll wrap that in a version check.

I don't really see the advantage of using slots, so given that dropping 3.9 is years away, I would avoid it.

> @jobovy, are you interested in a Protocol class that defines what is an Orbit duck-type? I haven't yet made one for OrbitDistribution, instead relying just on `__getattr__` for more generic overloading.

I don't think this is necessary, because I can't imagine that there will be many other children of this parent class. 

I also noticed that you included a lot of typing code. While I understand that it doesn't interfere with the running of the code, it's a bit weird that have one file typed when nothing else is. So perhaps you could leave this out.

To be merged, this PR will also need:
- [ ] A section added to the ``orbits.rst`` documentation file illustrating its use
- [ ] API docs
- [ ] Tests, probably in a new file `test_orbitdistribution.py` in `tests/`. Maybe add to the CI job that also does `test_quantity.py`, since this will require astropy.

nstarman:
It now works given samples, and ICRS or galactocentric covariances. The latter keeps the "primary" orbit as the first sample.
It works with non-scalar orbits s.t. the last axis is the sampling axis (so an orbit of shape (N, M, ..., S) will have an (N, M, ...) primary orbit and S sample orbits of the same shape). So now we could do, e.g. all stars in a GC in one go.

For non-scalar orbits the covariance matrix can be a float (interpreted as a scaled identity matrix) a single matrix (made into a block-diagonal and applied to every orbit in the (N, M, ...)) or a block-diagonal matrix of shape (NxMx...x6, NxMx...x6).

The plotting is much better, with the samples plotted behind the primary as gray, with lower alpha.

Instantiation is much more similar to Orbit, accepting the same arguments. I've kept the class-method constructors so as not to complicate the instantiation.

I've renamed some stuff to be closer in line to the Astropy class and added properties like ``n_samples``.

Here's a nice example

```python
import galpy
import numpy as np
from galpy.orbit import Orbit
import astropy.units as u
from astropy.coordinates import SkyCoord
from galpy.orbit import OrbitDistribution
from galpy.potential import MWPotential2014
import matplotlib.pyplot as plt

np.random.seed(1)

nrand= (2, 2, 40)
ras= np.random.uniform(size=nrand)*360.*u.deg
decs= 90.*(2.*np.random.uniform(size=nrand)-1.)*u.deg
dists= np.random.uniform(size=nrand)*10.*u.kpc
pmras= 20.*(2.*np.random.uniform(size=nrand)-1.)*20.*u.mas/u.yr
pmdecs= 20.*(2.*np.random.uniform(size=nrand)-1.)*20.*u.mas/u.yr
vloss= 200.*(2.*np.random.uniform(size=nrand)-1.)*u.km/u.s
# Without any custom coordinate-transformation parameters
co = SkyCoord(ra=ras,dec=decs,distance=dists,
                 pm_ra_cosdec=pmras,pm_dec=pmdecs,
                 radial_velocity=vloss,
                 frame='icrs')
orbits = Orbit(co)

od2 = OrbitDistribution.from_cov_icrs(orbits[:, :, 0], cov=0.1, n_samples=10)
od2.integrate(np.linspace(0, 0.1, 10), MWPotential2014)
od2.plot()
```
df313b9b-dd53-4b5b-8625-5e04eae82808

galpy/orbit/OrbitDistribution.py
```diff
@@ -0,0 +1,352 @@
...
+        OrbitDistribution
+            The distribution of orbits.
+
+        Raises
+        ------
+        ValueError
+            If the orbit is not an array of orbits.
+        """
+        if orbit.shape == ():
+            raise ValueError("Orbit must be an array of orbits.")
```
nstarman, line 170:
```suggestion
```

nstarman:
Oops. Not stale. I'll get back on this!

PR #532, opened 2022-12-11
Add animate3d and general improvement to animate

henrysky:
This pull request included the following changes:

- Updated the code to use plotly.js v2, the `plotly-latest.min.js` will stay pinned to 1.58.4 according to plotly
- Used webgl for animate(), it is much faster in my testing. Now I can animate ~20 orbits together without much slowing down.
- Improved information displayed on hover. Now the names of orbits (got from from_name, otherwise naming them object 0, object 1....), current coordinates and time are displayed on hover like the following image.
image
- Added animate3d to Orbit class, only one subplot is allowed with the option of displaying the plane of milkyway galaxy when d1, d2, d3 are x, y, z like the following images.
image
image


galpy/orbit/Orbits.py
```diff
@@ -5597,28 +5652,566 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+                json.dump(jsonDict,jfile)
+            json_code= f"""Plotly.d3.json('{json_filename}',function(data){{"""
+            close_json_code= "});"
+        self.divid= 'galpy-'\
+            +''.join(choice(ascii_lowercase) for i in range(24))
+        button_width= 419.51+4.*10.
+        button_margin_left= int(numpy.round((width-button_width)/2.))
+        if button_margin_left < 0: button_margin_left= 0
+        # Configuration options
+        config= """{{staticPlot: {staticPlot}}}"""\
```
jobovy, line 5893: Does `staticPlot` actually work for 3D? Doesn't seem to work for me in VS Code.

galpy/orbit/Orbits.py
```diff
@@ -5597,28 +5652,566 @@ def animate(self,*args,**kwargs): #pragma: no cover
...
+        shape: 'spline',
+        width: 8.,
+        color: '{line_color}',
+        }},
+        type: "scatter3d",
+    }};
+    """.format(line_color=line_colors[0], name=names[0], xlabel=xlabels[0], ylabel=ylabels[0], zlabel=zlabels[0])
+        traces_cumul= """trace1,trace2"""
+        # milkyway plane surface
+        mw_bg_surface = f"""let mw_bg = {{
```
jobovy, line 5970: Woah! I think there's an assumption here that the scale is in kpc? But when the x/y/z are in internal units, it should be x `1/_ro`. The most robust way to check might be to do
```
'kpc' in labeldict.get(d1,r'\mathrm{No\ xlabel\ specified}'))
```
because you can't directly use `xlabel` in case it's been overwritten. Also see my comment on perhaps making the black part of the image transparent.

jobovy (changes requested):
Thanks @henrysky, this looks great! I think it's largely okay, but I left a bunch of small comments. Other things to do:
- [ ] Add an API doc at `doc/source/reference/orbitanimate3d.rst` and link to it in `doc/source/reference/orbit.rst`. You can just edit the `doc/source/reference/orbitanimate.rst` one.
- [ ] Add an entry to HISTORY.txt saying that you added `animate3d` (include the #532 of the PR). Maybe also add an entry saying that you edited `animate` and sped it up using webgl.
We might also want to add an example [here](https://docs.galpy.org/en/latest/orbit.html#animating-the-orbit), but I can do that later, it's a bit tricky embedding the animations in the docs.

henrysky:
Hi Jo. I should have fixed all of your suggestion with addition of the following changes

- Play/Pause/Resume/Finished! now share the same button (e.g. play will change to pause when pressed, pause will change to resume when pressed, the button change to finished! when the animation is done)
- When the animation is finished, plotting the final position of the orbit as a marker
- Much more natural looking MW galaxy, now it has a solid bar, semi-transparent spiral arms and disc, otherwise almost transparent like the image below for the solar orbit. The MW galaxy image need to keep low resolution (currently 50x50) otherwise the animation will be slowed down quite a bit
image
- Updated existing docs animation and added a 3d animation example

galpy/orbit/Orbits.py
henrysky: I have removed `staticPlot` for animate3d. Seems like a plotly.js issue (https://github.com/plotly/plotly.js/issues/457)

jobovy:
Thanks @henrysky, this all looks great! Not sure why the RTD check didn't work, I can build the docs locally and they seem fine. So that's probably not a problem. I have two more minor comments:
* I'm not a big fan of showing the final point as a big fat marker at the end of the orbit integration. I don't think I want this for most of the animations that I use (e.g., ones like [this](https://galaxiesbook.org/chapters/III-02.-Orbits-in-Triaxial-Mass-Distributions.html#galpy-fumbxltnzjbxuhbxguuxauei)). I would suggest entirely dropping this, but if you think it can be useful sometimes, maybe it could be optional and by default off?
* I like having just the single play/pause/resume/finished button, but it's not ideal that the button now changes size based on what it says and the other buttons move around. If it's not too hard to fix, could you set its width such that it is always the same? I'm not sure how to best do this, maybe there is some way of calculation what the width of the button is when it says Finished! and then fix the width to that? If this is too hard, don't worry about it and I can just try to fix this at some later time.

jobovy:
> On the side note, I just noticed if using the web python thing on galpy doc homepage. The first time calling `o.animate()` just return blank, only works if calling `o.animate()` the second time

Oh that's interesting! Yes, I also noticed that it doesn't work in the browser version and, similarly, noticed that `animate` also doesn't work in jupyterlab (or in retrolab). It seems like the issue is that the `require` library is somehow not available (check the console errors). But I didn't know that it works if you just call `animate` again! That also fixes it in retrolab (and I assume jupyterlab). It must be that one should really 'await' `require` being available? Anyway, I think this is something I can fix later. It might be that the better way to fix this is to not use `require` at all, but I'm not sure how to replace it.

jobovy:
I think I figured out how to do it by waiting until `require` is defined. I can just apply this later:

<details>

```
From 633640cdb0a9a1d149bb6bb81b2862075a77e984 Mon Sep 17 00:00:00 2001
From: Jo Bovy <bovy@astro.utoronto.ca>
Date: Tue, 13 Dec 2022 23:29:17 -0500
Subject: [PATCH] Wait until require is loaded before using it

---
 galpy/orbit/Orbits.py | 9 ++++++---
 1 file changed, 6 insertions(+), 3 deletions(-)

diff --git a/galpy/orbit/Orbits.py b/galpy/orbit/Orbits.py
index c5837f33..8cc28b79 100644
--- a/galpy/orbit/Orbits.py
+++ b/galpy/orbit/Orbits.py
@@ -5529,14 +5529,15 @@ class Orbit:
 {load_jslibs_code}
 
 <script>
+var require_interval;
+require_interval = setInterval(function() {{
+  if (typeof require !== 'undefined') {{
+      clearInterval(require_interval);
 require.config({{
   paths: {{
     Plotly: 'https://cdn.plot.ly/plotly-latest.min',
   }}
 }});
-</script>
-
-<script>
 require(['Plotly'], function (Plotly) {{
 {json_code}
   let layout = {layout};
@@ -5610,6 +5611,8 @@ require(['Plotly'], function (Plotly) {{
     }}, 30);
     }}
 {close_json_code}}});
+}}
+}}, 30);
 </script>""".format(json_code=json_code,close_json_code=close_json_code,
                     divid=self.divid,width=width,height=height,
                     button_margin_left=button_margin_left,config=config,
-- 
2.37.1 (Apple Git-137.1)
```

</details>

jobovy (approved):

jobovy:
> I think I figured out how to do it by waiting until `require` is defined. I can just apply this later:

See #534 for a fix that actually works (what I suggested didn't seem to work well in all contexts).

henrysky:
I just noticed I made a typo in the galpy doc. the animate3d example code is `op.animate(mw_plane_bg=True)` but should be `op.animate3d(mw_plane_bg=True)`.

jobovy:
> I just noticed I made a typo in the galpy doc. the animate3d example code is `op.animate(mw_plane_bg=True)` but should be `op.animate3d(mw_plane_bg=True)`.

Okay, I've fixed this in 4bf1fe3f5dcb16262c7861eda3e325d3a8c310db.

Issue #506, opened 2022-09-14
``estimateDeltaStaeckel`` at z=0


jobovy:
## 🐛 Bug

``estimateDeltaStaeckel`` returns NaN when evaluated at z=0. This is a real, physical issue, not a bug per se, because for a location in the mid-plane, one cannot know anything about the vertical extent of the potential and, thus, cannot determine whether or not the potential is of Staeckel form and what its focal length would be.

Because this is a real effect, it's not entirely clear how to fix this. Options include

- Evaluated at a small, positive z instead. However, it's unclear what this positive z should be. Some investigation of the gradient of the estimated Delta as a function of z for, e.g., ``MWPotential2014`` or ``McMillan17`` would be useful in deciding this
- Raise an error or a warning (error is annoying when using this to estimate actions/... for an Orbit)

### Reproducible example

```
from galpy.orbit import Orbit
from astropy import units
from galpy.potential.mwpotentials import McMillan17
from galpy.actionAngle import estimateDeltaStaeckel 

o = Orbit(vxvv=[8.462* units.kpc, 0.6 * units.km / units.s, 236.1* units.km / units.s, 0.0 * units.kpc, -7.5 * units.km / units.s],ro=8.21,vo=233.1)
estimateDeltaStaeckel(McMillan17,o.R(),o.z())
```

(from https://github.com/jobovy/galpy/issues/495#issuecomment-1245001187)

### Expected behavior

A non-NaN value.

### Additional context

Originally reported here: https://github.com/jobovy/galpy/issues/495#issuecomment-1245001187

jamesmlane:
FWIW I have run into this issue and have found that for many potentials `delta` monotonically approaches a constant as z->0 and therefore using some small value of z is satisfactory when z=0. Can provide code to show this.

jamesmlane:
Here's a figure showing that Staeckel delta approaches a constant value within |z|~0.01 for a wide range of radii in MWPotential2014.

![staeckel_delta_at_z0_mwpotential2014](https://user-images.githubusercontent.com/20076217/219717600-04cebb61-5150-4b2d-b61a-36d6bcb42463.png)

The same is approximately true for McMillan17 potential. For some combinations of (R,z) the Staeckel delta doesn't work, but when it does it does a reasonable job of approaching a constant value.

![staeckel_delta_at_z0_mcmillan17](https://user-images.githubusercontent.com/20076217/219718875-54f97650-af14-4a0f-8a39-f58e3d5b7239.png)

PR #551, opened 2023-02-17
Fix for issue #506

jamesmlane:
Adds a check for z=0 in estimateDeltaStaeckel and sets to 1e-4. This figure shows that delta approaches a constant value for z->0 and setting z=1e-4 returns an appropriate value (dashed lines)

staeckel_delta_z0_fix


jobovy:
This looks great! Can we add a test that running `estimateDeltaStaeckel` works at z=0 for scalars and arrays?

PR #577, opened 2023-05-30
minor astropy units + mass calc fixes

jamesmlane:
- Passing `rmin` as astropy quantity to `sphericaldf.sample()` always fails the comparison with `self._rmin_sampling`, so parse to internal units first.
- Uncomment the try-except sequence in `_make_cmf_interpolator()` to allow galpy potentials that don't have vectorized mass calculations to work properly.
- Add `use_physical=False` to the mass normalization for `self._rmin_sampling > 0` so it works properly when denspot has ro,vo set.

### DIFF
--- a/galpy/df/sphericaldf.py
+++ b/galpy/df/sphericaldf.py
@@ -477,16 +477,11 @@ class sphericaldf(df):
         ximax = _RToxi(self._rmax, a=self._scale)
         xis = numpy.arange(ximin, ximax, 1e-4)
         rs = _xiToR(xis, a=self._scale)
-        # try/except necessary when mass doesn't take arrays, also need to
-        # switch to a more general mass method at some point...
-        try:
-            ms = mass(self._denspot, rs, use_physical=False)
-        except (ValueError, TypeError):
-            ms = numpy.array([mass(self._denspot, r, use_physical=False) for r in rs])
+        ms = mass(self._denspot, rs, use_physical=False)
         mnorm = mass(self._denspot, self._rmax, use_physical=False)
         if self._rmin_sampling > 0:
             ms -= mass(self._denspot, self._rmin_sampling, use_physical=False)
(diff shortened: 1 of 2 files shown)


jobovy:
Thanks! Looks like there's a typo in the code that needs to be fixed. Otherwise, two more comments:
* The `self._rmin_sampling = conversion.parse_length(rmin, ro=self._ro)` line should now just be `self._rmin_sampling = rmin`
* We'l probably need a test involving a potential that doesn't allow vectorized mass calculations. I think that code was commented out, because there was no potential that doesn't allow vectorized mass calculations that we were using, but it sounds like you must have hit upon one? So perhaps just a quick sampling test with that potential to verify that new code.

jamesmlane:
Let me double check but I'm positive I uncommented the try-catch statement to fix something (one of the power law potentials I think) that was failing because it was passing into the numerical integration in `Potential.mass()`

jamesmlane:
Okay I confirmed that vectorized mass calculations fails for `PowerSphericalPotential`. Maybe the right approach here is to just address this in `Potential.mass()`. Adding list comprehension there should solve this issue while also preventing the issue arising for casual users of `PowerSphericalPotential` when calculating masses?

```
from galpy import potential
import numpy

pot = potential.PowerSphericalPotential()
rs = numpy.linspace(1,10,num=10)
mass = pot.mass(rs) # Fails -> TypeError during Gauss thm. integration
```

jobovy:
I think we should just use the ``try/except`` clause in ``sample`` and add a quick sampling test using the `PowerSphericalPotential` potential. There is currently no guarantee that potential functions like `mass` (or `evalautePotential`, `evaluateRforces`, etc.) work with arrays and properly making this work is a bit of work (really should then be for all potential functions, maybe properly deal with multi-dimensional arrays [so difficult w/ list comprehesions]). So I think the best way forward is to just use the ``try/except`` clause and just make sure there's a test that hits it.

PR #629, opened 2024-01-29
Fix a bug in streamspraydf to remove the streaks.

jl3937:
Fix a bug in streamspraydf to remove the streaks.

Before this fix: 
download (1)
After the fix:
download

### DIFF
--- a/.github/workflows/build_pyodide_wheels.yml
+++ b/.github/workflows/build_pyodide_wheels.yml
@@ -39,7 +39,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.10.2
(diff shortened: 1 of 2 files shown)


jobovy:
Thanks!  Could you write this without the `for` loop?

jobovy:
Thanks again for finding this issue! I implemented a much simpler fix that appears to do the same, but if you have time to check this, that would be great.

jl3937:
> Thanks again for finding this issue! I implemented a much simpler fix that appears to do the same, but if you have time to check this, that would be great.

Oh this works perfectly! Thanks for the clean fix!

PR #641, opened 2024-04-29
update docs in triaxialNFW

nstarman:
I think there's a small error in the bounds of integration in the docstring.


jobovy:
Thanks!

PR #643, opened 2024-05-12
Ias15 implentation

johnwez1:
Implemented the ias15 algorithm in c with python wrappers, and added to relevant tests.
Some notes:
- Didnt implement the “begins to oscillate”
- Used "global" calculations.
- Didn't implement user warning for 12 iterations
- My implementation treats each step as an independant integration, with dynamic stepping in between. This part seems a bit silly but I thought at least I would give a working version at first. Otherwise, I would need to create an interpolation function to match the other integrators.
- Slight concern that it might get caught in infinite loop, in the while time remaining > 0 loop.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -2,7 +2,7 @@ v1.10.2 (expected around 2025-03-01)
 ====================
 
  - Implemented the IAS15 integrator of Rein & Spiegel 2015.
- 
+
  - Combine the drift calculations in the Python leapfrog integrator (#365).
 
  - Move the checks for non-axisymmetric and dissipative potentials from internal to
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,6 +1,8 @@
 v1.10.2 (expected around 2025-03-01)
 ====================
 
+ - Implemented the IAS15 integrator of Rein & Spiegel 2015.
+ 
  - Combine the drift calculations in the Python leapfrog integrator (#365).
 
  - Move the checks for non-axisymmetric and dissipative potentials from internal to
--- a/doc/source/orbit.rst
+++ b/doc/source/orbit.rst
@@ -1014,9 +1014,11 @@ the ``orbit.integrate`` method. Currently available integrators are
 * rk6_c
 * dopr54_c
 * dop853_c
+* ias15_c
 
-which are Runge-Kutta and Dormand-Prince methods. There are also a
-number of symplectic integrators available
+which are Runge-Kutta, Dormand-Prince methods and the IAS15 
+integrator in `Rein & Spiegel (2014) <https://doi.org/10.1093/mnras/stu2164>`_. 
+There are also a number of symplectic integrators available
 
 * leapfrog_c
 * symplec4_c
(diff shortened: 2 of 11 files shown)


jobovy:
Also, to avoid the ``pre-commit`` commits here (which you should pull in before making any further changes), install ``pre-commit`` locally:
```
pip install pre-commit
pre-commit install
```
This will run the pre-commit hooks whenever you commit locally, automatically fix issues (like indentation and code style), and tell you when you need to fix something manually.

jobovy:
Thanks for fixing the error message bug! Looks like there's some flakiness in the integrator, because the energy conservation tests pass for some Python/OS versions but fail for others (at least that's what it seems like from the test runs). It seems like the integrator sometimes just returns NaNs. Also the backwards integration tests seem to always be failing, maybe the integrators doesn't immediately work for backwards integration?

johnwez1:
I made the PR after every test in test_orbit was passing on local, so yeah I guess something dependant on OS/Python version. I've recreated some issues with a fresh python 3.12 so hopefully that will lead to a fix. Anyway, leave it with me! I don't have a whole lot of free time at the moment to work on it but hopefully it shouldn't take too long.

jobovy:
Thanks for the helpful comments @hannorein!

I'll pin this PR so it doesn't get automatically closed, in case @johnwez1 or someone else wants to finish it soon.

jobovy:
Looks like the orbit integration tests are still hanging, so I think I'll cancel them again to not waste compute. Do they finish locally for you in a short time? I assume it's that they're going into an infinite loop, is this possible in the code?

jobovy:
Hi @johnwez1,

Thanks for these latest changes and apologies for my slow response, things have been very busy at the end of the term. Looks like all the tests are passing now 🎉 ! I also briefly tested the code locally and things seem to work. On a simple test problem, the `ias15_c` integrator is about 6 times slower than `dop853_c`, but it is more precise by a few orders of magnitudes.

There are just a few things left to do to fully finish this PR:
- [ ] There are five new lines that aren't covered by the tests, all in `[galpy/util/wez_ias15.c](https://app.codecov.io/gh/jobovy/galpy/pull/643/blob/galpy/util/wez_ias15.c)`. Three lines are related to being able to interrupt the code; to cover these you should add `ias15_c` to the integrators in the `test_orbit_c_sigint_full` test in test_orbit.py and in the other `...sigint...` tests as well in that file. The other two lines that are not covered are in the `double seventhroot` function, specifically the lines in the `while(a<1e-7 && isnormal(a)){` loop. I don't know whether you know a test case that would hit those lines, that would be ideal to add. If not, we should probably just be okay with slightly increasing the number of uncovered lines.
- [ ] Could you add the `ias15_c` integrator to the docstrings of the relevant methods in `Orbits.py`? You can just search for all mentions of `dop853_c` and add to those docstrings
- [ ] It would also be good to add to the documentation under `doc/source/orbit.rst` to mention the new integrator. The most minimal addition is to mention it in the "Fast orbit integration and available integrators" section. You might want to add a note on when using `ias15_c` would be appropriate; I imagine it's mostly for cases where very high precision is required.
- [ ] Please add an entry to the HISTORY.txt file.

I think that should be it and hopefully this will be pretty straightforward. Thanks for your contributions so far!

jobovy:
Thanks for the updates! I rebased your branch to the latest `main`, resolving some issues, and force pushed the result, so if you make any other changes, make sure to **first ``git pull`` to get the latest version of your branch**.

There was a test failure that led me to realize that I think IAS15 in practice acts as a symplectic integrator in galpy, that is, it can only be used where symplectic integrators can be used (because you use the force in a = F rather than the derivative as in d(x,v) /dt = D). So I removed it from all functions where you need the latter integrator (``integrate_dxdv`` and the surface-of-section integrations).

I noticed that there is still a TODO listed in the C code, is this something that needs to be done?

johnwez1:
Hi Jo, the TODO was for a warning to be surfaced in the case of infinite loops potentially occurring in the predictor–corrector loop when using a static time step. In our implementation, we only use dynamic time stepping and so I've removed the TODO.

jobovy:
Great, thanks! I think this all looks good now, so I'll merge it, thanks again for your contribution!

PR #652, opened 2024-06-20
Test wheels in a conda environment

jobovy:


jobovy:
Failure of wheels in a `conda` environment appears to be due to `libopenblas` version > 0.3.23 causing an OpenMP runtime issue. So I have pinned `libopenblas==0.3.23` for now and opened up #653 to track that issue.

PR #660, opened 2024-07-02
Add OpenMP support to Mac builds and wheels

jobovy:


jobovy:
Difficult to avoid the issue with combining conda-``numpy`` and pip-``galpy``, so approach taken here is to detect and warn about the issue, falling back on non-C version of ``galpy``. 

Supersedes and thus closes #651.

PR #670, opened 2024-08-15
Implement the new particle spray model by Chen+24

ybillchen:
Implement the new particle spray model by Chen+24 and include progenitor's potential for stream integration


jobovy:
Thanks for starting the PR. You might want to install and use precommit as discussed in the development installation docs that I linked to in my email. That would avoid all the tweaks by precommit.

ybillchen:
> Thanks for starting the PR. You might want to install and use precommit as discussed in the development installation docs that I linked to in my email. That would avoid all the tweaks by precommit.

Thank you! This is very useful.

ybillchen:
Hi @jobovy,

This PR have passed most checks! I also added some new tests, documentation, and updated `HISTORY.txt`.

Let me know if more is needed!

-Bill

jobovy:
> Hi @jobovy,
> 
> This PR have passed most checks! I also added some new tests, documentation, and updated `HISTORY.txt`.
> 
> Let me know if more is needed!
> 
> -Bill

Thanks! I’ll review this next week.

galpy/df/streamspraydf.py
```diff
@@ -355,3 +296,366 @@ def _setup_rot(self, dt):
...
+        theta = posvel[:, 2] * numpy.pi / 180
+        alpha = posvel[:, 4] * numpy.pi / 180
+        beta = posvel[:, 5] * numpy.pi / 180
+
+        xst = xyzpt[:, 0] + Dr * numpy.cos(theta) * numpy.cos(phi)
+        yst = xyzpt[:, 1] + Dr * numpy.cos(theta) * numpy.sin(phi)
+        zst = xyzpt[:, 2] + Dr * numpy.sin(theta)
+        vxst = vxyzpt[:, 0] + Dv * numpy.cos(beta) * numpy.cos(alpha)
+        vyst = vxyzpt[:, 1] + Dv * numpy.cos(beta) * numpy.sin(alpha)
+        vzst = vxyzpt[:, 2] + Dv * numpy.sin(beta)
```
jobovy, line 477: Use ``util.coords.spher_to_cyl(_vec)`` and ``util.coords.cyl_to_rect(_vec)`` here.

galpy/df/streamspraydf.py
```diff
@@ -355,3 +296,366 @@ def _setup_rot(self, dt):
...
+
+
+class streamspraydf(fardal15spraydf):
+    def __init__(self, args, **kwargs):
+        """
+        For backward compatibility
+        """
+        super().__init__(args, **kwargs)
+        warnings.warn(
+            "Class `streamspraydf` will be deprecated in a later version. "
```
jobovy, line 656:
```suggestion
            "Class `streamspraydf` will be deprecated in version 1.11. "
```

galpy/df/streamspraydf.py
```diff
@@ -355,3 +296,366 @@ def _setup_rot(self, dt):
...
+            "Please use class `fardal15spraydf` for the Fardal+15 particle spray model.",
+            FutureWarning,
```
jobovy, line 658:
```suggestion
            DeprecationWarning,
```

galpy/df/streamspraydf.py
```diff
@@ -62,8 +58,9 @@ def __init__(
         -----
         - 2018-07-31 - Written - Bovy (UofT)
         - 2021-05-05 - Added center keyword - Yansong Qian (UofT)
+        - 2024-08-11 - Generalized for more df - Yingtian Chen (UMich)
```
jobovy, line 61:
```suggestion
        - 2024-08-11 - Generalized to allow different particle-spray methods - Yingtian Chen (UMich)
```

galpy/df/streamspraydf.py
```diff
@@ -124,6 +120,8 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
             If True, also return the time since the star was stripped. Default is False.
         integrate : bool, optional
             If True, integrate the orbits to the present time. If False, return positions at stripping (probably want to combine with returndt=True then to make sense of them!). Default is True.
+        pot_prog : galpy.potential.Potential or list of such instances or None, optional
+            Potential for the progenitor. Ignored if None.
```
jobovy, line 124: Since the potential of the progenitor and the ```MovingOrbitPotential`` that it gives rise to doesn't change, the progenitor potential and its ``MovingOrbitPotential`` counterpart should just initialized when the object is instantiated. And then you can just add it to ``_pot`` I think (but not ``_rtpot`` or ``_centerpot``) so it gets automatically used in the sampling routine.

jobovy (changes requested):
@ybillchen --I did a first review mainly focusing on the code in ``streamspraydf`` and the tests, but not yet the documentation. Overall it looks great and great job on the tests! I have a bunch of comments that should be addressed before this can be merged. I might have more comments later, these are just some major ones to get started. Thanks!

galpy/df/streamspraydf.py
```diff
@@ -355,3 +296,366 @@ def _setup_rot(self, dt):
...
+                        use_physical=False,
+                    )
+                    for ii in range(len(Rpt))
+                ]
+            )
+
+        # Sample positions and velocities in the instantaneous frame
+        posvel = numpy.random.multivariate_normal(self._mean, self._cov, size=len(dt))
+        Dr = posvel[:, 0] * rtides
+        v_esc = numpy.sqrt(2 * self._progenitor_mass / Dr)
```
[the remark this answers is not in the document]
ybillchen: Yes, I used vesc = sqrt(2GM/r), although I'd expect self._pot_prog.vesc has no significant difference. But I suggest use vesc = sqrt(2GM/r) to match the published version.

galpy/df/__init__.py
```diff
@@ -66,4 +66,7 @@
 osipkovmerrittdf = osipkovmerrittdf.osipkovmerrittdf
 osipkovmerrittNFWdf = osipkovmerrittNFWdf.osipkovmerrittNFWdf
 constantbetadf = constantbetadf.constantbetadf
+basestreamspraydf = streamspraydf.basestreamspraydf
```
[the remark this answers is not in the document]
ybillchen: This is a bit tricky. I did this because I want to test the `basestreamspraydf.spray_df()` method. Otherwise, that part of codes would be reported as uncovered by tests.

Normally, I would decorate the top-level class as an abstract class (and make the `spray_df()` method an abstract method) so that I don't need to bother the test coverage. But this is not how `galpy` works. Do have any suggestions?

galpy/df/streamspraydf.py
ybillchen: I think moving `pot_prog` to `__init__()` is a good idea. But I would not combine it with `self._pot`. Otherwise, the user would have to integrate the progenitor's orbit first, and then create a `MovingOrbitPotential` object, and finally pass it to the `__init__()` function. This is too much extra work. I still keep it as a non-moving potential, and create `MovingOrbitPotential` object when initialized.

galpy/df/streamspraydf.py
```diff
@@ -355,3 +296,366 @@ def _setup_rot(self, dt):
...
+        -------
+        xst, yst, zst : array, shape (N,)
+            Positions of points on the stream in the progenitor coordinates.
+        vxst, vyst, vzst : array, shape (N,)
+            Velocities of points on the stream in the progenitor coordinates.
+        """
+        Rpt, phipt, Zpt = coords.rect_to_cyl(xyzpt[:, 0], xyzpt[:, 1], xyzpt[:, 2])
+        vRpt, vTpt, vZpt = coords.rect_to_cyl_vec(
+            vxyzpt[:, 0], vxyzpt[:, 1], vxyzpt[:, 2], Rpt, phipt, Zpt, cyl=True
+        )
```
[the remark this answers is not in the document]
jobovy: I see what you mean, I hadn't noticed that you don't use the velocities. Yeah, so then it's probably best to keep it this way.

galpy/df/__init__.py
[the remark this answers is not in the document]
ybillchen: Thanks for the suggestions! 

For 1: I have redirected `galpy.df.streamspraydf` to `fardal15spraydf` for backwards compatibility, so this won't work unless I change the module's name

3 is a good idea! I didn't know it would be ignored. But good to know!

galpy/df/streamspraydf.py
jobovy: > I think moving `pot_prog` to `__init__()` is a good idea. But I would not combine it with `self._pot`. Otherwise, the user would have to integrate the progenitor's orbit first, and then create a `MovingOrbitPotential` object, and finally pass it to the `__init__()` function. This is too much extra work. I still keep it as a non-moving potential, and create `MovingOrbitPotential` object when initialized.

I didn't mean to put it in `pot` (I considered that, but decided against it for the reasons that you mention), but just that once you have integrated the progenitor in `pot` in the `__init__` function and you have set up the `MovingObjectPotential`, you can just add the `MovingObjectPotential` to `self._pot`, because it's only used to integrate stream stars after the initialization. That way you avoid having to check whether the progenitor should be included in the `sample` method.



> I noticed that my implementation of `pot_prog` doesn't work correctly if `center` or `centerpot` is set. In this case, I should replace `MovingObjectPotential(orbit=self._progenitor, ...)` with the true orbit considering the center. Is there a easy way to combine `self._progenitor` and `self._center` into a new `Orbit` object?

I believe the progenitor's orbit integrated in `pot` is the true orbit even when `center` is set (see the `test_center` code). How does it not work?

galpy/df/streamspraydf.py
```diff
@@ -102,11 +105,18 @@ def __init__(
...
         else:
             self._center = None
+        if progpot is not None:
+            progtrajpot = MovingObjectPotential(
+                orbit=self._progenitor,
+                pot=progpot,
+                ro=self._ro,
+                vo=self._vo,
+            )
+            self._pot = flatten_potential([self._pot, progtrajpot])
```
jobovy, line 115: Given that ``self._pot`` is already flattened, this can just be
```suggestion
            self._pot = self._pot+progtrajpot
```

galpy/df/streamspraydf.py
```diff
@@ -297,6 +295,56 @@ def _setup_rot(self, dt):
...
+                            Zpt[ii],
+                            phi=phipt[ii],
+                            t=-dt[ii],
+                            use_physical=False,
+                        )
+                    )
+                    for ii in range(len(Rpt))
+                ]
+            )
+        return rtides, vcs
```
jobovy, line 346: Since in your method you only need the tidal radius, not the circular velocity, it would be somewhat more efficient to have two methods, one for the tidal radius and one for the circular velocity, so you can just use the tidal radius one.

jobovy (changes requested):
Few more small comments on the latest changes.

galpy/df/streamspraydf.py
ybillchen: > I believe the progenitor's orbit integrated in pot is the true orbit even when center is set (see the test_center code). How does it not work?

I misunderstood this. Then it should work!

galpy/df/__init__.py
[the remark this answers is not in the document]
ybillchen: Not sure why solution 3 is not working. I will use `# pragma: no cover`. But this is a small edit. I'll save this for the next commit to save test times.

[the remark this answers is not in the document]
jobovy: It's because you have to do `raise NotImplementedError` rather than raise a warning.

jobovy:
Sorry for the delay in finishing this review, I'm hoping to get to this later this week.

jobovy (approved):

jobovy:
Sorry for the delay, but I was finally able to do a final review of the documentation updates and after a few small tweaks was able to merge this. Thanks again for your contribution @ybillchen!

PR #698, opened 2024-12-12
Update potential.rst for changes in pynbody docs/data

apontzen:
This PR updates potential.rst to reflect changes made in pynbody (primarily the documentation and test data URLs)

It also changes the example so that it only makes use of public interface rather than protected members, and in doing so reduces the number of calls made to pynbody.


jobovy:
Thanks!

Issue #701, opened 2025-01-04
Performance improvements for MW orbit integration


jamesgrimmett:
Hi @jobovy, I have recently been looking into whether I could improve the runtime for integrating orbits when using the `MWPotential2014` potential. (I'm trying to model up to 1000 orbits on a small number of cpus, so any speedup is useful).

I've noted two things that could help to reduce runtime;

 (i) In the force calculations in `PowerSphericalPotentialwCutoff.c`, the interior mass is found using 
```
gsl_sf_gamma(x)  - gsl_sf_gamma_inc(x,a)
```
i.e., 

```math
\Gamma(x) - \Gamma(x,a) = \int^\infty_0t^{x-1}\mathrm{exp}(-t)dt - \int^\infty_a t^{x-1}\mathrm{exp}(-t)dt
```

I've found that a fair amount of the runtime is spent in `gsl_sf_gamma_inc`. A cheaper way to compute the lower integral is with 

```
gsl_sf_gamma(x) * gsl_sf_gamma_inc_P(x, a)
``` 
i.e., 
```math
\Gamma(x) \times P(x,a) = \Gamma(x) \times 1/\Gamma(x)\int^a_0 t^{x-1}\mathrm{exp}(-t)dt
```

 I don't completely understand why `gsl_sf_gamma_inc_P` is signficantly faster than `gsl_sf_gamma_inc`, to be honest. I think it is to do with GSL's internal optimisations for the lower integral vs. the upper unbounded integral. The speedup also depends on the value of $x, a$ (i.e., $alpha, R$). 
In practice (on 12 cpus) I find that this change reduces `Orbit().integrate()` runtime by 40%, and `Orbit.from_name('MW globular clusters').integrate()` runtime by 85% (over a 10Gyr integration period).

 (ii) As the Rforce and zforce computations tend to be very similar, there are some redundant calculations made during each call to `evalRectForce`. E.g., for a `PowerSphericalPotentialwCutoff` potential, the interior mass is calculated once for the Rforce and again for the zforce (see screenshot below showing a section of the call map, `gsl_sf_gamma_inc` is proxy for the mass calculation, percentages indicate the amount of time that each function was active during runtime (% relative to calling function total)).

<img width="673" alt="callmap" src="https://github.com/user-attachments/assets/763f5fb4-9a9d-4560-ade5-3d0d0ea49c33" />
 
[...]

PR #720, opened 2025-03-10
Implement caching of the mass for the PowerSphericalPotentialwCutoff potential

jamesgrimmett:
Hi @jobovy, here are the second set of changes to close off https://github.com/jobovy/galpy/issues/701.

I've taken two different approaches in two commits, happy to rebase this branch to contain whichever you prefer.

In https://github.com/jobovy/galpy/commit/020eb773eb2ad2b7b7e277cf55f331927cee1827 I've implemented the caching of the mass in the `Rforce` and `zforce` functions. I left the Planar functions as-is (including the nargs/potentialArgs) because I don't think there would be any performance benefit from caching in the Planar case, but just let me know if you'd prefer the caching to be included for the Planar functions as well for consistency (or if I'm mistaken about the lack of benefit).

In https://github.com/jobovy/galpy/commit/84da14b4bd885465f86f16bf8e988729650928ff I instead implemented the caching inside of the mass function itself. This requires that the potentialArgs be passed to the mass function, but reduces the duplicate code in each of the calling functions. In this case the caching needs to be set up for any situation that the mass function going to be called, i.e. including the Planar case.

I guess the tradeoff is that first option is potentially more flexible, whereas the second option is a bit tidier and reduces some repetition in the code.

In terms of performance both commits are the same; compared to the `main` branch I see a 25% performance improvement. E.g., using this simple test case;

```
    ts = np.linspace(0,1000,100001)*u.Gyr
    sun = Orbit()
    t = time.time()
    sun.integrate(ts, MWPotential2014, method="symplec4_c")
    print(f"Duration: {time.time() - t}")
```

With the current `main` branch I get
```
Duration: 7.169827222824097
```
and with this change I get
```
Duration: 5.4198899269104
```

Finally I haven't implemented any caching for the Python integrator, but am happy to include that here if you like (on that front, [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) is an interesting option).

### DIFF
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -138,6 +138,7 @@ def _parse_pot(pot, potforactions=False, potfortorus=False):
         elif isinstance(p, potential.PowerSphericalPotentialwCutoff):
             pot_type.append(15)
             pot_args.extend([p._amp, p.alpha, p.rc])
+            pot_args.extend([0.0, 0.0])  # for caching
         elif isinstance(p, potential.MN3ExponentialDiskPotential):
             # Three Miyamoto-Nagai disks
             npot += 2
--- a/galpy/orbit/orbit_c_ext/integrateFullOrbit.c
+++ b/galpy/orbit/orbit_c_ext/integrateFullOrbit.c
@@ -234,7 +234,7 @@ void parse_leapFuncArgs_Full(int npot,
       potentialArgs->ntfuncs= 0;
       potentialArgs->requiresVelocity= false;
       break;
-    case 15: //PowerSphericalwCutoffPotential, 3 arguments
+    case 15: //PowerSphericalwCutoffPotential, 5 arguments
       potentialArgs->potentialEval= &PowerSphericalPotentialwCutoffEval;
       potentialArgs->Rforce= &PowerSphericalPotentialwCutoffRforce;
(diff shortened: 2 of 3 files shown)


jamesgrimmett:
Noted the tests failing for `tests/test_orbits.py -k 'not test_energy_jacobi_conservation'`, I've converted back to draft and will take a look at the cause

jobovy:
Hi @jamesgrimmett , thanks for this PR and implementing the two options! While I think doing the caching in the mass function is more elegant, I think it might be best to go with the caching in the force functions themselves, because (a) it's more consistent with how caching is doing for other potentials, (b) it avoids unnecessary caching when evaluating the potential or using ``planarRforce``, and (c) you could then actually cache the entire ``- amp * m / pow(r2,1.5)`` part of the force, just needing to multiply by ``R`` or ``Z`` to get the ``R/Zforce``. It probably won't make a real difference for the performance, but if we're going to be caching, we might as well cache as much as possible. Thanks!

jamesgrimmett:
Thanks @jobovy, all good points! Agreed that is the more practical change, and consistency is good.
I've reverted back to the first option, and made the change to cache that entire force magnitude(?) term. As you guessed, still the same ~25% performance improvement, but there isn't any sense in not caching it and should save a few numbers from being crunched. Just let me know if there should be any other adjustments.

jobovy:
Looks great, thanks so much!

PR #731, opened 2025-05-30
Add Fuzzy Dark Matter (FDM) dynamical friction

Adrian998obs:
Fuzzy Dark Matter model modifies the dynamical friction compared to Cold Dark Matter. This class computes the dynamical friction in FDM universe.


galpy/potential/FDMDynamicalFrictionForce.py
```diff
@@ -0,0 +1,147 @@
+import astropy.units as u
+import numpy as np
+import scipy.special as sp
+from astropy.constants import c, hbar
+from scipy import interpolate
+from scipy.integrate import quad
+
+from galpy.potential import ChandrasekharDynamicalFrictionForce
```
jobovy, line 8: This should fix the test failures (and probably failures when you try to run this code)
```suggestion
from . ChandrasekharDynamicalFrictionForce import ChandrasekharDynamicalFrictionForce
```

jobovy:
Just a quick note that I rebased to pull in an update to the wheel-buiding workflow, so make sure to `git pull` before committing, because otherwise there will be trouble!

galpy/potential/FDMDynamicalFrictionForce.py
```diff
@@ -79,13 +63,10 @@ def FDMfactor(self, r, vs):
         if self._lnLambda:
             return self._lnLambda
         else:
-            kr = self.krValue(r, vs)
+            kr = 2 * self.krValue(r, vs)
+            I = -sp.sici(kr)[1] + np.log(kr) + 0.5772156649015329
```
jobovy:
```suggestion
            I = -sp.sici(kr)[1] + np.log(kr) + numpy.euler_gamma
```
Not a full review yet, but just noticed this...

Adrian998obs: yes, thank you !

PR #748, opened 2025-08-23
Symbolic potential

Yuzhe98:
This fork has been synced with the jobovy/galpy main before this pull request. 

This fork contains updates of:
1. Optional dependency of sympy (mentioned in Readme.de). 
2. new class SymbolicSphericalPotential. 
3. new symbolic potential EarthPREMPotential (earth potential based on Preliminary reference Earth model). 
4. Related tests are also included.

### DIFF
--- a/tests/test_BurkertSymbolicPotential.py
+++ b/tests/test_BurkertSymbolicPotential.py
@@ -203,7 +203,7 @@ def _test_efficiency():
     import time
 
     pot_num = BurkertPotential(amp=2.1341897, a=1.13124)
-    pot_sym = BurkertSymbolicPotential(amp=2.1341897, a=1.13124, backend='numpy')
+    pot_sym = BurkertSymbolicPotential(amp=2.1341897, a=1.13124, backend="numpy")
     r_vals = numpy.logspace(0, 10, 1_000_000)
 
     print(f"\n")


galpy/potential/SymbolicSphericalPotential.py
```diff
@@ -0,0 +1,126 @@
+###############################################################################
+#   SymbolicSphericalPotential.py: base class for symbolic potentials
+#   corresponding to spherical density profiles
+###############################################################################
+import numpy
+
+from ..util._optional_deps import _SYMPY_LOADED
+from .Potential import Potential
```
jobovy:
```suggestion
from .Potential import SphericalPotential
```

galpy/potential/SymbolicSphericalPotential.py
```diff
@@ -0,0 +1,126 @@
...
+
+if _SYMPY_LOADED:
+    import sympy
+
+
+class SymbolicSphericalPotential(Potential):
```
jobovy:
```suggestion
class SymbolicSphericalPotential(SphericalPotential):
```

galpy/potential/SymbolicSphericalPotential.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+    def _dens(self, R, z, phi=0.0, t=0.0):
+        r = numpy.sqrt(R**2.0 + z**2.0)
+        return self._rdens(r, t=t)
+
+    def _mass(self, R, z, phi=0.0, t=0.0):
+        R, z = numpy.float64(R), numpy.float64(z)  # Avoid indexing issues
+        r = numpy.sqrt(R**2.0 + z**2.0)
+        return self._rmass(r=r)
```
jobovy:
```suggestion
```
By inheriting from ``SphericalPotential``, all you need to do is implement the ``_revaluate``, ``_rforce``, ``_r2deriv``, and ``_rdens`` functions. All of the others are then implemented in ``SphericalPotential``

galpy/potential/__init__.py
```diff
@@ -10,6 +10,7 @@
     DehnenSmoothWrapperPotential,
     DiskSCFPotential,
     DoubleExponentialDiskPotential,
+    EarthPREMPotential,
```
jobovy:
```suggestion
    EarthPREMPotential,
    SymbolicSphericalPotential,
```
Also expose the base symbolic potential so people can implement their own one easily.

galpy/potential/__init__.py
```diff
@@ -213,6 +214,7 @@
     AnyAxisymmetricRazorThinDiskPotential.AnyAxisymmetricRazorThinDiskPotential
 )
 AnySphericalPotential = AnySphericalPotential.AnySphericalPotential
+EarthPREMPotential = EarthPREMPotential.EarthPREMPotential
```
jobovy:
```suggestion
EarthPREMPotential = EarthPREMPotential.EarthPREMPotential
SymbolicSphericalPotential = SymbolicSphericalPotential. SymbolicSphericalPotential
```

jobovy (changes requested):
Thanks for opening the PR! Here are some first review comments. Currently, the test suite fails because the ``EarthPREMPotential`` gets picked up automatically for orbit tests (all potentials are automatically picked up for those unless they are excluded). But rather than directly fixing that issue, I'd rather that this PR doesn't add ``EarthPREMPotential`` at all, because I don't think it's a useful potential to add to ``galpy``. Instead, it could be used in a test as a test case of ``SymbolicSphericalPotential``  and/or in documentation of ``SymbolicSphericalPotential``. A better test within galpy would just be to take something like a ``LogarithmicHaloPotential`` and check that the symbolic version agrees with the numerical version that's already implemented. Tests of that would have to be added to the test suite and sympy would have to be installed in the test suite for those tests to run.

Take a look at https://docs.galpy.org/en/latest/installation.html#development-installation for some development pointers, including how to run pre-commit locally (currently pre-commit fails because there are some suspected typos in the code).

Yuzhe98:
Hi Jo,
I sent some test results to you by email. Don't forget to check

PR #752, opened 2025-08-26
include rtol and atol in integration of orbits

jennywagner:
In order to make the precision of the integration routines accessible to users, I included the rtol and atol parameters into the integrate-function for orbits. (rtol and atol were already accessible from all other subroutines, just not in the higher-level function that calls orbit integration.)


jobovy:
Looking through the `galpy` code while reviewing this pull request, I found a small bug in how `rtol` was handled by some of the C integrators. I just fixed that, so I rebased this pull request to include that fix. Please make sure to `git pull` before making changes, otherwise things will get messy, thanks!

PR #760, opened 2025-09-29
Potential/einasto

johnwez1:
Added the Einasto density profile using many of results and conventions in https://arxiv.org/pdf/1202.5242


johnwez1:
Thanks Jo. I'll add the other options in.

PR #861, opened 2026-04-14
Smooth stream track for streamspraydf models

jobovy:
## Summary

- Adds `StreamTrack` (and `StreamTrackPair` for `tail='both'`) — a lightweight smooth phase-space track parameterized by the progenitor's time coordinate `tp` in `[-tdisrupt, 0]`. Exposes `Orbit`-like accessors for Cartesian, cylindrical, and heliocentric coordinates (`x`/`y`/`z`/`vx`/`vy`/`vz`, `R`/`vR`/`vT`/`phi`, `ra`/`dec`/`dist`/`ll`/`bb`/`pmra`/`pmdec`/`pmll`/`pmbb`/`vlos`) plus a `cov(tp)` and `plot(d1, d2, spread=...)`.
- Adds `basestreamspraydf.streamTrack(n=5000, tail=..., smoothing=..., niter=0, particles=..., order=2)` which samples particles (or reuses pre-computed ones) and fits the track. The track is the smoothed binned mean of particles as a function of `tp=-dt` with a smoothed 6×6 covariance.
- Tests, API reference docs, and a new notebook section in the spraydf tutorial.

A comment in `streamdf.py` points at `StreamTrack` as a candidate shared backend for a future follow-up PR that could simplify streamdf's track plumbing.

## Test plan

- [x] 11 new unit tests in `tests/test_streamspraydf.py` pass (progenitor recovery, sample consistency, interface, PSD covariance, both arms, iteration, chen24+fardal15, center/satellite streams, physical units, particle reuse, plot smoke).
- [x] Existing `test_streamspraydf.py` tests still pass (28 total passing locally).
- [x] Notebook section uses existing `spdf_both` fixture; cells run interactively against a local execution.


jobovy:
## StreamTrack implementation alternatives — comparison summary

Seven alternative branches were explored, each changing one axis of the algorithm. All branch off `streamspray-track` and are pushed (no PRs). Comparison plots (Bovy14 + Pal 5 + warm 10^7 Msun streams, aligned phi_1/phi_2 coordinates) are in the repo root as `alt_*.png` files.

### Main branch (`streamspray-track`)
Closest-point projection (3D position) onto a finely-integrated, short-range progenitor orbit (`0.03*tdisrupt`). Bin-then-smooth offsets-from-progenitor with `UnivariateSpline` (chi-square s=N). Percentile-trimmed tp_grid.

### Alternatives tested

| Branch | What it changes | Cold stream (Bovy14) | Pal 5 |
|---|---|---|---|
| `alt-gcv` | GCV auto-smoothing (`make_smoothing_spline`) | Eliminates leading-arm wiggle | Clean |
| `alt-no-binning` | Per-particle splines, no binning | Slight tip oscillation | Small spikes at tips |
| `alt-rotating-frame` | Smooth in progenitor-aligned rotating frame | Clean | Clean |
| `alt-6d-closest` | 6D (position+velocity) closest-point | Clean | Clean |
| `alt-strip-time-affine` | tp = arm_sign × dt (no closest-point) | **Dramatic wrap-aliasing failure** | Same |
| `alt-auto-timerange` | Data-driven track_time_range from particle spread | Clean | Clean |
| `alt-kdtree` | cKDTree for closest-point (speed) | Identical to main | Identical |

### Key findings

1. **`alt-strip-time-affine` (tp=dt directly) fails catastrophically** for multi-wrap streams. The track loops across the sky multiple times. This validates the closest-point matching approach.

2. **`alt-gcv` (GCV smoothing) fixes the leading-arm wiggle** visible in the main branch's Bovy14 fardal15 track. The main branch's chi-square s=N criterion slightly under-smooths; GCV auto-selects a better lambda.

3. **`alt-auto-timerange` is essential for warm streams** (see next comment).

4. `alt-no-binning`, `alt-rotating-frame`, `alt-6d-closest`, `alt-kdtree` produce tracks within ≤0.05 galpy units of main for cold streams — the algorithmic choices they test are secondary for typical use.

### Recommendation

Combine: **auto-timerange** (scales with stream width) + **GCV smoothing** (eliminates under-smoothing) + **6D matching** (robustness for eccentric orbits) + **boundary pile-up exclusion** + percentile trimming. Require `scipy >= 1.10` (released Jan 2023).

### Comparisons

#### Main branch (`streamspray-track`, PR #861)
Closest-point projection of each particle onto a finely-integrated, short-range (`0.03*tdisrupt` by default) progenitor orbit that spans both sides of `tp=0`. Bin-then-smooth offsets-from-progenitor with `UnivariateSpline` (chi-square-like auto-s). Covariance smoothed per-entry and PSD-projected.

The public `tp_grid` is **trimmed to the 99th percentile** of the
assigned `tp` values (symmetric 1st percentile on the trailing side),
so a handful of outlier particles matched to sparsely-populated
boundary bins can't drag the smoothing spline around at the tips.
`trim_percentile` is a local constant in `StreamTrack.__init__` — we can
expose it as a kwarg later if there's demand.

**Bovy14 stream** (LogarithmicHaloPotential, q=0.9, tdisrupt=4.5 Gyr):

image

**Pal 5 stream** (MWPotential2014, Orbit.from_name("Pal 5"), tdisrupt=5 Gyr):

image

All 7 alternatives below inherit the same trim and can be toggled by `git checkout alt-<name>`. Each plot now has two panels: an aligned `(phi_1, phi_2)` view with the alt's track on top of the sample cloud and the `streamspray-track` reference as dashed black lines, plus a difference curve per arm against the saved reference.

> **What the bottom panel plots.** The y-axis on the difference panel is the 3D Euclidean distance between the two tracks' **galactocentric Cartesian positions** at each `tp`: 
>
> `d(tp) = ||(x, y, z)_alt(tp) − (x, y, z)_main(tp)||`
>
> in **galpy internal length units** (multiply by `ro ≈ 8` for kpc). Not `phi_2`, not a single coordinate. Velocities are ignored in the diff metric; they're smoothed the same way as position, so positional agreement is a good proxy. 

##### 1. `alt-gcv` — auto-smoothing via GCV

`scipy.interpolate.make_smoothing_spline` replaces `UnivariateSpline`; the smoothing parameter is selected automatically by Generalized Cross Validation, so the user doesn't tune `s`.

- **Pros**: no smoothness knob; data-driven bias/variance tradeoff.
- **Cons**: requires `scipy>=1.10`; slower per fit.

image
image

##### 2. `alt-no-binning` — per-particle smoothing splines

Skip binning entirely. Each particle contributes one `(tp_i, offset_i)` point to six cubic smoothing splines; covariance comes from smoothing splines on residual outer-products.

- **Pros**: no `ntp` knob, no empty-bin edge cases, adapts to uneven particle density along tp.
- **Cons**: fits N points rather than ntp (slower for large n); FITPACK occasionally warns when s is mis-estimated from pairwise noise.

image
image

##### 3. `alt-rotating-frame` — smooth in the progenitor-aligned frame

At each particle's tp, rotate the raw offset into a progenitor-aligned frame (L along z, progenitor at +x). Smooth in that frame, then rotate back. This was the plan's original proposal.

- **Pros**: rotated offsets carry clean physical meaning (along-stream / transverse / vertical); dynamic range of the smoothed signal is minimal.
- **Cons**: per-tp rotation matrices add bookkeeping; interpolated matrices need SVD re-orthogonalization.

image
image

##### 4. `alt-6d-closest` — closest-point in 6D instead of 3D

Match particles to the progenitor orbit in full 6D phase space (position + velocity, both in galpy internal units) rather than 3D position alone.

- **Pros**: unambiguous at orbit self-intersections (same xyz, different v); cleaner tp assignments near pericenter/apocenter.
- **Cons**: larger distance-matrix memory; the position/velocity weighting is fixed by the choice of units rather than physically motivated.

image
image

##### 5. `alt-strip-time-affine` — tp = arm_sign * dt

Simplest possible mapping from particles to the affine parameter: use the known stripping time `dt` directly (positive for leading, negative for trailing). `track_time_range` defaults to `tdisrupt` so the progenitor is integrated over the full `[-tdisrupt, +tdisrupt]`.

- **Pros**: fully deterministic, no closest-point failure modes, exact per-particle affine value.
- **Cons**: tp and along-stream position are correlated with scatter (ΔΩ varies per particle), so the mean-track at a given tp isn't exactly on the stream's median curve. Three tests are relaxed / one skipped on this branch because their tp semantics assume the closest-point mapping.

image
image

##### 6. `alt-auto-timerange` — adaptive track_time_range from particle spread

Replace the fixed `0.03*tdisrupt` default with a data-driven estimate: measure the stream's spatial extent from a probe sample, convert to orbital-time via progenitor speed, pad by 4x, clamp to `[1, tdisrupt]`.

- **Pros**: adapts to narrow vs wide streams automatically; safe across a wider range of progenitor/stream regimes.
- **Cons**: needs one small extra sample draw when `particles=` not supplied; speed metric can be fragile for highly eccentric orbits.

image
image

##### 7. `alt-kdtree` — KD-tree closest-point matching

Replace the O(N*M) pairwise distance matrix with a `scipy.spatial.cKDTree` query. When dt/arm-sign masks restrict the allowed neighbors per particle, query K nearest and pick the closest allowed one, growing K 4x per pass until every point has a match.

- **Pros**: sublinear per-query complexity; no O(N*M) memory footprint (matters for very large N or denser progenitor grids).
- **Cons**: a Python-level fallback loop runs when masks are restrictive (still fast for N up to ~1e5).

image
image

jobovy:
## Warm-stream comparison (10^7 Msun progenitor)

To test robustness for wider streams (dwarf-galaxy-mass progenitors), we ran all alternatives with the Bovy14 orbit but 1000× larger progenitor mass (10^7 Msun instead of 10^4 Msun). This produces a much fatter stream with ~10× larger tidal radius and velocity kicks.

### Results

The **main branch (3D closest-point, `track_time_range = 0.03*tdisrupt = 3.8`)** fails: the leading arm track flies off to phi_2 ~ -35, far from the sample cloud. The root cause is **insufficient progenitor orbit coverage**: with only 3.8 galpy time units (~0.3 orbital periods), far-flung warm-stream particles can't find a close match on the short progenitor arc and pile up at the boundary.

| Branch | Warm stream result | Why |
|---|---|---|
| main (3D, short range) | **Broken** — leading arm flies off | Progenitor arc too short for wide stream |
| `alt-6d-closest` (6D, short range) | Still problematic | Same short arc; 6D can't help if orbit isn't long enough |
| `alt-gcv` (GCV, short range) | Same failure | Better smoothing can't fix bad tp assignment |
| `alt-rotating-frame` (short range) | Same failure | Better frame can't fix bad tp assignment |
| **`alt-auto-timerange`** (3D, data-driven range) | **Works** — track follows cloud | Auto range = 13.6 (~1.1 orbital periods), covers all particles |

### Key numbers

- Cold stream: `d_max = 1.7`, auto range = 5.2 (1.4× main)
- **Warm stream: `d_max = 4.4`, auto range = 13.6 (3.6× main)**

The auto-timerange scales naturally with stream width because it measures the actual particle spread. The fixed `0.03*tdisrupt` doesn't adapt.

### Conclusion

For warm streams, **progenitor orbit coverage is the primary bottleneck**, not matching dimension or smoothing method. The auto-timerange fix is essential. 6D matching and GCV smoothing are secondary improvements that help with cold-stream edge cases (orbit self-intersection, under-smoothing) but cannot compensate for insufficient orbit coverage.

### Boundary pile-up

Even with auto-timerange, particles at the edge of the progenitor arc are mismatched (their true closest point is outside the arc). These create a "pile-up" at the extreme tp values that biases the track. Fix: exclude particles whose tp_assign equals the far boundary of `track_t_grid` before fitting.

### Example warm stream

image

jobovy:
## Warm-stream comparison (10^7 Msun progenitor, TriaxialNFW not included here)

Bovy14 orbit in LogarithmicHaloPotential(q=0.9), tdisrupt=4.5 Gyr, fardal15spraydf. Six projections: aligned sky (phi_1, phi_2), galactocentric (x, y), cylindrical (R, z), equatorial (RA, Dec), Galactic (l, v_los), heliocentric (dist, pm_ra).

### Cold stream (10^4 Msun) — main branch

warm_test_cold_1e4

### Warm stream (10^7 Msun) — main branch

warm_test_warm_1e7

### Key findings

- **Cold stream**: tracks follow the sample cloud cleanly in all six projections. GCV smoothing produces a smooth curve with no leading-arm wiggle.
- **Warm stream**: tracks follow the sample cloud in all projections. The auto-timerange correctly picks track_time_range ≈ 27 (vs 3.8 for cold), covering the larger spatial extent. At every tp evaluation point, 74–436 sample particles are within 3 kpc of the track — confirming the track is correctly tracing through the thick cloud.
- Some sky-coordinate panels show straight-line artifacts where matplotlib connects tp-consecutive points across the RA=0/360 wrap boundary. This is a visualization issue, not a track error.

### Diagnostics

| | Cold (10^4) | Warm (10^7) |
|---|---|---|
| auto track_time_range | ~3.4 | ~27 |
| leading tp range | [0, 1.1] | [0, 11.8] |
| particles after boundary exclusion | 3000/3000 | 3000/3000 |

jobovy:
## Triaxial NFW comparison: main vs stream-orbit v1 vs v2

TriaxialNFWPotential(normalize=1, a=20/8, b=0.8, c=0.6), Bovy14 orbit, tdisrupt=2 Gyr, n=1000 particles. Three projections: aligned sky, (x, y), (R, z).

### Cold stream (10^4 Msun)

**Main** (progenitor orbit only):
triaxial_main_cold_1e4

**v1** (one-shot orbit from raw particle-mean tip IC):
triaxial_v1_cold_1e4

**v2** (two-step orbit from smoothed track tip IC):
triaxial_v2_cold_1e4

**Cold verdict**: All three nearly identical. The cold stream stays close to the progenitor orbit even in the triaxial potential.

---

### Warm stream (10^7 Msun)

**Main** (progenitor orbit only):
triaxial_main_warm_1e7

**v1** (one-shot orbit from raw particle-mean tip IC):
triaxial_v1_warm_1e7

**v2** (two-step orbit from smoothed track tip IC):
triaxial_v2_warm_1e7

**Warm verdict**:
- **Main**: both arms follow the thick cloud in all projections. Clean.
- **v1**: trailing OK, but **leading arm shoots off to R~26 in (R,z)** — the noisy particle-mean tip IC leads to an orbit that diverges from the actual stream.
- **v2**: matches main quality. The smoothed-track tip IC avoids the v1 divergence.

---

### Summary

| | Cold (1e4) | Warm (1e7) |
|---|---|---|
| **Main** (progenitor only) | Clean | Clean |
| **v1** (particle-mean tip) | Clean | Leading diverges in (R,z) |
| **v2** (smoothed-track tip) | Clean | Clean, matches main |

The main branch already handles the triaxial warm case well — the progenitor orbit is a good enough reference. v2 matches main; v1 still fails from IC noise. The stream-orbit refinement might become more impactful for more extreme triaxiality (b < 0.6) or very long tdisrupt where cumulative orbital precession builds up.

jobovy:
## Triaxial NFW comparison (updated: 5 Gyr tdisrupt)

TriaxialNFWPotential(normalize=1, a=20/8, b=0.8, c=0.6), Bovy14 orbit, **tdisrupt=5 Gyr**, n=1000 particles.

### Cold stream (10^4 Msun)

**Main** (progenitor orbit only):
triaxial_main_cold_1e4 (1)

**v1** (one-shot orbit from raw particle-mean tip IC):
triaxial_v1_cold_1e4 (1)

**v2** (two-step orbit from smoothed track tip IC):
triaxial_v2_cold_1e4 (1)

---

### Warm stream (10^7 Msun)

**Main** (progenitor orbit only):
triaxial_main_warm_1e7 (1)

**v1** (one-shot orbit from raw particle-mean tip IC):
triaxial_v1_warm_1e7 (1)

**v2** (two-step orbit from smoothed track tip IC):
triaxial_v2_warm_1e7 (1)

---

### Summary

| | Cold (1e4) | Warm (1e7) |
|---|---|---|
| **Main** (progenitor only) | Clean | Clean |
| **v1** (particle-mean tip) | Clean | Leading arm diverges |
| **v2** (smoothed-track tip) | Clean | Clean, matches main |

With 5 Gyr of disruption in the triaxial potential, the main branch still handles both cold and warm streams well. v1's noisy IC diverges for the warm case; v2's smoothed IC fixes this and matches main quality. The stream-orbit refinement doesn't visibly improve over using the progenitor orbit directly at this level of triaxiality (b=0.8, c=0.6).

jobovy:
## Status update: warm-stream testing, combined improvements, and stream-orbit exploration

### Warm-stream testing (10^7 Msun progenitor)

Stress-tested the algorithm with a 1000x larger progenitor mass. The fixed `track_time_range = 0.03 * tdisrupt` was too short for warm streams — particles spread far from the progenitor couldn't find matches on the short progenitor arc, causing boundary pile-up and a broken track. Fix: **auto-timerange from particle spread** (`8 * d_max / |v_prog|`), which scales naturally with stream width.

### Combined implementation

Merged four orthogonal improvements, each validated by the branch comparisons:

- **GCV smoothing** (`make_smoothing_spline`, scipy ≥ 1.10) as default — eliminates the leading-arm under-smoothing wiggle. `smoothing=<float>` still available for explicit `s`.
- **6D closest-point matching** (position + velocity) — more robust at orbital self-intersections. No cost for cold streams.
- **Auto-timerange** — essential for warm streams, harmless for cold.
- **Boundary pile-up exclusion** — removes particles matched to the far edge of the progenitor arc before fitting.

### Warm-stream deep dive

After initially thinking the warm stream's track broke in 3D projections, discovered it was a plotting bug (`pyplot.sca(ax)` missing). The track is correct in all projections — verified numerically (74–436 sample particles within 3 kpc of each track evaluation point).

### Stream-orbit refinement exploration (`alt-stream-orbit`)

Explored using a "stream orbit" integrated from the stream tip as the smoothing reference instead of the progenitor orbit:

- **v1** (one-shot from raw particle-mean tip IC): diverges — noisy IC amplified over integration. Track flies outside sample cloud.
- **v2** (two-step: fit preliminary track on progenitor → evaluate smooth track at tip for noise-free IC → integrate stream orbit from that): works — avoids divergence, matches main-branch quality.
- **Conclusion**: v2 reliably avoids v1's noise problem, but for the streams tested (LogarithmicHaloPotential and TriaxialNFW b=0.8, c=0.6, tdisrupt=5 Gyr), the main branch (progenitor orbit only) already works well. The refinement doesn't visibly improve the track. Left as experimental on `alt-stream-orbit`.

### Triaxial NFW comparison

Ran main vs v1 vs v2 in TriaxialNFWPotential(b=0.8, c=0.6), tdisrupt=5 Gyr, for both cold (10^4 Msun) and warm (10^7 Msun) streams. Main handles both cleanly; v1 diverges for warm; v2 matches main. See earlier PR comments for plots.

### Timing (track construction only, excluding particle sampling)

| | n=1000 | n=3000 |
|---|---|---|
| **Main** | ~1.0s | ~2.3s |
| **v2 (stream-orbit)** | ~2.0s | ~4.6s |

v2 is 2x main (fits the track twice). Progenitor integration is negligible (<60ms). Cost dominated by the N×M distance matrix and GCV spline fits. Potential choice and progenitor mass don't affect timing.

### Current state

- `streamspray-track`: GCV + 6D + auto-timerange + boundary exclusion + percentile trim. 23 tests passing.
- `alt-stream-orbit`: experimental v2 refinement. Works but doesn't improve over main for tested potentials.
- 7 other alternative branches available for reference.

jobovy:
## Timing benchmarks (current implementation)

All times are **track construction only** — particle sampling/orbit integration excluded. The `particles=` argument lets users pre-sample once and reuse; sampling cost scales linearly with `n` and depends on the potential (e.g., ~17s for n=3000 in LogarithmicHalo, ~170s in TriaxialNFW).

scipy 1.17.1, `make_smoothing_spline` (GCV) for auto-smoothing.

### LogarithmicHaloPotential (q=0.9), tdisrupt=4.5 Gyr

| n | order=1 (mean) | order=2 (mean+cov) | reuse-s (order=2) |
|---|---|---|---|
| 1000 | 0.13s | 0.50s | 0.09s |
| 3000 | 0.22s | 0.75s | 0.17s |
| 5000 | 0.33s | 0.98s | 0.24s |

Cold (10⁴ M☉) and warm (10⁷ M☉) streams give essentially identical timing — progenitor mass doesn't affect track construction cost.

### TriaxialNFWPotential (b=0.8, c=0.6), tdisrupt=5 Gyr

| n | order=1 (mean) | order=2 (mean+cov) | reuse-s (order=2) |
|---|---|---|---|
| 1000 | 0.19s | 0.52s | 0.14s |
| 3000 | 0.28s | 0.79s | 0.22s |

Slightly higher than LogarithmicHalo due to the longer progenitor arc (5 vs 4.5 Gyr), but same order of magnitude.

### Cost breakdown

- **GCV spline fitting** dominates: order=2 fits 27 splines (6 mean + 21 unique covariance elements) vs. 6 for order=1, explaining the ~3x difference.
- **KDTree matching**: negligible (<50ms for n=3000). This replaced an N×M distance matrix that was ~1.6s for n=3000.
- **Progenitor integration**: <1ms (already integrated during spraydf setup).
- **`smoothing_s` reuse**: bypasses GCV entirely, using `UnivariateSpline(s=...)` with the stored per-spline smoothing values. Gives **3–4x speedup** over GCV for order=2. Useful when re-fitting with different particles but similar stream properties.

### Iteration (`niter`)

Each iteration re-matches particles to the current smooth track and refits. Cost is ~0.1s per iteration (n=3000, order=1):

| niter | time (order=1) |
|---|---|
| 0 | 0.21s |
| 1 | 0.31s |
| 2 | 0.41s |
| 3 | 0.51s |

### Summary

Track construction is fast (sub-second for typical use). The dominant cost is always the upstream particle sampling, which scales with `n` and potential complexity. The `particles=` interface and `smoothing_s` reuse are the two main levers for avoiding redundant work.

doc/source/reference/df.rst
```diff
@@ -361,3 +362,102 @@ Specific particle-spray models
...
    Chen et al. (2024) <streamspraydfchen24.rst>
    Fardal et al. (2015) <streamspraydffardal15.rst>
+
+Smooth stream-track object
++++++++++++++++++++++++++++
+
+.. toctree::
+   :maxdepth: 1
+
+   StreamTrack and StreamTrackPair <streamtrack.rst>
```
jobovy: @claude Please split up the API pages for StreamTrack and StreamTrackPair

doc/source/reference/df.rst
```diff
@@ -361,3 +362,102 @@ Specific particle-spray models
...
+
+General instance routines
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. toctree::
+   :maxdepth: 1
+
+   __call__ <streamtrackcall.rst>
+   from_particles <streamtrackfromparticles.rst>
```
jobovy, line 413: @claude Rather than just putting this here as a method, include a brief discussion of how to set these up, mentioning ``from_particles`` as a way to do it. Does StreamTrackPair have ``from_particles``?

doc/source/reference/df.rst
```diff
@@ -361,3 +362,102 @@ Specific particle-spray models
...
+
+These require ``custom_transform`` to have been set at construction.
+
+.. toctree::
+   :maxdepth: 1
+
+   phi1 <streamtrackphi1.rst>
+   phi2 <streamtrackphi2.rst>
+   pmphi1 <streamtrackpmphi1.rst>
+   pmphi2 <streamtrackpmphi2.rst>
```
jobovy: @claude In keeping with other pages, don't split up the methods by type of accessor, just have a single list

doc/source/reference/df.rst
```diff
@@ -361,3 +362,102 @@ Specific particle-spray models
...
+
+StreamTrackPair routines
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Attribute:
+
+* .. autoattribute:: galpy.df.StreamTrackPair.particles
```
jobovy, line 458: @claude don't we need similar documentation for StreamTrack's particles attribute?

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        positions (matched by the leading arm). Because stream particles
+        have small velocity offsets from the progenitor, they lie spatially
+        close to a short arc of the progenitor's orbit — the relevant ``tp``
+        range is much smaller than ``tdisrupt``.
+
+        Parameters
+        ----------
+        n : int, optional
+            Number of particles to draw (per arm when ``tail='both'``).
+            Ignored if ``particles`` is provided. Default is 5000.
```
jobovy: @claude It doesn't seem ideal that for streamTrack with tail='both' n is the number of particles / arm, while for sample it's the total number of particles. So we should probably switch to total number here for consistency

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        track_time_range : float or Quantity, optional
+            Half-range (symmetric about tp=0) of the finely-integrated
+            progenitor orbit used for closest-point matching. Default is
+            data-driven: ``8 * d_max / |v_prog|`` clamped to ``[1,
+            tdisrupt]``, where ``d_max`` is the farthest particle's
+            distance from the progenitor.
+        track_n_dense : int, optional
+            Number of time points on the finely-integrated progenitor
+            orbit. The actual grid has ``2 * (track_n_dense+1)//2 - 1``
+            points (forward + backward with shared t=0). Default 10001.
```
jobovy: @claude Is there a good reason for this to be exposed? Seems like it could just be set to a high-enough value and not be adjusted.

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        ntp : int, optional
+            Number of binning nodes. Default ``sqrt(N)`` clipped to
+            ``[21, 201]``.
```
jobovy: @claude Not clear that clipping to 201 is a great thing to do for very long streams. Maybe the clip should be tied to track_time_range? That is, have a larger max for longer streams.

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        ninterp : int, optional
+            Resolution of the public fine-grid track arrays. Default 1001.
```
jobovy: @claude Again, not clear that this needs to be exposed and that it wouldn't be better to just set it and forget it

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        smoothing : None, float, array-like, or dict, optional
+            Smoothing parameter(s). ``None`` (default) uses GCV
+            auto-tuning. A float sets a single ``s`` for all coords. An
+            array-like of length 6 (mean only) or 27 (mean + covariance)
+            sets per-spline ``s`` values — pass a previous call's
+            ``track.smoothing_s`` to reproduce the same smoothness
+            without re-running GCV. A dict keyed by coordinate name
+            sets per-coordinate ``s`` for the 6 mean splines only.
```
jobovy: @claude What is the dict input useful for? Is there a use case where this is to be preferred given the other methods? Maybe remove as a possibility

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+            computing 6D distances during the closest-point projection.
+            Default ``'auto'`` learns the weight from the inner-half
+            particle dispersion (``σ_pos / σ_vel``, clipped to
+            ``[1, 10]``); typically lands at ~2–3 for both clean and
+            perturbed streams. Values > 1 make velocity matches more
+            important than position matches — useful when the
+            progenitor orbit revisits regions of phase space (e.g., in
+            strongly-perturbed potentials with a massive LMC). Pass
+            ``1.0`` for the legacy unweighted Galpy-natural-units
+            metric.
```
jobovy: @claude Galpy-natural-units --> natural-units

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        :class:`galpy.df.StreamTrack` or :class:`galpy.df.StreamTrackPair`
+            A single-arm track object, or a pair with ``.leading`` and
+            ``.trailing`` tracks when ``tail='both'``.
+
+        Notes
+        -----
+        - 2026-04-14 - Written - Bovy (UofT)
+        """
+        from ..orbit import Orbit
+        from .streamTrack import StreamTrack, StreamTrackPair
```
jobovy: @claude unless they need to be within a function to avoid circular imports, move all imports to the top of the files

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+                track_time_range, ro=self._ro, vo=self._vo
+            )
+
+        # Build a finely-sampled progenitor phase-space array spanning
+        # [-T, +T] around the present day. Integrate forward and backward
+        # separately from the progenitor's present-day state, then combine.
+        # Use the BASE potential (without progpot's MovingObjectPotential)
+        # because the MovingObjectPotential's internal progenitor was only
+        # integrated on [-tdisrupt, 0] and would give wrong or erroring
+        # results for positive times.
```
jobovy: @claude This explanation for excluding the progenitor potential is not correct. The progenitor potential shouldn't be included to integrate itself, that doesn't make sense. So perhaps just remove the explanation and just say that we use the base potential

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        _track_pot = self._pot
+        if hasattr(self, "_orig_pot"):
+            _track_pot = self._orig_pot
```
jobovy: @claude Replace with ternary assignment

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        track_prog_cart = numpy.column_stack(
+            [
+                numpy.concatenate([prog_back.x(t_back)[::-1], prog_fwd.x(t_fwd)[1:]]),
+                numpy.concatenate([prog_back.y(t_back)[::-1], prog_fwd.y(t_fwd)[1:]]),
+                numpy.concatenate([prog_back.z(t_back)[::-1], prog_fwd.z(t_fwd)[1:]]),
+                numpy.concatenate([prog_back.vx(t_back)[::-1], prog_fwd.vx(t_fwd)[1:]]),
+                numpy.concatenate([prog_back.vy(t_back)[::-1], prog_fwd.vy(t_fwd)[1:]]),
+                numpy.concatenate([prog_back.vz(t_back)[::-1], prog_fwd.vz(t_fwd)[1:]]),
+            ]
+        )
```
jobovy: @claude this should all be able to be done using a single orbit and using orbit integration continuation. First integrate forward, then backward, and the integrations will be automatically stitched together in the way that you need.

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+        # Inherit unit metadata from the original progenitor Orbit. Pass
+        # ``ro``/``vo`` only when the progenitor had them explicitly set —
+        # StreamTrack mirrors Orbit's "ro/vo unset means use the config
+        # default and keep _roSet=False" pattern, so we propagate the
+        # progenitor's ``_roSet``/``_voSet`` state via *not passing* the
+        # value rather than via a separate flag.
+        prog_ro = self._orig_progenitor._ro if self._orig_progenitor._roSet else None
+        prog_vo = self._orig_progenitor._vo if self._orig_progenitor._voSet else None
+        prog_zo = self._orig_progenitor._zo
+        prog_sm = self._orig_progenitor._solarmotion
```
jobovy, line 431: @claude This should be tied in with the general spraydf object's physical conversions for ro/vo. Those should be consistent with those of the progenitor (should be done in __init__). For zo and solarmotion, yes, use the progenitor.

galpy/df/streamspraydf.py
```diff
@@ -207,6 +216,239 @@ def sample(self, n, return_orbit=True, returndt=False, integrate=True):
...
+                + float(self._progenitor.vy(0.0)) ** 2
+                + float(self._progenitor.vz(0.0)) ** 2
+            )
+            _d_max = numpy.sqrt(
+                numpy.max((_xs - _px) ** 2 + (_ys - _py) ** 2 + (_zs - _pz) ** 2)
+            )
+            track_time_range = float(
+                numpy.clip(8.0 * _d_max / max(_pv, 1e-6), 1.0, self._tdisrupt)
+            )
+        else:
```
jobovy: @claude It's not ideal that this estimate is done based on a probe rather than just using the sampled particles that will be used for the track itself. So switch to re-using those, even if that requires a rejiggering of this code. Estimate will be more precise and be more efficient

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+    from astropy import units
+
+
+def _particles_to_cartesian(xv_particles):
+    """Convert particles from (R,vR,vT,z,vz,phi) to galactocentric
+    Cartesian 6-vectors. Returns array of shape (N, 6)."""
+    R, vR, vT, z, vz, phi = xv_particles
+    x_p, y_p, z_p = coords.cyl_to_rect(R, phi, z)
+    vx_p, vy_p, vz_p = coords.cyl_to_rect_vec(vR, vT, vz, phi)
+    return numpy.column_stack([x_p, y_p, z_p, vx_p, vy_p, vz_p])
```
jobovy: @claude It seems worthwhile to just adding ``galcencyl_to_galcenrect`` to ``util.coords`` and then use it here. In case in the future it's useful elsewhere

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+    Uses ``scipy.spatial.cKDTree`` for the unmasked case (sublinear
+    per query). When a per-particle ``mask`` of shape (N, M) is given,
+    queries K nearest neighbors and picks the closest allowed one,
+    growing K until every point has a match.
+
+    ``velocity_weight`` (default 1.0): when D == 6, scale the velocity
+    components (last 3 columns) by this factor before computing distances.
+    See :func:`_fit_track_from_particles` for usage and ``'auto'`` mode.
+    """
+    from scipy.spatial import cKDTree
```
jobovy: @claude Again, please move imports to the top of the file

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            allowed = cand[j][mask[r][cand[j]]]
+            if allowed.size:
+                chosen[j] = allowed[0]
+        hit = chosen >= 0
+        tp_out[remaining[hit]] = curve_t[chosen[hit]]
+        remaining = remaining[~hit]
+        if k >= M:
+            tp_out[remaining] = 0.0
+            break
+        k = min(k * 4, M)
```
jobovy, line 181: @claude Simplify based on plan in frolicking-exploring-hamster.md

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+    order = int(order)
+
+    prog_splines = [
+        interpolate.InterpolatedUnivariateSpline(track_t_grid, prog_cart[:, i], k=3)
+        for i in range(6)
+    ]
+
+    def _prog_at(tp):
+        tp = numpy.atleast_1d(tp)
+        return numpy.column_stack([spl(tp) for spl in prog_splines])
```
jobovy: @claude why not just make use of the spline interpolation done inside Orbit objects?

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            idx_inner = numpy.argmin(
+                numpy.abs(track_t_grid[None, :] - probe_tp[inner, None]), axis=1
+            )
+            prog_at_inner = prog_cart[idx_inner]
+            dpos = particles_cart[inner, :3] - prog_at_inner[:, :3]
+            dvel = particles_cart[inner, 3:] - prog_at_inner[:, 3:]
+            sigma_pos = numpy.sqrt(numpy.mean(numpy.sum(dpos**2, axis=1)))
+            sigma_vel = numpy.sqrt(numpy.mean(numpy.sum(dvel**2, axis=1)))
+            if sigma_vel > 0:
+                velocity_weight = float(numpy.clip(sigma_pos / sigma_vel, 1.0, 10.0))
```
jobovy: @claude Why clip to a minimum of 1? Maybe for some streams position differences are larger than velocity differences in galpy units? So maybe clip to something lower than 1

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+                f"streamTrack: tp_assign histogram has a gap spanning "
+                f"{gap_span / populated_span:.0%} of the populated range. "
+                f"{n_far} of {abs_tp.size} particles ({n_far / abs_tp.size:.1%}) "
+                "are isolated past the gap and will pile up at the boundary "
+                "of the smoothed track, likely producing a kink. This "
+                "typically happens when the progenitor orbit revisits its "
+                "phase-space trajectory (e.g., near apocenter under strong "
+                "perturbation). Consider passing velocity_weight>=2 (preferred), "
+                "or niter>=2 with the per-iteration retrim, or both.",
+                galpyWarning,
```
jobovy, line 462: @claude Simplify this warning message. No need to, e.g., mention the per-iteration retrim

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+    """Smooth phase-space track for a single arm of a stream.
+
+    A StreamTrack holds a smooth mean curve (and optional covariance) in
+    galactocentric Cartesian phase space, parameterized by an affine curve
+    parameter ``tp``. The class is a pure precomputed-track container —
+    accessors evaluate the stored splines and ``cov(basis=...)`` transforms
+    the stored covariance into other bases via analytical Jacobians. The
+    fitting pipeline (closest-point + offset smoothing) is provided by the
+    :meth:`from_particles` classmethod and lives in
+    :func:`_fit_track_from_particles`.
```
jobovy: @claude No need to mention an internal function here

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            ``normalization.ro`` config value and physical-units output is
+            off by default (``_roSet=False``); when set explicitly,
+            physical-units output is on by default. Mirrors ``Orbit``.
+        vo : float or Quantity, optional
+            Circular velocity scale (km/s). Same semantics as ``ro``.
+        zo : float or Quantity, optional
+            Sun's height above the midplane (kpc). Default None.
+        solarmotion : str, numpy.ndarray or Quantity, optional
+            ``'hogg'``, ``'dehnen'``, ``'schoenrich'``, or ``[-U, V, W]``
+            in km/s. Default None.
```
jobovy, line 627: @claude I think None here means falling back onto the progenitor orbit's values for these? If so, mention this

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+                "matrix (3x3) from (ra, dec) to the custom sky frame. "
+                "See galpy.util.coords.align_to_orbit for a helper that "
+                "builds one from a progenitor Orbit."
+            )
+
+    def _radec_internal(self, tp):
+        """ra, dec along the track in degrees (no unit attached)."""
+        X, Y, Z, _, _, _ = self._helio_xv(tp)
+        lbd = coords.XYZ_to_lbd(X, Y, Z, degree=True)
+        return coords.lb_to_radec(lbd[:, 0], lbd[:, 1], degree=True)
```
jobovy, line 1508: @claude Better to also use this for the actual ra and dec accessors

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+    def _pmrapmdec_internal(self, tp):
+        """pmra, pmdec along the track in mas/yr (no unit attached)."""
+        lbd, vrpmllpmbb = self._vrpmllpmbb(tp)
+        return coords.pmllpmbb_to_pmrapmdec(
+            vrpmllpmbb[:, 1],
+            vrpmllpmbb[:, 2],
+            lbd[:, 0],
+            lbd[:, 1],
+            degree=True,
+        )
```
jobovy, line 1519: @claude also better to use this for the actual pmra and pmdec accessors

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            T=self._custom_transform,
+            degree=True,
+        )
+        return self._maybe_scalar(tp, pm12[:, 0])
+
+    @physical_conversion("proper-motion_masyr", pop=True)
+    def pmphi2(self, tp, **kwargs):
+        """Proper motion in custom-frame ``phi2`` along the track.
+
+        Requires ``custom_transform`` to have been set at construction.
```
jobovy: @claude should we allow per-method-use overrides of custom_transform or updating the attribute? And if we do, document clearly in API and tutorial!

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+        J_cs = J_cs.copy()
+        J_cs[0] *= 1.0 / coords._DEGTORAD
+        J_cs[1] *= 1.0 / coords._DEGTORAD
+        return J_cs
+
+    # -----------------------------------------------------------------
+    # Unit toggles
+    # -----------------------------------------------------------------
+    def turn_physical_on(self, ro=None, vo=None):
+        """Turn on physical-units output. Mirrors ``Orbit.turn_physical_on``.
```
jobovy: @claude don't add the "Mirrors" sentence

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            Passed to matplotlib.pyplot.plot.
+
+        Notes
+        -----
+        Uses the same physical-unit logic as ``Orbit.plot``: physical
+        output is on iff ``_roSet`` and ``_voSet`` are True (or the user
+        passes ``use_physical=True``/explicit ``ro=`` / ``vo=``); axis
+        labels are picked accordingly.
+        """
+        from matplotlib import pyplot
```
jobovy: @claude Again, move imports to top of file unless problematic

galpy/df/streamTrack.py
```diff
@@ -0,0 +1,2059 @@
...
+            Coordinate names. Any of: x, y, z, vx, vy, vz, R, vR, vT, phi,
+            ra, dec, dist, ll, bb, pmra, pmdec, pmll, pmbb, vlos, plus
+            phi1, phi2, pmphi1, pmphi2 if ``custom_transform`` is set.
+        spread : float, optional
+            If > 0, draw a ±spread·sigma band around the track along
+            ``d2`` using the projected covariance. Works for any ``d2``
+            that has a basis in :attr:`_COORD_BASIS`. Silently skipped
+            for axes outside the dispatch (e.g. ``ll``, ``bb``).
+        n : int, optional
+            Number of evaluation points (default: self._ninterp).
```
jobovy: @claude I don't think this needs to be exposed, just use ``self._ninterp``

galpy/df/streamTrack.py
```diff
@@ -17,9 +17,7 @@ def _particles_to_cartesian(xv_particles):
     """Convert particles from (R,vR,vT,z,vz,phi) to galactocentric
     Cartesian 6-vectors. Returns array of shape (N, 6)."""
     R, vR, vT, z, vz, phi = xv_particles
-    x_p, y_p, z_p = coords.cyl_to_rect(R, phi, z)
-    vx_p, vy_p, vz_p = coords.cyl_to_rect_vec(vR, vT, vz, phi)
-    return numpy.column_stack([x_p, y_p, z_p, vx_p, vy_p, vz_p])
+    return coords.galcencyl_to_galcenrect(R, vR, vT, z, vz, phi)
```
jobovy: @claude Can't we just get rid of this function and directly use coords.galcencyl_to_galcenrect(R, vR, vT, z, vz, phi) where necessary?

galpy/df/streamTrack.py
```diff
@@ -1478,14 +1478,28 @@ def vlos(self, tp, **kwargs):
...
+        """3x3 rotation matrix from (ra, dec) to the custom sky frame, or
+        None. Settable post-construction: assigning a new matrix updates
+        the frame used by the ``phi1``/``phi2``/``pmphi1``/``pmphi2``
+        accessors and by ``plot``/``cov`` in the ``customsky`` basis."""
+        return self._custom_transform
+
+    @custom_transform.setter
+    def custom_transform(self, T):
+        self._custom_transform = None if T is None else numpy.asarray(T, dtype=float)
+
```
jobovy, line 1491: @claude Add an example of this to the tutorial notebook and make sure to re-run the notebook properly before committing it

PR #949, opened 2026-06-12
Backend f32 policy: table-backed potentials cast to input dtype at exit (f64 interior preserved)

jobovy:
Implements the agreed float32 policy for the table-backed backend potentials — **keep the deliberately-float64 interiors** (SCF expansion tables, Ogata quadrature nodes/weights, spline coefficients, Multipole tables), **cast the result to the input dtype at method exit**: f32 in → f32 out at f64 interior quality.

- New `galpy.backend.match_input_dtype(out, *coords)` helper, applied at the exits of SCF (+ the SphericalHarmonic force/2nd-deriv mixin shared with Multipole), DoubleExponentialDisk, interpSpherical (King and DiskSCF/DiskMultipole inherit), and MultipoleExpansion. Strict no-op — same object back — for the f64 numpy path; plain-float scalar outputs keep their type unless *all* coords are narrower.
- **The gate caught a fifth instance while being built**: the `galpy.backend.special` router *fallbacks* carry f64 Gauss–Legendre/trapezoid node tables that promoted f32 (surfaced through the freshly-routed generic TwoPowerSpherical) — same exit cast applied in `_dispatch`.
- New `tests/test_backend_dtype_policy.py`: a 46-entry registry gate over every migrated potential (torch f32→f32 and f64→f64, **zero skips** — the two in-flight fixes landed mid-work and are folded in), f64-quality pins for the table-backed five (an input-anchored implementation would fail them), numpy no-op/cast semantics, AD-through-the-cast smoke tests (jax + torch).
- **numpy byte-identity**: 92-record `sha256(tobytes())` proof (all instances × methods × {f64 grids, python scalars, np.float64 scalars}) identical pre/post; independently re-verified by a second harness with 324 strict records incl. Quantity call paths.
- **Approved semantic change**: numpy users passing all-float32 arrays/scalars to these potentials now get float32 back (computed at f64 quality); previously float64.

Known residual (documented, not a regression): f32 deltas vs cast-f64 reach tens of f32-ulp where chain-rule prefactors run on f32 coordinates before the f64 interior — input-stage rounding, not table anchoring. Separately surfaced for follow-up: CUDA device preservation is broken for the table-backed potentials (+SpiralArms) — stored tables materialize on CPU; needs a device-anchoring pass validated on a GPU box.

### DIFF
--- a/galpy/backend/__init__.py
+++ b/galpy/backend/__init__.py
@@ -7,7 +7,12 @@
 #   of the array arguments), with an explicit ``xp=`` override and a
 #   context-manager/global default as fallbacks. See ``_resolver`` for details.
 ###############################################################################
-from ._namespaces import is_backend_array, match_input_dtype
+from ._namespaces import (
+    asarray_on_device,
+    device_of,
+    is_backend_array,
+    match_input_dtype,
+)
 from ._resolver import (
     _seed_from_config,
     backend,
@@ -26,4 +31,6 @@ __all__ = [
     "set_default_backend",
     "is_backend_array",
     "match_input_dtype",
+    "device_of",
+    "asarray_on_device",
 ]
--- a/galpy/backend/__init__.py
+++ b/galpy/backend/__init__.py
@@ -7,7 +7,7 @@
 #   of the array arguments), with an explicit ``xp=`` override and a
 #   context-manager/global default as fallbacks. See ``_resolver`` for details.
 ###############################################################################
-from ._namespaces import is_backend_array
+from ._namespaces import is_backend_array, match_input_dtype
 from ._resolver import (
     _seed_from_config,
     backend,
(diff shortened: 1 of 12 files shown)


jobovy:
Added the device-anchoring commit (12000d06), per review: stored tables (and the special-router fallbacks' bessel/hyp2f1 node tables, which also failed on CUDA) now anchor on the input device via new `device_of`/`asarray_on_device` helpers — numpy path untouched (device=None ⇒ kwarg omitted; 24-record bitwise spot-proof). Validated live on a TITAN Xp: 9 potentials × methods × {f32,f64} × {scalar,array} all on cuda:0 with input dtype, f64 CUDA↔CPU ≤7e-15, autograd on-device; jax CudaDevice + jit-traced paths too. 33 new CUDA tests skip visibly on CPU CI. One pre-existing out-of-scope finding for the cleanup PR: SpiralArms under jax promotes f32→f64 (the dtype gate is torch-only).

PR #960, opened 2026-06-15
backend: central coordinate coercion — fix systemic torch scalar-input rejection

jobovy:
Fixes torch's dominant failure mode under the backend test harness (~1670 torch
xfail-ledger entries): when the namespace resolves to torch (forced harness, or a
user mixing a torch tensor with a numpy/python arg) but a coordinate is still
numpy/python, torch rejects it strictly — `torch.sqrt(numpy.float64)` and
`numpy.ndarray * Tensor` raise. jax is lenient, so jax mostly escaped.

## The fix (one central boundary)
`potential_physical_input` (`galpy/util/conversion.py`) is the **outermost
decorator carried by every coordinate-taking public method** of `Potential` /
`planarPotential` / `linearPotential` / `DissipativeForce` **and** every
module-level `evaluate*` function — all potentials inherit it. It already
enumerates the coordinate args for unit conversion, so coercing them onto the
active backend there fixes **~39 potentials at once** instead of a per-file
whack-a-mole.

- `galpy/backend/_namespaces.py`: new `coerce_coords` — numpy is a strict
  pass-through (byte-identical); float-dtype coords keep their dtype (so the
  float32/`match_input_dtype` policy holds); python scalars / int arrays lift to
  the backend's **float64** (galpy's interior precision — a bare `asarray` of a
  python float gives torch float32 and misses the tolerances).
- `conversion.py`: coerce positional `R/z/x` + `phi/t/R/z/x/v` kwargs; control
  kwargs (`dR/dphi/...`) untouched (derivative flags stay ints).

## Validated
- **numpy byte-identical** (gated no-op + spot check).
- **torch `test_potential.py`: 140 passing** (this was the dominant failure); mixed
  (torch tensor + python float) and numpy-array inputs now work, at float64
  matching numpy exactly (d=0).
- **jax untouched.**

## Scope
The ~75 remaining torch `test_potential` failures are a per-potential long tail
(internal numpy-array×tensor mixing + a handful of `coords.py` bypasses) — normal
migration work, not this systemic issue. **`coords.py`** (the df/actionAngle/orbit
route) is the natural next central boundary, a separate follow-up.

Ledger un-defer of the now-passing torch tests follows via the scheduled regen.


jobovy:
**Draft — CI red, real design issue found.** The coerce works for potential
compute (numpy byte-identical; torch `test_potential` +140; jax unaffected), but
`potential_physical_input` is too broad: it's also carried by utilities/df/
actionAngle functions (`estimateDeltaStaeckel`, `diskdf`, `estimateBIsochrone`,
`interp_potential`, …) that do **scalar** operations on the coords. Coercing
python scalars to 0-d backend arrays at this public boundary breaks them:
`ValueError: truth value of an array … ambiguous`, `float() … not 'coords…'`,
`len() of unsized object`, scalar-Quantity indexing.

**Fix direction:** inject the coercion at the **private compute layer** (the
potential `_evaluate`/`_Rforce`/… or the shared `*_nodecorator` dispatch points),
per the original plan — *not* the public `potential_physical_input` decorator,
which is one level too high and hits non-compute consumers. Redesign pending.

(Separately surfaced: a pre-existing latent bug — `coords.py:_promote` passes a
numpy-2.0 array's `.device='cpu'` *string* to `jnp.asarray`, which wants a
Sharding/Device; it uses `getattr(ref,'device')` directly instead of the safe
`device_of`. Independent of this PR, ledgered today.)

jobovy:
**Ready for review — green and undrafted.**

All real checks pass: every `backend jax`/`backend torch` shard, the full `build` matrix, and **codecov patch + project both green**. The only red checks are the `run-notebooks` jobs, which are the known VizieR/conda-CDN environmental flakes hitting every PR right now (not related to this change).

**What this does:** central coordinate coercion at the `potential_physical_input` boundary, gated on `_check_backend_compatible(Pot)`, so migrated potentials accept torch/jax inputs (incl. mixed scalar/array) instead of torch's strict scalar-rejection. numpy path is a strict pass-through → **byte-identical** (the coerce branch is skipped when `xp is numpy`).

**Ledger note (expected "ascent"):** because the coercion newly *exercises* previously-numpy-inert paths under forced-jax/torch, several still-unmigrated slow paths surface as 300s timeouts and were added to `backend_slow_skip.txt` (they test functionality not yet backend-ready, so they're deferred, not failing):
- FDM friction (`central_limit`, jax+torch)
- spherical-DF eddington (×2, jax)
- interpRZ interpolated density (jax)
- Chandrasekhar/MovingObject/RZTo-conversion `test_potential` tests (×5, jax) — dissipative forces (Track B group-b) + Orbit (Track D)

numpy still exercises every one of these. After this merges I'll do the milestone ledger review (this is burndown milestone #1 — the torch count should drop substantially) and retarget the stacked #962/#963 onto the updated `feat/backends`.

PR #970, opened 2026-06-16
EllipsoidalPotential: support scalar-R + array-phi (broadcast the cache key & frame rotation)

jobovy:
## The bug

Evaluating an `EllipsoidalPotential` subclass at **one position over many azimuths** —
`pot(1.5, 0.3, phi=numpy.array([...]))` — crashes:

```
ValueError: setting an array element with a sequence. The requested array has an
inhomogeneous shape after 1 dimensions.
```

This works on every *other* non-axisymmetric potential (LogarithmicHalo, TriaxialNFW,
DehnenBar, SoftenedNeedleBar, … all broadcast fine), but fails for **all** `EllipsoidalPotential`
subclasses — `PowerTriaxialPotential`, `PerfectEllipsoidPotential`, `TriaxialGaussianPotential`,
`TwoPowerTriaxialPotential`, etc. — on all three forces, all six second derivatives, and (for
rotated frames) the potential and density. It is pre-existing (not a regression).

### Cause

The force / 2nd-derivative md5 cache keys and the rotated-frame transform stack the
coordinates with `numpy.array([x, y, z])`. When a scalar `R, z` broadcasts against an array
`phi`, `x = R·cos(phi)` and `y = R·sin(phi)` become arrays while `z` stays a scalar, so the
stack is ragged and numpy raises.

## The fix

A small `_stack3(a, b, c)` helper that broadcasts the three coordinates before stacking
(`numpy.array(numpy.broadcast_arrays(a, b, c))`), used for the two cache keys and the three
frame rotations. The coordinates fed to the quadrature are untouched.

For inputs that already conform (all scalar, or all the same shape) broadcasting is a **no-op**,
so the stacked array — and every existing result — is unchanged.

## Verification

- **Byte-identical** for all existing-working inputs: 72 scalar + same-shape-array outputs
  (eval / 3 forces / 6 second derivatives / dens, aligned **and** rotated) match the pre-fix
  values exactly. An adversarial dtype sweep (float64/float32/int64/int32/complex128 + all
  mixed-dtype permutations) found zero differences in value, dtype, or md5 bytes.
- The previously-crashing **scalar-R + array-phi** case now returns correct results, matching
  an element-wise loop to machine precision (≤ 4.4e-16).
- The full `test_forceAsDeriv_potential`, `test_2ndDeriv_potential`, and `test_evaluateAndDerivs`
  loops over **every** potential pass (C library loaded).
- New regression test `test_EllipsoidalPotential_broadcast_scalarR_arrayphi` covers forces +
  all second derivatives (aligned) and forces/eval/dens (rotated) vs the element-wise loop.

### DIFF
--- a/galpy/potential/EllipsoidalPotential.py
+++ b/galpy/potential/EllipsoidalPotential.py
@@ -17,6 +17,17 @@ from ..util import _rotate_to_arbitrary_vector, conversion, coords
 from .Potential import Potential
 
 
+def _stack3(a, b, c):
+    """Stack three coordinate arrays into a homogeneous ``(3, ...)`` array.
+
+    ``numpy.array([a, b, c])`` raises on a broadcasting mix (e.g. a scalar R
+    with an array of azimuths gives array x,y but scalar z); broadcasting first
+    makes the stack -- used for the cache keys and the frame rotation -- accept
+    it. For already-conformable inputs (all scalar, or all the same shape) this
+    is a no-op, so the stacked array is identical to before."""
+    return numpy.array(numpy.broadcast_arrays(a, b, c))
+
+
 class EllipsoidalPotential(Potential):
     """Base class for potentials corresponding to density profiles that are stratified on ellipsoids:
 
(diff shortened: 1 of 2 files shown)


jobovy:
Closing per review discussion. The scalar-R + array-phi broadcast was never supported on the numpy backend, isn't tested or documented, and hasn't been reported — and galpy's existing precedent for these rotated-frame coordinate stacks (RotateAndTiltWrapperPotential) is to *reject* array inputs with a clear error, not silently broadcast them. So this would widen the supported input surface against precedent for a pattern nobody asked for, and the _stack3 helper doesn't generalize (the only other site, RotateAndTilt, deliberately rejects arrays). Not worth it. If the cryptic numpy error ever bites, the consistent fix would be a clear 'not supported' message rather than broadcast support.

PR #976, opened 2026-06-18
Add ExpTruncNFWPotential: exponentially-truncated NFW profile

sgpfaff:
## Summary

Adds `ExpTruncNFWPotential`, an NFW profile multiplied by an exponential
truncation `exp(-r/rc)` with truncation radius `rc`. The truncation gives the
profile a **finite total mass** while leaving the inner structure NFW-like.

All methods are closed-form:
- enclosed mass, potential, radial force, and second derivative via the
  exponential integral `E_1` (`scipy.special.exp1`);
- a small-radius Taylor expansion avoids catastrophic cancellation when
  `r << a, rc`.

`amp` follows the `NFWPotential` mass-scale convention, so the two potentials
coincide as `rc -> infinity` for equal `amp` and `a`. Alternatively the total
mass can be set directly with the `mass=` keyword.

## Implementation notes

- Inherits from `SphericalPotential` (option (b) in the docs), implementing
  `_revaluate`, `_rforce`, `_r2deriv`, `_rdens`, plus `_mass`, `_ddensdr`,
  `_d2densdr2`.
- `_mass` is overridden so `mass(numpy.inf)` returns the finite total mass
  rather than `NaN`.
- **Python only** (`hasC=False`): no C implementation yet, so orbit
  integration with this potential uses the Python integrators. Happy to add a
  C version in a follow-up if preferred.

## Tests & docs

- Dedicated tests: total mass / `mass=` round-trip, reduction to NFW as
  `rc -> infinity`, and agreement of the small-`r` series with the closed form
  across the branch switch. The potential is also picked up by the generic
  potential test sweep.
- Docs page + toctree entry, and a `HISTORY.txt` changelog note.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -15,10 +15,12 @@ v1.12.0 (expected around 2026-07-01)
   ``r`` is much smaller than ``a`` and ``rc``. The ``amp`` follows the
   ``NFWPotential`` mass-scale convention (the two potentials coincide as
   ``rc -> infinity``); alternatively the total mass can be set directly with
-  the ``mass=`` keyword. Has a C implementation (``hasC=True``), including the
-  full 3D Hessian for the variational equations (``hasC_dxdv3d=True``), giving
-  fast orbit integration on par with the other closed-form spherical
-  potentials.
+  the ``mass=`` keyword. Also provides a ``from_nfw`` classmethod that truncates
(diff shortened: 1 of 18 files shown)


jobovy:
Thanks, I think this looks good now! I'll merge it soon, but currently the CI is failing for an unrelated reason because SIMBAD is down for some of the `from_name` tests. So I'll wait until that clears and then merge.

PR #1023, opened 2026-06-27
c_stm_forward: stacked single integrate_dxdv for all STM columns (perf)

jobovy:
**P0-3 from the backend speed-up roadmap.** The differentiable fast-C orbit path (`galpy.backend._reference.inbackend_stm.c_stm_forward`, wrapped by the jax/torch C-STM autodiff rules) built the state-transition matrix `M(t)=∂x(t)/∂x(0)` by calling `Orbit.integrate_dxdv` **6N times** — once per canonical basis deviation vector per IC — each constructing a fresh `Orbit` and entering the C integrator separately.

This stacks all 6 basis vectors of all N ICs into **one** `integrate_dxdv` on 6N orbits (`numpy.repeat(ics, 6)` + `numpy.tile(eye(6), (N,1))`), keeping `rectIn=False, rectOut=False` so `M` stays in Orbit order, then reshapes `getOrbit_dxdv()`/`getOrbit()` back to `(N,nt,6,6)` / `(N,nt,6)`.

### Correctness
**Bit-identical** `x(t)` and `M(t)` to the per-column loop: 24/24 cases exact (`max|Δx|=max|ΔM|=0`) across MW2014 / triaxial-Log / MiyamotoNagai × rk4_c/rk6_c/dopr54_c/dop853_c × single+batch. The full C-STM/integrate suite (`tests/test_backend_orbit_stm.py` + `tests/test_backend_orbit_integrate.py`) passes — **117 tests** under numpy/jax/torch.

### Performance (N=20 ICs, dop853_c, 101 output times)
| | time |
|---|---|
| old (6N separate calls) | 881 ms |
| new (one call, `numcores`=1) | 750 ms (~1.2×, call-overhead) |
| new (one call, `numcores`=8) | 480 ms (~1.8×, parallel) |

The win is (a) collapsing 6N `Orbit` constructions + C entries into one and (b) letting the existing multi-orbit `numcores` parallelism span the 6N independent integrations (the per-call loop integrated one orbit at a time, so it couldn't). It uses galpy's default `numcores`, so this is automatic.

### Not in scope (noted for the optimization phase)
This does **not** eliminate the redundant base-orbit integration (each of the 6 stacked copies re-integrates the same base trajectory) — that is inherent to `integrate_dxdv`'s base+1-deviation design. The true ~6× would require a base+6-deviation **augmented** (42-state) C variational integrator; tracked as a follow-up.

### DIFF
--- a/galpy/backend/_reference/inbackend_stm.py
+++ b/galpy/backend/_reference/inbackend_stm.py
@@ -54,30 +54,28 @@ def c_stm_forward(pot, vxvv, ts, method, rtol, atol):
     vxvv = numpy.asarray(vxvv, dtype=numpy.float64)
     single = vxvv.ndim == 1
     ics = vxvv[None] if single else vxvv
-    basis = numpy.eye(6)
-    xts, Ms = [], []
-    for ic in ics:
-        cols = []
-        base = None
-        for i in range(6):
-            o = Orbit(list(ic))
-            o.integrate_dxdv(
-                basis[i],
-                ts,
-                pot,
-                method=method,
-                rectIn=False,
-                rectOut=False,
-                rtol=rtol,
-                atol=atol,
-            )
-            cols.append(o.getOrbit_dxdv())  # (nt,6): column i of M
-            if base is None:
-                base = o.getOrbit()  # (nt,6): the base orbit, Orbit order
-        xts.append(base)
-        Ms.append(numpy.asarray(cols).transpose(1, 2, 0))  # (nt,6,6)
-    xt = numpy.asarray(xts)
-    M = numpy.asarray(Ms)
+    n = ics.shape[0]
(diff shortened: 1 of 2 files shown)


jobovy:
Holding off per discussion. The batched `c_stm_forward` (one stacked `integrate_dxdv` for all 6N STM columns) is bit-identical and ~1.2× (call overhead) to ~1.8× (numcores across the 6 deviation columns) faster on the single-orbit C-STM gradient — but that's marginal standalone, and the real wins are the augmented 42-state STM integrator (true ~6×, base integrated once) and the multi-orbit C-STM lift. Folding this batching into that optimization work, where it compounds, rather than spending CI on a small standalone PR. The verified work is preserved on `feat/cstm-batch`.

PR #1027, opened 2026-06-28
inbackend ODE: batch N orbits in one diffrax/torchdiffeq solve (#99)

jobovy:
## What

The in-backend differentiable ODE path (`Orbit.integrate(..., method='diffrax'/'torchdiffeq')`)
was **single-orbit only** — a multi-orbit Orbit raised. This makes it a **native batch**:
a leading `N` axis is threaded through the rectangular EOM on the **trailing component
axis**, so N orbits integrate in **one adaptive solve** (one shared `saveat` grid + one
controller), mirroring the C-STM multi-orbit lift (#1024).

This is the diffrax/torchdiffeq analogue of #1025 (which batched IsochroneApprox over the
C-STM path), and it unlocks **IsochroneApprox-on-GPU + higher-order AA** (#102).

## How (native batch axis, not outer vmap)

- `_eom_rhs` / `_to_eom` index components on the trailing axis (`y[..., k]`,
  `vxvv[..., i]`) and stack on `axis=-1`; each backend `field` stacks `axis=-1` too. An
  explicit `dim` (2 or 6) replaces `len(y)` for the 1D-vs-3D dispatch (`len` is the batch
  size once a leading axis is present). A single `(phasedim,)` IC is the no-leading-axis
  special case → **identical code, identical single-orbit results**.
- `_integrate_inbackend`: drop the single-orbit guard; for a batch flatten the raw
  `_ic_backend` to `(size, phasedim)` (as the numpy/C-STM paths do), one solve returns
  `(nt, size, phasedim)`, `moveaxis` → canonical `(size, nt, phasedim)`.
- Native batch is chosen over `vmap(diffeqsolve)` because torchdiffeq has no native
  `vmap` (would force a Python loop) and `jax.vmap(diffeqsolve)` re-traces; native batch
  is symmetric across both backends and keeps the one-shared-`ts` Orbit contract.

## numpy byte-identity

numpy's in-backend branch still raises `NotImplementedError`; only backend-array ICs reach
the changed code → numpy path **untouched**. (No numpy-path edits to flag.)

## Correctness (verified, jax + torch)

- Batch & `(2,3)`-grid `getOrbit()` match per-orbit single solves to **1.2e-10** (the
  shared-controller coupling is far sub-tolerance).
- Gradients exact: `jacrev` off-diagonal (orbit *i* w.r.t. orbit *j≠i*) **exactly 0.0**;
  diagonal matches the single-orbit grad to 1e-11.
- **119** (`test_backend_orbit_integrate` + `test_backend_inbackend_ode`) + **62**
  (`test_backend_orbit_accessors` + new) backend tests green.

## Speed

The batched solve cost is **~flat in N** (one adaptive controller; force evals broadcast),
so on CPU N=8 jax is **~8.4×** faster than the per-orbit loop (**1.67 s vs 13.9 s**),
growing with N. The largest win is on GPU (one batched kernel vs N launches) — a clean GPU
number is a Track-G follow-up (the dev box's TITAN Xp / sm_61 is impractical for diffrax
timing).

## Limitations (documented + tested)

- A batch shares **one `max_steps`** and one error norm — a much stiffer orbit can exhaust
  `max_steps` for the whole batch where it would succeed singly (docstring warns; raise
  `max_steps` or split it off).
- **Per-orbit time arrays** (2-D `t`) raise a clear `NotImplementedError` (the batch shares
  one `saveat` grid).
- **Off-grid time accessors** (`o.R(t_offgrid)`, `o(t)`) on a multi-orbit backend Orbit raise
  a clear `NotImplementedError` instead of an opaque reshape crash (the interp is
  single-orbit; this also fixes the pre-existing C-STM batch path from #1024). Use
  `getOrbit()`.

Consumes the #1024 batching convention; out of scope: `_integrate_cstm` (already batched by
#1024). Found by an adversarial review of the diff (both review findings folded in).

### DIFF
--- a/galpy/backend/_jax/orbit_ode.py
+++ b/galpy/backend/_jax/orbit_ode.py
@@ -7,17 +7,21 @@
 ###############################################################################
 
 
-def integrate(pot, y0, ts, *, rtol, atol, max_steps):
+def integrate(pot, y0, ts, *, dim, rtol, atol, max_steps):
     """Integrate the EOM with diffrax (Dopri8, adaptive). y0/ys in rectangular
-    EOM variables [x, vx, y, vy, z, vz]. Reverse-mode differentiable (diffrax uses
-    a custom_vjp -> forward-mode jacfwd is unavailable; use jacrev)."""
+    EOM variables [x, vx, y, vy, z, vz], shape (dim,) for one orbit or (N, dim) for
+    a batch (integrated in one solve, shared adaptive controller -> ys (nt, N, dim)).
+    Reverse-mode differentiable (diffrax uses a custom_vjp -> forward-mode jacfwd is
+    unavailable; use jacrev)."""
     import diffrax
     import jax.numpy as jnp
 
     from .._reference.inbackend_ode import _eom_rhs
 
     def field(t, y, args):
-        return jnp.stack(_eom_rhs(y, pot, t, jnp))
+        # stack on the trailing (component) axis so a batch (N, dim) state maps to
+        # an (N, dim) derivative; for a single (dim,) state axis=-1 == axis=0.
+        return jnp.stack(_eom_rhs(y, pot, t, jnp, dim), axis=-1)
 
     sol = diffrax.diffeqsolve(
         diffrax.ODETerm(field),
--- a/galpy/backend/_jax/orbit_ode.py
+++ b/galpy/backend/_jax/orbit_ode.py
@@ -10,28 +10,39 @@
 def integrate(pot, y0, ts, *, dim, rtol, atol, max_steps):
     """Integrate the EOM with diffrax (Dopri8, adaptive). y0/ys in rectangular
     EOM variables [x, vx, y, vy, z, vz], shape (dim,) for one orbit or (N, dim) for
-    a batch (integrated in one solve, shared adaptive controller -> ys (nt, N, dim)).
+    a batch.
+
+    ``ts`` shape (nt,): one shared output grid -> all N orbits in ONE solve (one
+    shared adaptive controller). ``ts`` shape (N, nt): a PER-ORBIT grid -> jax.vmap
+    over (y0, ts) so each orbit gets its own saveat/span and its own (independent)
+    controller. Either way returns ys (nt, N, dim) for a batch ((nt, dim) single).
     Reverse-mode differentiable (diffrax uses a custom_vjp -> forward-mode jacfwd is
     unavailable; use jacrev)."""
     import diffrax
+    import jax
     import jax.numpy as jnp
 
     from .._reference.inbackend_ode import _eom_rhs
 
-    def field(t, y, args):
-        # stack on the trailing (component) axis so a batch (N, dim) state maps to
-        # an (N, dim) derivative; for a single (dim,) state axis=-1 == axis=0.
-        return jnp.stack(_eom_rhs(y, pot, t, jnp, dim), axis=-1)
-
-    sol = diffrax.diffeqsolve(
-        diffrax.ODETerm(field),
-        diffrax.Dopri8(),
-        t0=ts[0],
-        t1=ts[-1],
-        dt0=None,
-        y0=y0,
-        saveat=diffrax.SaveAt(ts=ts),
-        stepsize_controller=diffrax.PIDController(rtol=rtol, atol=atol),
-        max_steps=max_steps,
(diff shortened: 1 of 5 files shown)


jobovy:
**Update (`196363eb`): off-grid time evaluation now works for multi-orbit batches** (per review feedback — was a clean `NotImplementedError` in the first commit).

`_call_internal_backend_interp` now handles a batched in-backend Orbit: since the batch is one solve, every orbit shares the same integration grid, so each orbit is splined on that shared grid (the in-backend differentiable cubic) and the per-orbit results are stacked on the orbit axis → `(phasedim, nt, size)` (same layout as the on-grid batch branch). A single orbit (`size==1`) reproduces the prior `(phasedim, nt, 1)` byte-identically.

So `o.R(t)`, `o.x(t)`, `o(t)`, `o.E(t)`, … at **off-grid** times now work for a batched Orbit (differentiable w.r.t. ICs / params / eval times), returning `(N, nt)` for array `t` and `(N,)` for scalar. Verified: 6 accessors match the per-orbit single solve to 1e-6; `o(t)` → `(N,)`-shape Orbit; **67** backend tests green (single-orbit accessors byte-identical + new multi-orbit off-grid). The off-grid fix also covers the C-STM batch path (#1024).

The "off-grid accessors raise" limitation in the PR description above is superseded by this commit.

jobovy:
**Update (`82a00cdb`): per-orbit time arrays now supported** (per review feedback — the C integrators allow each orbit its own time array of the same length via `indiv_t`, used by `streamspraydf`).

- `ts` shape `(nt,)` → one shared grid, all N orbits in **one native-batch solve** (the fast #99 path, unchanged).
- `ts` shape `self.shape + (nt,)` → a **per-orbit grid**: each orbit integrates on its own times/span with its own independent adaptive controller. **jax** uses `jax.vmap` over `(y0, ts)` (`out_axes=1`); **torch** loops `odeint` per orbit and stacks (torchdiffeq has no `vmap`). Either way → `(size, nt, phasedim)`, so `getOrbit()` is correct and per-orbit `t0`/`t1` are both honored (the streamspraydf pattern).

`_integrate_inbackend` drops the per-orbit-`t` `NotImplementedError`, validates `ts.shape[:-1] == self.shape` (clear `ValueError` otherwise), and flattens to `(size, nt)`. `self.t` is stored 2-D so accessors route to the existing `_call_internal_indiv_t`. `jax.vmap` is vectorization (not `jit`), so it composes with an outer user `jit`. numpy untouched.

**Verified (jax + torch):** per-orbit batch & `(2,2)`-grid `getOrbit()` match the per-orbit single solves to ~3e-11 (different `t0`/`t1` per orbit); shared 1-D `ts` unchanged; gradient off-diagonal exactly 0.0 (vmap independence); shape-mismatch raises; concrete per-orbit accessors work. **127** backend tests green.

This supersedes the per-orbit-`t` limitation noted earlier in the thread.

jobovy:
**Update (`d9fa833e`): per-orbit accessors made backend-safe + differentiable (self-review fixes).**

An adversarial review of the per-orbit commit caught that the **accessor** path (a per-orbit Orbit reaches `_call_internal_indiv_t`, `self.t` being 2-D) wasn't backend-safe — my `getOrbit()`-only tests missed it. Fixed 4 issues:
- **(blocker)** the per-orbit exact-grid fast path did `self.orbit.transpose(2,1,0).copy()` — torch has no 3-arg transpose/`.copy()`, so every concrete accessor on a per-orbit **torch** Orbit crashed. Now dispatches on `is_backend_array` → `_backend_safe_copy(xp.permute_dims(...))`.
- **(major)** `self.t` stored unflattened for grid-shaped per-orbit Orbits → off-grid accessor crash. Now flattened to `(size,nt)` (mirroring the numpy `indiv_t` path).
- **(minor)** per-orbit off-grid accessors fell through to scipy (numpy, non-differentiable; crash on a traced orbit). New `_indiv_t_backend_interp` splines each orbit **in-backend** on its own grid (differentiable, on-backend) — the per-orbit-grid analogue of `_call_internal_backend_interp` (shared `_backend_spline_orbit` helper).
- **(test gap)** added `test_integrate_multiorbit_inbackend_perorbit_t_accessors` (jax + torch) covering exact-grid **and** off-grid queries, asserting backend results match `getOrbit()` / the per-orbit single solve.

13 + 127 backend tests green; shared-grid off-grid path unchanged (refactor verified). Thanks — good catch surfacing per-orbit; the review found the accessor gap it created.

PR #1033, opened 2026-06-29
AA backend completion: forced-backend numpy-island query paths + potentials promotion

jobovy:
## Summary

Completes the action-angle (Track E) backend story so the **entire `test_actionAngle.py` suite runs cleanly under a forced numpy / jax / torch backend** — no unexpected failures, numpy byte-identical.

The arc (5 commits, linear off `feat/backends`):
1. **coords delta-promote + Staeckel C-dispatch routing** — backend arrays bypass the C path into the vectorised core.
2. **xp-ify** IsochroneApprox / Spherical / Adiabatic / Vertical / Orbits-interface paths.
3. **grid-build under `use("numpy", force=True)`** — StaeckelGrid / AdiabaticGrid / interpSphericalPotential build their scipy-spline tables under numpy regardless of the outer forced backend.
4. **potentials forced-backend scalar-input promotion** — `Pot(R,z)` / `_evaluatePotentials` / force funnels promote numpy/scalar inputs to the active backend (so a potential evaluated under a forced backend returns a backend array). Byte-identical on numpy.
5. **numpy-island the AA query paths** + fix two potential-side leaks the promotion exposed (this PR's final commit — see below).

## The numpy-island fix (commit 5)

Once potentials promote under a forced backend, the AA query paths that compute against an inherently-numpy core (vectorised Staeckel core; scipy-spline grids; Spherical scipy loop; Adiabatic numpy path) mixed a backend array into numpy ops (`ndarray (-|*) Tensor`). Fix: a shared **`galpy.backend.numpy_island`** decorator (+ a local `_island(xp)` helper in Staeckel) that forces numpy for the whole call **when no argument is a backend array**, so the numpy-input path stays byte-identical and returns numpy; a real backend-array input runs the backend-native path untouched. Mirrors the grid-build boundary.

Two potential-side leaks surfaced by the same promotion, also fixed here:
- **`planarForce`**: guard `phi=None` before `promote_scalars` (axisymmetric `vcirc`/`rl`/`epifreq`/`omegac` pass `phi=None`, which can't be coerced → `torch.asarray(None)`). Regression from commit 4.
- **`CylindricallySeparablePotentialWrapper`**: store the `Phi(Rp,0)` reference offset as a backend-agnostic Python scalar (was a construction-time-cached backend array → cross-backend mix).

## Results (forced backend; numpy byte-identical preserved)

- **torch** `test_actionAngle`: **0 failed, 61 xpassed (flips), 105 xfailed** (248s).
[...]

PR #1038, opened 2026-06-30
AA backends run for real: resolved-namespace dispatch (replaces #1033 numpy_island)

jobovy:
## What & why

Replaces the closed **#1033** approach. #1033 added `@numpy_island`, which made the
action-angle methods fall back to **numpy** whenever no input was already a backend
array — so under a forced backend (`pytest --backend=jax/torch`) the existing
`test_actionAngle.py` suite ran on **numpy**, not the backend ("hollow green"). That
defeats Pillar 1 (the *existing* suite must run on every backend). #1033 is closed; the
only salvageable, backend-agnostic fixes are folded in here.

### The redesign
Every AA dispatch site changes from an `is_backend_array(R)` gate to **resolved-namespace
dispatch**:

```python
xp = get_namespace(R, vR, vT, z, vz)          # honors use(backend, force=True)
if xp is not numpy:
    R, vR, ... = (xp.asarray(a) for a in (...))
    return self._<...>_backend(...)
```

`get_namespace(numpy_array)` is `numpy` by default (→ **byte-identical** numpy path) but the
forced backend under the conftest `--backend` fixture, so the existing tests now run the
**real** vectorized jax/torch path. All 8 backend-capable AA modules are converted:
Spherical, Vertical, Adiabatic, Harmonic, AdiabaticGrid, StaeckelGrid, IsochroneApprox, and
**Staeckel (c=False)** — the c=False path was already fully vectorized/backend-agnostic
(#1010/#1013/#1014/#1015), it just had the same hollow gate. Only the **c=True C forward**
(Staeckel/Adiabatic) stays ledgered until analytic derivatives land.

## Changes
- **8 AA modules**: resolved-namespace dispatch on every `_evaluate`/`_actionsFreqs`/
  `_actionsFreqsAngles`/setup site.
- **Adiabatic**: gate the backend path on `not use_c` so an explicit `c=True` is honored.
- **Vertical**: mask J/Omega/angle to the unbound sentinel in the backend path; the unbound
  **angle** sentinel is `(9999.99**2) % 2pi` to match numpy (which sets angle=9999.99 before
  `angle *= Omega` then `% 2pi`).
- **`_torch/optimize.py`**: `brentq` returns a clean detached forward unless grad is actually
  needed (the unconditional `requires_grad_(True)` broke `.numpy()` on forward-only calls).
- **`test_actionAngle.py`**: two helpers (`conserved_actions`, `linear_angles`) reduce
  backend-agnostically via `_to_numpy` (identity on numpy) so torch tensors survive
  `numpy.mean/amax/fabs` — the orbit-test pattern, not ledgering.
- **`backend_xfail.txt`**: honest AA prune from a full-file REGEN run per backend.

## Ledger: jax 99→68, torch 166→120
Regen ran the whole `test_actionAngle.py` per backend in REGEN mode (xfail off): **207
collected, 0 collection errors**. Verification:
- **torch ledger == full regen, exactly** (120 == 120, zero diff) against the final code.
- **jax**: regen + live re-check of the boundary tests (2 fail/ledgered, 10 pass/un-ledgered);
  a full jax confirmation run is in flight.
- `test_actionAngleVertical_unbound` was **fixed** (not ledgered).
- The single honest add, `test_physical_vertical`, is a **pre-existing** decorator quirk: a 1D
  `__call__` returns a 1-tuple `(val,)`; numpy broadcasts `tuple - ndarray` but jax/torch raise
  on `tuple - array`. Not introduced here; fixing needs a shared-decorator change (numpy
  byte-identity risk) → ledgered honestly.

Remaining ledgered AA: c=True C forward, the Inverse modules (not in the 8-module set, separate
redesign), and ~1e-7 torch precision gaps the real backend path now exposes
(linear_angles/kepler/against_orbit).

## Guarantees
- **numpy byte-identical**: the only library change off the backend branch is in
  `_actionsFreqsAngles_backend` (a branch numpy never executes); test reductions are identity
  on numpy.
- Non-AA ledger lines are byte-identical (only `test_actionAngle.py` entries + one orphaned
  comment changed).

### DIFF
--- a/galpy/actionAngle/actionAngleAdiabatic.py
+++ b/galpy/actionAngle/actionAngleAdiabatic.py
@@ -129,15 +129,17 @@ class actionAngleAdiabatic(actionAngle):
             z = numpy.array([z])
             vz = numpy.array([vz])
         xp = get_namespace(R, vR, vT, z, vz)
-        if xp is not numpy:
-            # Backend path: runs under a forced backend even for numpy inputs
-            # (promote first), exercising the backend for real, not the numpy core.
-            R, vR, vT, z, vz = (xp.asarray(a) for a in (R, vR, vT, z, vz))
-            return self._evaluate_backend(R, vR, vT, z, vz)
-        if (
+        use_c = (
             (self._c and not ("c" in kwargs and not kwargs["c"]))
             or (ext_loaded and ("c" in kwargs and kwargs["c"]))
-        ) and _check_c(self._pot):
+        ) and _check_c(self._pot)
+        if not use_c and xp is not numpy:
+            # Backend path: runs under a forced backend even for numpy inputs
+            # (promote first), exercising the backend for real -- unless the C
+            # path was explicitly requested (then fall through to it below).
+            R, vR, vT, z, vz = (xp.asarray(a) for a in (R, vR, vT, z, vz))
+            return self._evaluate_backend(R, vR, vT, z, vz)
+        if use_c:
             Lz = R * vT
             jr, jz, err = actionAngleAdiabatic_c.actionAngleAdiabatic_c(
                 self._pot, self._gamma, R, vR, vT, z, vz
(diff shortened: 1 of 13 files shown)


jobovy:
Folded in the coercion unification per review (cc1d51e0): the 7 split-structure AA modules now coerce inputs via the shared `promote_scalars` (galpy.backend._coerce) instead of inline `xp.asarray`, matching Harmonic/Isochrone and aligned with the potentials' `coerce_coords` family. This adds device-anchoring (a numpy scalar beside a GPU array now lands on the array's device) + the float64-scalar dtype policy, while genuine backend arrays pass through object-identically. Kept inside the `if xp is not numpy` gate (numpy byte-identical) and preserved the isinstance(float) shape preamble + Staeckel's device-aware phi/delta folds. Verified: `test_backend_actionAngle.py` 299 passed under jax and 299 under torch (parity + gradcheck).

PR #1049, opened 2026-07-02
actionAngleStaeckelGrid: true-action gradients via t²-donor graft (Track E deriv fix, PR B)

jobovy:
Follow-on to #1047. StaeckelGrid's backend gradients previously came from AD through the cubic interpolation — which differentiates the **interpolation error** along with the action: measured **12%–290×** off FD-of-the-direct-action on a MWPotential2014 grid (5–10% on dominant components, wrong sign on small ones). That's bug-level, not polish.

## Fix
Under the AD gate (`is_backend_array` + jax-trace/torch-grad), graft the t²-donor gradient of the **true** action (`_staeckel_prep` + `_staeckel_t2_action` at the query points, same args/prefactors as `_staeckel_jr_jz`) onto the untouched interpolated value, via the same `graft_gradient` construction as #1047.

- Post-fix: grafted AD ≤ **5.2e-4 (jr) / 2.7e-4 (jz)** of FD gold, jax ≡ torch — the direct-path accuracy.
- Forward value stays exactly the interpolated one (traced == untraced to 7e-16); numpy byte-identical (`tobytes`-equal, incl. scalar-loop and EccZmaxRperiRap); zero cost off the AD path.
- The donor arg-tuple duplication vs `_staeckel_jr_jz` is deliberate (a shared helper would touch `actionAngleStaeckel.py`, owned by #1049's branch when written; DRY follow-up possible after both land).

## Test-side
- New `tests/test_backend_staeckelgrid_grad.py` (per-module convention): jr/jz × 5 coordinates vs FD-of-direct on generic + vz=0/z=0 edge orbits × both backends + value-unchanged; 10 passed.
- `test_grid_jz_grad_vs_fd_wrt_vz` drops its `staeckel` case: it asserted AD == FD-of-the-interpolant — precisely the error this PR removes; the adiabatic case (still spline-AD) is unchanged.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelGrid.py
+++ b/galpy/actionAngle/actionAngleStaeckelGrid.py
@@ -17,6 +17,7 @@ from .. import potential
 from ..backend import get_namespace
 from ..backend import interpolate as backend_interpolate
 from ..backend import is_backend_array, promote_scalars
+from ..backend._namespaces import graft_gradient, under_jax_trace, under_torch_grad
 from ..potential.Potential import (
     _check_potential_list_and_deprecate,
     _evaluatePotentials,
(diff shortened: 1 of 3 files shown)


jobovy:
Closing unmerged after review discussion: the donor graft defeats the grid's purpose under AD (measured: grad-through-grid is 1.18× SLOWER than the direct path, vs 6× FASTER forward) and makes the gradient inconsistent with the interpolated value (false precision against a percent-level tool — two reasonable grid resolutions differ by 32% in value at a generic point). Design decision: grids keep self-consistent spline-AD gradients (the derivative of the surface actually evaluated, error commensurate with grid accuracy); accurate true gradients live on the direct (#1047) and c=True paths. Investigation + baseline measurements (spline-AD is 5–10% off the true action derivative on dominant components) documented in the galpy-jax lab notebook; a future opt-in 'gridded donor-gradient tables' idea is noted there if a fast true-gradient is ever needed.

PR #1050, opened 2026-07-02
actionAngleStaeckel: differentiable c=True actions — C value + donor gradient (Track E deriv fix, PR C)

jobovy:
Follow-on to #1047, completing the "not just in Python but also C" half: the fast **`c=True` path now accepts backend arrays and carries accurate action gradients**.

## Design
Backend-array inputs in the `c=True` branches of `_evaluate` / `_actionsFreqs` / `_actionsFreqsAngles` route to backend helpers:
- **Forward value from C** (unchanged numerics): torch / eager jax via detach/asarray round-trip; jax **under trace** via `jax.pure_callback` (new `galpy/backend/_jax/staeckel_c.py`, mirroring `orbit_stm.py`). One subtlety: the callback has no JVP rule, so the coords are `lax.stop_gradient`-ed *into* it — intended semantics, since the gradient comes from the graft, not the callback.
- **Gradient under AD** from the same t²-donor graft as #1047 (`_staeckel_prep` + `_staeckel_t2_action` in-backend). `Lz = R·vT` stays in-backend (no round-trip). Freqs/angles **values** pass through ungrafted — their gradients are Phase-2 second-order objects (see #1048's xfail baselines).

## Validation
- **c=True backend forward == numpy c=True exactly** (0.0 relative — same C computation), eager == jit.
- **numpy path byte-identical**: 72/72 arrays `tobytes`-equal vs pristine base — 10 orbits (incl. vR=0/vz=0/z=0 edges) × {actions, actionsFreqs, actionsFreqsAngles, EccZmaxRperiRap} × {MiyamotoNagai, MN+useu0, MWPotential2014} + delta-array + u0-kwarg.
- **Gradients**: 4.6e-4 relative of FD gold (the known donor-vs-plain-GL quadrature offset from #1047), both backends, through `_evaluate` and through `actionsFreqs`' action outputs; jit-stable through the pure_callback.
- Tests parametrized over `c ∈ (False, True)` + new c=True-specific parity/jit/useu0 tests → **56 passed**; numpy Staeckel suite 72/72.

`_uminumaxvmin` (EccZmax backend path) deliberately untouched here.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckel.py
+++ b/galpy/actionAngle/actionAngleStaeckel.py
@@ -325,6 +325,189 @@ def _staeckel_actions(xp, R, vR, vT, z, vz, pot, delta, order):
     return jr, s["Lz"], jz
 
 
+# ------------------------------------------------- c=True backend round-trip
+# jax/torch inputs on the c=True path: the forward VALUE comes from the
+# compiled C wrapper (eager detach->numpy->C->backend round-trip; under a jax
+# trace a jax.pure_callback so the user's jit/grad stays traceable) and, under
+# AD, the ACTIONS' gradient is grafted from the in-backend t^2-substituted
+# donor quadrature (same donor as _staeckel_jr_jz's c=False graft). Lz = R*vT
+# is computed in-backend (trivially differentiable). Frequency/angle values
+# pass through UNGRAFTED (their gradients are second-order objects; Phase 2).
+
+
+def _backend_to_numpy(x):
+    """Backend array -> numpy for the C wrapper (detach + off-device)."""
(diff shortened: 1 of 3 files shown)


jobovy:
Closing unmerged after review: redirecting to the C-native design — implement the necessary gradient functions in C and tie them in as gradients exactly like the C-STM (pure_callback custom_vjp / autograd.Function with a C-computed Jacobian residual), rather than sourcing the gradient from the Python donor. First-order only, accepted. The dispatch wiring, pure_callback skeleton, and c=False/True test parametrization from this branch will be reused in the reworked PR.

PR #1051, opened 2026-07-02
actionAngleStaeckel: C-native differentiable c=True actions — C Jacobian + custom_vjp (Track E deriv fix, PR C)

jobovy:
Reworked replacement for the closed #1050 (which sourced the gradient from the Python donor). Per the design decision, the `c=True` fast path's gradient is now computed **natively in C** and tied in exactly like the C-STM — **no Python fallback in the gradient path**. First-order only (accepted).

## Design
- **New C entry `actionAngleStaeckel_actionsJac`** assembles the full 2×5 Jacobian `d(jr,jz)/d(R,vR,vT,z,vz)` from the *existing* t²-substituted derivative integrands (`dJR/dJz` w.r.t. `E,Lz,I3`) plus the elementary chain through the reference geometry:
  - **Jr** closes with the effective `Ĩ3U = I3U − (sinh²u0 + sin²v0)·potu0v0`, which removes all explicit `u0` dependence → `dJr = djrdE·dE + djrdLz·dLz + djrdI3·dĨ3U`.
  - **Jz** is a genuine `F(E,Lz,I3V,u0)`: the assembly uses the *total* `dI3V` (incl. its `du0` term) plus a **new** `∂Jz/∂u0` integrand (`calcdJzdU0Staeckel`) — no double-counting.
  - Honors the `useu0`/`u0`-kwarg gate (`du0/dx = 0` for a fixed user `u0`).
- **Registered like C-STM**: `jax.pure_callback` + `custom_vjp` (new `galpy/backend/_jax/staeckel_c.py`) and `torch.autograd.Function` (`_torch/staeckel_c.py`); saved residual is the C Jacobian, backward is a matvec. Dispatch in the `c=True` branches of `_evaluate`/`_actionsFreqs`/`_actionsFreqsAngles` routes backend arrays to the vjp path (`Lz=R·vT` in-backend); numpy path untouched. Frequency/angle **values** pass through ungrafted — their gradients are second-order (Phase 2, see #1048's xfail baselines).

## Validation
- **Decisive cross-check**: C-Jacobian vs `jax.grad` through the merged c=False donor path (#1047) = **1.2e-12** (machine precision — two fully independent true-gradient mechanisms agree), across generic + vR=0/vz=0/z=0 edge orbits, both `jr` and `jz`. vs FD gold = 1.9e-4 (the documented plain-GL-vs-donor quadrature floor).
- **numpy `c=True` byte-identical** to base: `tobytes` 0.0 over 200 orbits × {actions, freqs, angles}, MiyamotoNagai + MWPotential2014.
- Tests parametrized over `c ∈ (False, True)`: **50 passed**; `test_actionAngle.py -k staeckel` **72 passed**; jit through `pure_callback` eager==jit.

## Scope note
For the rare combination of a *traced per-object* `delta`/`u0` under `jax.jit`, the reference-geometry `stop_gradient` isn't jit-traceable (scalar `delta` — the normal case — is fully jit-safe). Acceptable; revisit in Phase 2.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckel.py
+++ b/galpy/actionAngle/actionAngleStaeckel.py
@@ -589,6 +589,88 @@ def _staeckel_actions_freqs_angles(xp, R, vR, vT, z, vz, phi, pot, delta, order)
     return jr, s["Lz"], jz, Omegar, Omegaphi, Omegaz, angler, anglephi, anglez
 
 
+def _staeckel_c_grad_actions(pot, delta, R, vR, vT, z, vz, u0, order):
+    """Differentiable (jr, jz) via the C-native Staeckel action Jacobian.
+
+    For jax/torch inputs, wraps the compiled 2x5 d(jr,jz)/d(R,vR,vT,z,vz) C entry
+    (actionAngleStaeckel_actionsJac_c) in the backend custom_vjp / autograd.Function
+    (galpy.backend._{jax,torch}.staeckel_c): the forward is the plain round-trip C
+    action value; the backward is a matvec of the C-computed Jacobian. numpy inputs
+    never reach here. delta/u0 are fixed references (no gradient). First-order only."""
+    delta_np = numpy.atleast_1d(
+        numpy.asarray(stop_gradient(delta), dtype=numpy.float64)
+    )
+    u0_np = (
+        None if u0 is None else numpy.asarray(stop_gradient(u0), dtype=numpy.float64)
+    )
+
+    def host_jac(Rn, vRn, vTn, zn, vzn):
+        jr, jz, jac, err = actionAngleStaeckel_c.actionAngleStaeckel_actionsJac_c(
+            pot, delta_np, Rn, vRn, vTn, zn, vzn, u0=u0_np, order=order
+        )
+        return jr, jz, jac
+
+    name = getattr(get_namespace(R, vR, vT, z, vz), "__name__", "")
+    if "jax" in name:
+        from ..backend._jax.staeckel_c import actions_with_jac
+
+        return actions_with_jac(host_jac, (R, vR, vT, z, vz))
+    if "torch" in name:
+        from ..backend._torch.staeckel_c import actions_with_jac
+
+        return actions_with_jac(host_jac, R, vR, vT, z, vz)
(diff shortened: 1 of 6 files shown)


jobovy:
**Coverage-fix follow-up + one caveat for review.** Added tests for the c=True host-u0 / circular-freq branches (`333d66f8`). While doing so, surfaced a limitation worth flagging: with **`useu0=True`** (the niche numerical-stability option that computes a reference `u0=u0(E)` from the orbit), the C-native gradient holds that computed reference **fixed** under AD, so it omits the `dJ/du0·du0/dx` term — the `useu0` action gradient is ~20% off an FD that lets u0 vary (finite + same-order, but not exact). The **default** path (reference = the point's own `ux`) is unaffected and exact (1.2e-12 vs the donor), as is an explicit fixed `u0=` kwarg (genuinely du0/dx=0). Happy to either (a) leave it documented as a first-order limitation of `useu0`+AD, or (b) add the `du0/dx` chain term in C if you want `useu0` gradients exact too — let me know your preference.

jobovy:
**Resolved (`b9dc3541`): `useu0=True` gradients are now exact.** Added the `du0/dx` term per your call. The clean finding: **Jr is exactly u0-independent** (the `Ĩ3U` gauge cancellation gives `∂Jr/∂u0=0`, verified to 0), so only Jz needed it — and the C already had the two Jz u0-channels wired but dead (multiplied by `du0=0`). Activating them with the real `du0/dx = ∂u0/∂E·dE/dx + ∂u0/∂Lz·dLz/dx` (`∂u0/∂(E,Lz)` via implicit-diff of `calcu0`'s stationarity condition — forces-only, so no R2deriv-in-C requirement, matched to ~1e-10) closes it: `useu0` grad now matches FD-of-`useu0` to the quadrature floor (Jr 4.7e-4, Jz ≤5.3e-5), jax==torch bit-identical. The correction was large (up to ~60% of some dJz components). Regression guards: default reference=ux path still 1e-6 vs the donor, numpy c=True byte-identical (0.0). `useu0` is now a 3-way C mode (0=ux, 1=fixed kwarg, 2=calcu0). So all three c=True gradient configurations are now exact.

Issue #131, opened 2014-04-03
More clever time-stepping in orbit integration; add interpolation within integrator when output time step is smaller than adaptively necessary (see NR)


jobovy:

jobovy:
Closing this, because not a real bug/issue and moved to the [galpy extensions page](https://github.com/jobovy/galpy/wiki/Possible-galpy-extensions).

PR #1065, opened 2026-07-04
actionAngle: fix Staeckel EccZmaxRperiRap backend crash on default (c=True) objects (AA-milestone completeness)

jobovy:
Part of the action-angle milestone completeness sweep (the AA-completeness audit that precedes the ledger regen).

## The one real gap
`actionAngleStaeckel._uminumaxvmin` (backing `EccZmaxRperiRap` and the public turning-point query) lacked the `is_backend_array` dispatch its sibling compute methods (`_evaluate`/`_actionsFreqs`/`_actionsFreqsAngles`) all have. On a **default** object (`c=True` — the standard case whenever the C extension is loaded), `EccZmaxRperiRap` with jax/torch inputs entered the c=True branch and called the numpy/ctypes `actionAngleUminUmaxVminStaeckel_c` wrapper with backend arrays → `TypeError: Multiple namespaces...`. The existing tests missed it because they feed **numpy** arrays (byte-identical C path); only explicitly-`c=False` objects were exercised under a backend.

**Fix:** exclude backend arrays from the c=True guard so they fall through to the already-working vectorised `_staeckel_prep` path (no C-native uminUmaxVmin Jacobian exists, so no custom_vjp is needed). numpy stays on the C path — **byte-identical**. Verified jax/torch `EccZmax` + `_uminumaxvmin` now match numpy c=True to ~1e-14; the numpy Staeckel EccZmax + AdiabaticGrid tests are unchanged.

## Also in this sweep
- `actionAngleAdiabaticGrid.Jz`: the standalone (numpy scalar-only) accessor now guards on backend arrays and rides the migrated `_evaluate` path (numpy path untouched, byte-identical).
- `actionAngleIsochroneApprox._actionsFreqsAngles_backend`: corrected a stale `Axisymmetric only` docstring (the body implements the non-axi branches).

## Tests
`test_backend_actionAngle.py` adds `test_staeckel_default_c_ecczmax_backend` (EccZmax + `_uminumaxvmin` on a default c=True object, jax+torch vs numpy) and `test_adiabaticgrid_Jz_backend`.

Follow-ups (separate PRs): the coercion/device-anchoring mop-up (sphericaldf `Lz` numpy-bug + GPU-latent device anchoring) and the AA-milestone ledger regen.

### DIFF
--- a/galpy/actionAngle/actionAngleAdiabaticGrid.py
+++ b/galpy/actionAngle/actionAngleAdiabaticGrid.py
@@ -16,7 +16,7 @@ from scipy import interpolate
 from .. import potential
 from ..backend import get_namespace
 from ..backend import interpolate as backend_interpolate
-from ..backend import promote_scalars
+from ..backend import is_backend_array, promote_scalars
 from ..potential.Potential import (
     _check_potential_list_and_deprecate,
     _evaluatePotentials,
@@ -545,6 +545,10 @@ class actionAngleAdiabaticGrid(actionAngle):
         - 2012-07-30 - Written - Bovy (IAS@MPIA)
 
         """
+        if any(is_backend_array(a) for a in args):
+            # backend arrays: ride the migrated _evaluate path (numpy stays
+            # byte-identical below); the standalone grid lookup is numpy-only.
+            return self(*args, **kwargs)[2]
         self._parse_eval_args(*args)
         Phi = _evaluatePotentials(self._pot, self._eval_R, self._eval_z)
         Phio = _evaluatePotentials(self._pot, self._eval_R, 0.0)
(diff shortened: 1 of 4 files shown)


jobovy:
Closing per the plan: the EccZmaxRperiRap fix here routes c=True backend arrays to the c=False Python turning-point path — i.e. it makes EccZmax work by *avoiding* the C code. That's the opposite of 'differentiable when using C.' The follow-up C-native EccZmax PR (#131) will instead give c=True EccZmax/rperi/rap/zmax/e a C-native Jacobian via custom_vjp (like actions/freqs/angles), which both fixes the crash AND provides fast C-native gradients — reverting this PR's route-to-c=False. The AdiabaticGrid.Jz guard + IsochroneApprox docstring here will be redone properly (resolved-namespace) in the AA cluster redesign.

Issue #113, opened 2014-01-01
Perform actionAngle approximations as (x,v) -> (J,O,a) rather than (R,vR,...) -> (J,O,a)


jobovy:

jobovy:
Closing this, because not a real bug/issue and moved to the [galpy extensions page](https://github.com/jobovy/galpy/wiki/Possible-galpy-extensions).

PR #1066, opened 2026-07-04
potentials: leaf-guard get_namespace in MiyamotoNagai/Isochrone/TwoPowerSpherical (AA-milestone coercion mop-up)

jobovy:
Part of the action-angle milestone coercion mop-up (user directive: "fix all root causes"). One systematic root cause explains most AA forced-backend failures: the forced-backend test resolver makes `get_namespace()` return the forced backend **even when every argument is numpy**. Migrated potentials used a bare `xp = get_namespace(R, z)`, so a legitimate numpy-context caller (e.g. `estimateDeltaStaeckel → _evaluatezforces(pot, R[ii], z[ii])` with numpy scalars) got `xp=torch/jax` and applied a backend op to numpy data → `torch.sqrt(): must be Tensor, not numpy.float64` (torch) / jax `ArrayImpl` at the ctypes boundary (jax).

**Fix:** the established leaf-data-guard idiom (already used in `verticalPotential.py`):
```python
xp = get_namespace(R, z) if is_backend_array(R) or is_backend_array(z) else numpy
```
applied to every `get_namespace` site in the three files (7 + 8 + 39). numpy inputs stay numpy (**byte-identical** — `get_namespace(numpy…)` already returned numpy in non-forced mode, so this is a no-op there); genuine jax/torch inputs still get the backend namespace.

**Verified:** `_zforce` / internal `_evaluatezforces` return numpy under forced torch AND jax with numpy inputs (byte-identical to ref); `test_estimateDeltaStaeckel*` (5) flip to passing under both backends while numpy stays green.

Sibling PRs in the same mop-up: coords + Orbits.py leaf-guards (concurrent session); Staeckel `_useu0` numpy-island + IsochroneApprox `'ts'` fix (folded into the post-#1065 PR). The ledger prune follows once all land + a verified regen.

### DIFF
--- a/galpy/potential/IsochronePotential.py
+++ b/galpy/potential/IsochronePotential.py
@@ -7,7 +7,9 @@
 ###############################################################################
 import math
 
-from ..backend import get_namespace
+import numpy
+
+from ..backend import get_namespace, is_backend_array
 from ..util import conversion
 from .Potential import Potential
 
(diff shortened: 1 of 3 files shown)


jobovy:
Closing: this used the leaf-guard idiom (`... if is_backend_array(x) else numpy`), which keeps numpy inputs on numpy even under a forced backend — a temporary numpy island that passes the test WITHOUT exercising the backend. Per maintainer direction, leaf guards are being removed in favor of resolved-namespace + promote_scalars (coerce numpy inputs UP to the forced backend so the code genuinely runs on jax/torch), routing through the custom_vjp-wrapped C paths. The potential fixes will be redone with that pattern as part of the AA resolved-namespace redesign (#113). PoC proven on estimateDeltaStaeckel (runs on torch/jax, byte-identical numpy).

PR #1107, opened 2026-07-12
Check interpSphericalPotential range independently of FDM friction

steps-re:
In `Orbits.integrate()`, the post-integration diagnostic that warns when an orbit leaves an `interpSphericalPotential`'s radial interpolation range was chained as an `elif` off the `FDMDynamicalFrictionForce` minimum-radius check. So for a composite potential containing both, the FDM branch firing means the interpSphericalPotential range check is silently skipped and the user gets no warning that they are extrapolating outside a fitted potential.

Changed the `elif` to `if` so the two independent checks always run. This matches the very next block for `MultipoleExpansionPotential`, which already uses `if`. Reproduced with a composite of `LogarithmicHaloPotential + FDMDynamicalFrictionForce + interpSphericalPotential` and an orbit that exits the range: no warning before, expected `galpyWarning` after. test_FDMdynamfric, interpSpherical, and orbit warn/composite subsets pass.

### DIFF
--- a/galpy/orbit/Orbits.py
+++ b/galpy/orbit/Orbits.py
@@ -1899,7 +1899,7 @@ class Orbit:
                     """numerical instabilities)""",
                     galpyWarning,
                 )
-        elif (
+        if (
             isinstance(
                 pot,
                 (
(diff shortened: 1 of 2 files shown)


jobovy:
Thanks, but please add a simple test as well

steps-re:
Added a test covering the composite FDMDynamicalFrictionForce + interpSphericalPotential case. It fails on the previous elif (the range warning is silently skipped) and passes with the if fix.

jobovy:
Great, thanks!

PR #1155, opened 2026-07-19
streamdf Phase D: backend-native, differentiable inverse-CDF sampling (replaces ARS)

jobovy:
## Summary

streamdf **Phase D — sampling**: replace the adaptive-rejection sampling (ARS) frequency draw in `streamdf._sample_aAt` with a **spline inverse-CDF sampler** that is unified across numpy/jax/torch, differentiable (`jax.grad` / `torch.autograd`), and jit/GPU-able (no rejection loop, static shapes).

## What's new

- **`galpy/backend/sampling.py`** (new, reusable):
  - `spline_inverse_cdf_sample(xp, omega_grid, cdf_grid, u)` — builds the inverse spline (knots `x=cdf_grid`, values `y=omega_grid`) and returns `F⁻¹(u)`. The cubic coefficients are assembled **entirely in the namespace** (a static-index scatter into a zero matrix + `xp.linalg.solve`), so — unlike `interpolate.cubic_spline_coeffs`, which `numpy.asarray`'s its `x` and therefore **cannot be traced** (it raises `TracerArrayConversionError` under `jax.grad`/`jit`) — it is differentiable w.r.t. **both** `cdf_grid` and `omega_grid` and runs under jit. Evaluation reuses `interpolate.eval_ppoly` (already namespace-native).
  - `ensure_strictly_increasing(xp, cdf_grid)` — projects a float-saturated CDF grid to strictly increasing (the 8σ tails saturate `F` to float `1.0`, leaving zero steps → singular solve). Measure-preserving and differentiable.
- **`streamdf._dOmega_inverse_cdf_grid`** — the closed-form tilted-Gaussian CDF `p(Ω) ∝ Ω·exp(−½(Ω−mO)²/σ²)` on `Ω∈[max(ε, mO−8s), mO+8s]` (1000 pts), built through `xp` so on a backend it carries `d/d(mO)`, `d/d(σ²)` and traces under jit.
- **`streamdf._sample_aAt` / `sample_t` / `sample`** gain an optional `key=None` (mirrors `streamspraydf.sample`): `None` → global `numpy.random`; a backend key from `galpy.backend.random.key` → reproducible backend arrays. The **same algorithm** runs on every backend — only the RNG source + namespace are dispatched on the key. `ars`/`_h_ars`/`_hp_ars` and the `from ..util import ars` import are dropped (`util/ars.py` stays; `diskdf` still uses it). `streamgapdf._sample_aAt` threads `key=` through.

## numpy is INTENTIONALLY not byte-identical (user pre-approved)

The numpy path changes: ARS → spline inverse-CDF, and the RNG stream shifts (a uniform for the frequency draw replaces the ARS internals). This is a deliberate, approved change. The **distribution is preserved**, with the following evidence:

- **KS(old ARS vs new spline)** on the along-eigenvector frequency: **stat 0.0026, p≈0.53 at n=1e5** (indistinguishable; matches the prototype's 0.0023/0.64).
- **Moments**: new mean/std match ARS to <1%; the new mean is **10× closer to the analytic value** than ARS (err 3.9e-7 vs 4e-6).
- `sample_t(key=None)` stays **byte-identical** to the historical `numpy.random.uniform(size=n)·tdisrupt`.

## Verification (all in `tests/test_backend_streamdf.py`, Phase D section)

- **Cross-backend (same u):** numpy vs jax/torch agree to **~1e-15** (deterministic transform).
- **grad-vs-FD (the payoff):** through `_sample_aAt` w.r.t. `mO`/`σ²` at fixed `u`, AD h-converges to central FD (rel **~1e-11** in `mO`, **~1e-10** in `σ²`).
- **jit:** `jax.jit(_sample_aAt)` at fixed `n` matches eager.
- **Distributional parity vs old ARS:** KS + moment assertions (fixed seeds, non-flaky).
- Existing `test_streamdf.py` sample tests (already distributional) still pass; `streamgapdf` sample tests pass.

## Scope / follow-up

The full `(R,vR,...)` propagation still goes through the numpy `_approxaAInv` (a Python loop over track points). So a backend `key` with **`returnaAdt=True` works end-to-end** (differentiable `(Ω,angle,dt)`); **`returnaAdt=False` on a backend** awaits the **Phase-E** backend track inverse — noted, not forced.

## Note for review

`spline_inverse_cdf_sample` does **not** call `interpolate.cubic_spline_coeffs` (as originally sketched) because that helper treats its `x` as fixed numpy geometry and cannot be traced/differentiated in `x` — exactly what inverse-CDF needs. The dedicated in-backend builder is self-contained and does not touch the shared spline code. Float64 is required (galpy's backend contract).

### DIFF
--- a/galpy/backend/sampling.py
+++ b/galpy/backend/sampling.py
@@ -156,5 +156,15 @@ def spline_inverse_cdf_sample(xp, omega_grid, cdf_grid, u):
         cdf_grid[-1]]`` is clamped to that range (returns the edge ``omega``),
         so the sample never leaves ``[omega_grid[0], omega_grid[-1]]``.
     """
+    # The cubic-spline tridiagonal solve is singular in float32 (the
+    # near-zero tail steps fall below float32 eps), which returns SILENT NaN
+    # under jax's default (no jax_enable_x64) rather than failing loudly. galpy's
+    # backends run in float64; fail clearly instead of poisoning the samples.
+    if xp.finfo(cdf_grid.dtype).bits < 64:
+        raise ValueError(
+            "spline_inverse_cdf_sample requires float64 grids (the CDF-inversion "
+            "spline solve is singular in float32 and would return silent NaN); "
+            "for jax set jax_enable_x64=True."
+        )
     coeffs = _natknot_coeffs(xp, cdf_grid, omega_grid)
     return eval_ppoly(xp, cdf_grid, coeffs, u, extrapolate="clip")
--- a/galpy/backend/sampling.py
+++ b/galpy/backend/sampling.py
@@ -1,0 +1,160 @@
+###############################################################################
+#   galpy.backend.sampling: backend-agnostic, differentiable inverse-CDF sampling.
+#
+#   The single home for galpy's spline inverse-CDF sampler, so the same drawing
+#   code runs on numpy, jax, and torch and is differentiable (jax.grad /
+#   torch.autograd) and jit/GPU-able (no Python-level rejection loop, static
+#   shapes). It replaces per-DF adaptive-rejection sampling (ars) with a
+#   reparameterized transform of uniforms u ~ U[0,1]: the sample is a pure,
+#   differentiable function of u and the (distribution-parameter-dependent) CDF
+#   grid, which is exactly what a reparameterization / common-random-numbers
+#   pipeline needs.
+#
+#   WHY A DEDICATED SPLINE BUILDER (not interpolate.cubic_spline_coeffs)
+#   ------------------------------------------------------------------
+#   ``interpolate.cubic_spline_coeffs`` was built for the "fixed grid geometry,
+#   differentiable y-VALUES" case: it converts its ``x`` to numpy
+#   (``numpy.asarray(x)``) to assemble the tridiagonal matrix, so ``x`` is
+#   treated as a CONSTANT and a TRACED/backend ``x`` (as under jax.grad / jax.jit)
+#   raises a TracerArrayConversionError. Inverse-CDF sampling is the opposite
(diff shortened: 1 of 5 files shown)


jobovy:
Adversarial review complete (verdict: commit-with-1-fix). Applied the one confirmed fix in `947d0c3e`: `spline_inverse_cdf_sample` now raises a clear `ValueError` on float32 grids instead of returning silent NaN under jax's default (no `jax_enable_x64`) — the CDF-inversion solve is singular in float32. Everything else in the review held (spline builder ~1e-15 vs scipy, CDF ~1e-8, KS/grad/jit pass, ngrid=1000 confirmed better than 300, numpy byte-identity as claimed).

jobovy:
**py3.14 CI red fixed (`2f85c180`), with a flag for your review.** `test_streamspraydf.py::test_sample_bovy14` (compares streamdf.sample() vs streamspraydf.sample() means at a fixed seed) broke because Phase D's inverse-CDF sampler consumes a different amount of the numpy RNG stream than ARS. Root-caused with an n=3000 diagnostic: fardal15 is pure realization noise; chen24 has a small **real** residual on z1 comp0 (~0.063) and comp5 (~0.10) where the more-accurate spline sampler (KS-validated, ~10× closer to analytic than ARS) diverges slightly from chen24spraydf's *approximation* — the old ARS realization masked it. Fix = re-seed before the streamspray draw (decouples it from streamdf's RNG consumption, no weakening) + nudge only those 2 tolerances (6e-2→8e-2, 1e-1→1.2e-1). **Please confirm you're OK with the 2 tolerance bumps** — the diagnostic backs them but it's your physics test.

PR #1178, opened 2026-07-21
orbit machinery: forced-backend numpy integrator (no-C) + RazorThin scalar-fallback except (+3 torch)

jobovy:
## Summary

Two orbit-machinery fixes for the "scalar under forced backend" cluster (numpy byte-identical):

**Family C — no-C potential python integrator** (`integrateLinearOrbit.py`). `BurkertPotentialNoC` has no C, so the linear integrator falls back to the python leapfrog (`symplecticode`), where a numpy integrator-state array meets a torch force → `TypeError: Concatenation ... NumPy`. Fix: keep the per-step integrator internals in numpy (`with use("numpy", force=True):`, fast) **and** `as_numpy` the force output — needed because the potential's `_amp` is coerced to a torch tensor during `normalize()` under the forced backend and persists on the instance (`_amp(torch) * _force(numpy)` = torch even inside forced-numpy). Both no-ops on numpy → byte-identical.

**Family B — RazorThin scalar-only potential** (`Orbits.py` E()/Jacobi() fallback). `RazorThinExponentialDiskPotential` is scalar-only; `E()`/`Jacobi()` try a vectorized eval then fall back to a scalar loop on `except (ValueError, TypeError, IndexError)`. numpy raises `ValueError` (caught); torch raises `RuntimeError: Boolean value of Tensor ... ambiguous` (not caught) → crash. Fix: add `RuntimeError` to the 5 except tuples. numpy byte-identical (RuntimeError never fires on numpy).

## Flips (+3 torch)
`test_energy_conservation_linear[BurkertPotentialNoC--10.0-False]`, `test_energy_conservation_linear[RazorThinExponentialDiskPotential--10.0-False]`, `test_energy_jacobi_conservation[RazorThinExponentialDiskPotential--10.0--9.0-False]`.

**Timeout gate (RazorThin scalar-loop):** verified 13.5s / 40–56s under torch, and *faster* on fewer cores (the user-time was thread-thrash on scalar ops) → CI-safe. Verified: 3 torch PASS + numpy PASS; numpy regression on other energy/jacobi params unchanged.

### DIFF
--- a/galpy/orbit/Orbits.py
+++ b/galpy/orbit/Orbits.py
@@ -3304,7 +3304,7 @@ class Orbit:
                     )
                     + thiso[1] ** 2.0 / 2.0
                 ).T
-            except (ValueError, TypeError, IndexError):
+            except (ValueError, TypeError, IndexError, RuntimeError):
                 out = (
                     numpy.array(
                         [
@@ -3331,7 +3331,7 @@ class Orbit:
                     + thiso[1] ** 2.0 / 2.0
                     + thiso[2] ** 2.0 / 2.0
                 ).T
-            except (ValueError, TypeError, IndexError):
+            except (ValueError, TypeError, IndexError, RuntimeError):
                 out = (
                     numpy.array(
                         [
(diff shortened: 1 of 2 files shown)


jobovy:
Closing — the Family C fix used a `use("numpy", force=True)` boundary in integrateLinearOrbit, which is a numpy island (forces no-C potential linear integration onto numpy). Redoing properly as a backend-native migration of `symplecticode.leapfrog` (makes no-C potential linear orbits run + differentiate on the backend, no island), plus the RazorThin E()/Jacobi() except-broaden. New PR to follow.

PR #1187, opened 2026-07-23
interpRZPotential: backend-native evaluation (native 2D/1D interp) + fixes StaeckelGrid interpecc divergence

jobovy:
## Summary

The **2D** interpRZ backend evaluation (potential, forces, 2nd derivatives, density) landed in #1116/#1176. This PR completes the migration (the **1D** interpolators) and fixes a StaeckelGrid backend-vs-numpy divergence that the migration unblocks.

### 1. 1D interpolators backend-native (`vcirc` / `dvcircdR` / `epifreq` / `verticalfreq`)
Dual-path exactly like the 2D methods: a numpy `R` keeps calling the fitted scipy `InterpolatedUnivariateSpline` (**byte-identical**); a jax/torch `R` evaluates the *same* frozen piecewise polynomial via `spline_to_ppoly` + `eval_ppoly` (searchsorted + Horner in namespace ops), matching scipy to ~1 ulp and exactly autodifferentiable in `R`. `scalarDecorator` is made backend-aware (skips the numpy scalar normalization for a backend `R`), mirroring `scalarVectorDecorator`.

### 2. StaeckelGrid interpecc divergence fixed
When building an `actionAngleStaeckelGrid(interpecc=True)` on an `interpRZPotential` under a forced backend, the bad (extreme) orbits' `(ecc,zmax,rperi,rap)` were recomputed with the **original**-potential Staeckel (`tmpaA`) — because interpRZ could not be backend-evaluated — whereas the numpy build uses the **interp**-potential Staeckel (`self._aA`). Now that interpRZ evaluates in-backend, the backend recompute uses `self._aA` to match the numpy build. The recompute runs under `use("numpy", force=True)`, so it is byte-identical to the numpy build.

**Isolated effect** (same 226 bad orbits, `tmpaA` vs `self._aA`): mean ecc diff **0.67**. **Query-level** (off-grid orbits, numpy-grid vs jax-grid):

| quantity | before (buggy) | after (fixed) |
|---|---|---|
| ecc  | mean 28% / max 93% | mean 0.01% / max 0.5% |
| rperi | mean 26% / max 92% | mean 0.01% / max 0.6% |

## Verification
- **numpy BYTE-IDENTICAL**: `array_equal` (maxdiff 0.0) for Phi/Rforce/zforce/R2deriv/z2deriv/Rzderiv/dens + the four 1D quantities, before/after (A/B via `git checkout`).
- **backend == scipy**: 2D ~1e-16, 1D ~1e-16–1e-15; result `is_backend_array` asserted.
- **differentiable**: d(vcirc)/dR (and 2D force/pot) grad-vs-FD matched (jax + torch); jit-safe.
- **divergence fixed**: StaeckelGrid interpecc ecc/rperi query divergence ~28%→~0.01% (jax regression test).
- numpy `tests/test_interp_potential.py`: 40 passed. Backend `tests/test_backend_interprz.py` under the `-W error::Deprecation/Future` shard: 34 passed, 1 skipped.
- No `tests/backend_xfail.txt` edits.

## Deferred
- The torch StaeckelGrid-**build**-on-interpRZ path hits a *separate pre-existing* numpy-scalar-under-forced-torch crash (`torch.sqrt` rejects the numpy-scalar `vcirc(Rmax)` in `_build_grid_backend`) — crashes identically on the base branch, unrelated to this divergence fix. The divergence fix is backend-agnostic and verified on jax; the torch divergence test is skipped with a note.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelGrid.py
+++ b/galpy/actionAngle/actionAngleStaeckelGrid.py
@@ -555,8 +555,14 @@ class actionAngleStaeckelGrid(actionAngle):
                 mjr = asarray_on_device(xp, mjr_np, dev)
                 mjz = asarray_on_device(xp, mjz_np, dev)
                 if interpecc:
+                    # ecc-family recompute uses the INTERP-potential Staeckel
+                    # (self._aA), matching the numpy grid build (which uses
+                    # self._aA.EccZmaxRperiRap here, not the orig-potential
+                    # tmpaA used for the jr/jz sentinels above); now that
+                    # interpRZPotential evaluates in-backend this stays in sync
+                    # so the backend ecc/zmax/rperi/rap grids == the numpy ones.
                     with use("numpy", force=True):
-                        recc, rzmax, rrperi, rrap = tmpaA.EccZmaxRperiRap(
+                        recc, rzmax, rrperi, rrap = self._aA.EccZmaxRperiRap(
                             R_np[bad_np],
                             v_np[bad_np] * numpy.cos(psi_np[bad_np]),
                             Lz_np[bad_np] / R_np[bad_np],
--- a/galpy/actionAngle/actionAngleStaeckelGrid.py
+++ b/galpy/actionAngle/actionAngleStaeckelGrid.py
@@ -527,60 +527,55 @@ class actionAngleStaeckelGrid(actionAngle):
         if isinstance(self._pot, potential.interpRZPotential) and hasattr(
             self._pot, "_origPot"
         ):
-            # Interpolated potentials have problems with extreme orbits: recompute
-            # the 9999.99 flags with the original potential (rare; numpy solver).
+            # Interpolated potentials fail on extreme orbits (jr/jz == 9999.99);
+            # recompute those with the ORIGINAL potential's Staeckel, matching the
+            # numpy build.
             bad = (mjr == 9999.99) | (mjz == 9999.99)
-            if int(as_numpy(xp.sum(bad))) > 0:
-                bad_np = as_numpy(bad)
-                with use("numpy", force=True):
-                    tmpaA = actionAngleStaeckel.actionAngleStaeckel(
-                        pot=self._pot._origPot, delta=self._delta, c=self._c
-                    )
-                    R_np = as_numpy(thisR)
-                    v_np = as_numpy(thisv)
-                    psi_np = as_numpy(thispsi)
-                    Lz_np = as_numpy(thisLzs)
-                    rjr, _, rjz = tmpaA(
-                        R_np[bad_np],
-                        v_np[bad_np] * numpy.cos(psi_np[bad_np]),
-                        Lz_np[bad_np] / R_np[bad_np],
-                        numpy.zeros(int(bad_np.sum())),
-                        v_np[bad_np] * numpy.sin(psi_np[bad_np]),
-                        fixed_quad=True,
-                    )
-                    mjr_np = as_numpy(mjr).copy()
-                    mjz_np = as_numpy(mjz).copy()
-                    mjr_np[bad_np] = rjr
-                    mjz_np[bad_np] = rjz
-                mjr = asarray_on_device(xp, mjr_np, dev)
-                mjz = asarray_on_device(xp, mjz_np, dev)
-                if interpecc:
-                    # ecc-family recompute uses the INTERP-potential Staeckel
-                    # (self._aA), matching the numpy grid build (which uses
-                    # self._aA.EccZmaxRperiRap here, not the orig-potential
-                    # tmpaA used for the jr/jz sentinels above); now that
-                    # interpRZPotential evaluates in-backend this stays in sync
-                    # so the backend ecc/zmax/rperi/rap grids == the numpy ones.
-                    with use("numpy", force=True):
-                        recc, rzmax, rrperi, rrap = self._aA.EccZmaxRperiRap(
-                            R_np[bad_np],
-                            v_np[bad_np] * numpy.cos(psi_np[bad_np]),
-                            Lz_np[bad_np] / R_np[bad_np],
-                            numpy.zeros(int(bad_np.sum())),
-                            v_np[bad_np] * numpy.sin(psi_np[bad_np]),
-                        )
-                        mecc_np = as_numpy(mecc).copy()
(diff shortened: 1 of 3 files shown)


jobovy:
## Follow-up: backend-native jr/jz bad-orbit recompute (removed one numpy island)

Addressed the review that the `use("numpy", force=True)` islands in the StaeckelGrid interpecc bad-orbit recompute should go now that interpRZ evaluates in-backend. Outcome:

### jr/jz recompute → BACKEND-NATIVE (island removed)
The original potential's Staeckel action solve now takes the backend inputs directly (recompute for all orbits, keep only where the interp solve failed via `xp.where` — namespace-generic, dead-branch-safe). The C-native action-solve **value** path matches numpy and **preserves the `9999.99` failure sentinel**, so:
- backend tables are **byte-identical** to the numpy-island version (the `_nanmax_ne` per-Lz maxima still exclude unbound orbits),
- no `use("numpy")`/`as_numpy` in the jr/jz block (grep-confirmed).

### ecc-family recompute → stays a documented numpy computation (load-bearing)
This one **cannot** be made backend-native without reintroducing the divergence. **Root cause:** the backend C-STM `EccZmaxRperiRap` returns **finite AD-safety-guarded** values for the degenerate/unbound bad orbits (see `test_staeckel_ecczmax_grad_degenerate`), whereas numpy returns the **NaN/inf sentinels** that the grid's inf-excluding `_max_finite` per-Lz normalizers depend on. A fully backend-native ecc recompute makes `_zmaxLzE` explode from ~1.4 to **~9984** (the finite ~9999 gets *included* instead of *excluded*), wrecking the zmax/rperi/rap normalization → **~28% divergence** (measured). NaN/inf classification on the 226 bad orbits: numpy → (ecc=NaN, zmax/rperi/rap=inf) uniformly; backend → all finite. This is a **separate `actionAngleStaeckel` C-STM limitation on degenerate orbits, not an interpRZ gap** (interpRZ now evaluates fully in-backend). So the bad-orbit ecc-family is computed once in numpy and lifted back to the backend.

### Verification
- **numpy StaeckelGrid grid BYTE-IDENTICAL** (A/B vs base): maxdiff **0.0** across all 15 tables (`_jr,_jz,_ecc,_zmax,_rperi,_rap,_*LzE,_u0,_Lzs,_ERL,_ERa`).
- **grep**: jr/jz recompute region has no `use("numpy")`/`as_numpy`; the ecc block retains one documented numpy computation.
- **backend tables match numpy grid** (identical to prior commit): query-level ecc **0.036% mean / 0.49% max** — divergence stays fixed (vs ~28% if the ecc island is removed).
- **`is_backend_array` True** for every frozen table (GPU-resident).
- **`-W error` shard green** (34 passed, 1 skipped); **non-interpRZ native-setup backend grid green on jax AND torch** (7 passed).
- The **torch interpRZ-grid** path stays blocked by the pre-existing `vcirc(Rmax)` numpy-scalar crash in `_build_grid_backend` (unrelated; divergence test skips torch; non-interpRZ grids unaffected).

jobovy:
## Update: ecc-family recompute is now ALSO backend-native (sentinel-marking) — both islands removed

Following the review, I revisited the ecc-family island. It is **not** load-bearing after all — there's a cleaner route than a backend *recompute* (which returns finite AD-safe values and breaks the inf-excluding normalization). The amended commit removes **both** `use("numpy", force=True)` islands from the interpecc bad-orbit recompute.

### Rigorous uniformity check (the gate)
A `jr/jz` failure (action solve → `9999.99`) always coincides with a turning-point failure (unbound orbit), so numpy's `self._aA.EccZmaxRperiRap` **never recovers a finite value** for a bad orbit — it returns the potential's deterministic C turning-point sentinels. Verified across **4 potentials × multiple grid resolutions (696/696 bad orbits, zero finite/recovered ecc)**:

| config | n_bad | ecc | zmax | rperi | rap |
|---|---|---|---|---|---|
| MW2014 (fine) | 226 | NaN | +inf | **−inf** | +inf |
| thin-MN (b=0.05) | 309 | NaN | +inf | **−inf** | +inf |
| LogHalo (q=0.9) | 41 | NaN | +inf | **−inf** | +inf |
| NFW | 0 | — | — | — | — |
| MW2014 (coarse) | 120 | NaN | +inf | **−inf** | +inf |

These are a deterministic consequence of the C `umin/umax = -9999.99` failure flags (`rperi = δ·sinh(umin) = −inf`, `zmax = δ·sinh(|umax|) = +inf`, etc.).

### The fix
Mark the bad orbits with those exact sentinels via `xp.where` (backend-native, no numpy):
```python
mecc   = xp.where(bad, xp.full_like(mecc, numpy.nan),  mecc)
mzmax  = xp.where(bad, xp.full_like(mzmax, numpy.inf),  mzmax)
mrperi = xp.where(bad, xp.full_like(mrperi, -numpy.inf), mrperi)   # NOT +inf
mrap   = xp.where(bad, xp.full_like(mrap, numpy.inf),   mrap)
```
The raw tables then equal the numpy grid's, so the inf-excluding `_max_finite` per-Lz normalizers + the NaN/inf clamps reproduce the numpy grid exactly.

**Important correction to the suggested snippet:** `rperi` must be **−inf**, not +inf. After normalization, `+inf` clamps `>1 → 1`, but numpy's `−inf` clamps `isinf → 0` — so `+inf` would diverge. The table above confirms numpy returns −inf.

### Verification
- **sentinel-marking ≡ numpy-island: BIT-IDENTICAL** backend tables (`equal_nan`) across MW2014 + thin-MN — the marking is provably equivalent, not an approximation.
- **numpy StaeckelGrid grid BYTE-IDENTICAL** vs base (maxdiff **0.0**).
- **grep**: **ZERO** `use("numpy",force=True)`/`as_numpy` remaining in the entire interpecc recompute block (only the unrelated `rl`/`calcu0` solves stay numpy).
- **`is_backend_array` True** for every frozen table.
- **divergence stays fixed**: query ecc **0.036% mean** (MW2014). (thin-MN shows ~10% — that's the pre-existing `calcu0` bad-mask-flip floor, identical with the island.)
- **`-W error` shard green** (34 passed, 1 skipped); **non-interpRZ native-setup backend grid green on jax AND torch** (6 passed).
- Torch interpRZ-grid still blocked by the pre-existing `vcirc(Rmax)` numpy-scalar crash (unrelated; divergence test skips torch).

PR #1194, opened 2026-07-24
backend: @backend_kernel decorator lifting coerce boilerplate across potentials (+ jit seam)

jobovy:
## What

Introduce `galpy.backend.backend_kernel`, a decorator that lifts the inline

```python
xp = get_namespace(R, z)
R, z = coerce_coords(xp, R, z)
```

boilerplate — copy-pasted into every migrated potential method only because the
backend migration was incremental — into one decorator, and roll it out across
**~290 methods in ~40 potential files**. The decorated method becomes the pure
kernel: it declares its coordinate args explicitly (`@backend_kernel("R", "z")`,
the same information the old `coerce_coords` call carried) and receives them
coerced plus an injected `xp`:

```python
@backend_kernel("R", "z")
def _evaluate(self, R, z, phi=0.0, t=0.0, *, xp=None):
    r2 = R**2.0 + z**2.0
    return xp.log(r2) / 2.0
```

## Why

1. **Maintainability** — DRY; the coercion/namespace logic lives in one place.
2. **The jit seam** — the decorator separates outer Python glue (namespace
   resolution, coercion) from the inner traceable kernel, which is exactly the
   split `jax.jit` / `torch.compile` need. galpy itself never jits (no-internal-jit
   policy); the forthcoming `--backend jax-jit` / `torch-compile` test dimension
   wraps the decorated kernels. Prototyped and validated (jit==eager, grad flows,
   incl. the axisymmetric `phi=None` case).
3. **Torch-ledger burndown** — jax tolerates numpy inputs (`jnp.sin(numpy)`
   promotes) while torch rejects them (`torch.sin(numpy)` raises). So a method
   that resolved `xp` but didn't coerce ran green under `--backend jax` while
   being a latent torch crash. `@backend_kernel` always-coerces the declared
   coords, so every decorated entry point becomes torch-safe.

## Byte-identity

The numpy path is a strict pass-through — `get_namespace` returns numpy,
`coerce_coords` is object-identity, and the kernel body is textually unchanged —
so decorated methods are bit-for-bit identical to the inline version. This is
enforced by an audit (every changed line is one of {decorator, `*, xp=None`
signature, boilerplate/coerce deletion, import}) and confirmed empirically by
per-file stash-and-compare (hundreds of method outputs `array_equal` to
baseline). The full `test_potential.py` numpy suite passes.

## Scope / what's deferred

Applied to the primitive potential / force / density / derivative kernels. Left
inline (correctly): methods with instance-attr or no-arg `get_namespace`,
guard-first bodies (`_mass`), `if xp is numpy:` dual-path dispatchers
(MultipoleExpansion, SCFPotential, interpSpherical), `is_backend_array`
data-guards, and `promote_scalars`. Deferred to follow-ups:

- the base `Potential.py` and the derived-quantity accessors (`vcirc`/`omegac`/
  `epifreq`/`vesc`/`rforce`) — they interact with the unit-decorator stack;
- the `--backend jax-jit` / `torch-compile` test dimension + jit-xfail ledger.

One design nuance flagged in the commits: for the `get_namespace ⊃ coerce_coords`
mismatch (e.g. time-dependent potentials), whether to declare `t` as a coord is a
per-potential call (declaring it enables t-autodiff but makes `t` traced, which
breaks any Python `if t > tform` branch under jit) — the jit dimension will
surface these; byte-identity holds either way.

### DIFF
--- a/galpy/backend/__init__.py
+++ b/galpy/backend/__init__.py
@@ -17,6 +17,7 @@ from ._coerce import (
     promote_scalars,
     zeros_like_backend,
 )
+from ._kernel import backend_kernel
 from ._namespaces import (
     as_numpy,
     asarray_on_device,
@@ -41,6 +42,7 @@ _seed_from_config()
 
 __all__ = [
     "autodiff_ops",
+    "backend_kernel",
     "get_namespace",
     "backend",
     "use",
--- a/galpy/backend/_kernel.py
+++ b/galpy/backend/_kernel.py
@@ -1,0 +1,104 @@
+###############################################################################
+#   galpy.backend._kernel: the @backend_kernel decorator.
+#
+#   Lifts the inline
+#
+#       xp = get_namespace(R, z)
+#       R, z = coerce_coords(xp, R, z)
+#
+#   boilerplate -- copy-pasted into every migrated method only because the
+#   backend migration was incremental -- into one decorator. The decorated
+#   function becomes the PURE kernel: it receives the coerced coordinate
+#   arguments plus an injected ``xp`` keyword and does nothing but the math.
+#
+#   This also defines the jit seam. jax.jit / torch.compile want exactly this
+#   split: outer Python glue (namespace resolution, coercion) separated from the
+#   inner traceable kernel. galpy itself never jits (see the no-internal-jit
+#   policy); the test suite's --backend jax-jit / torch-compile dimension wraps
+#   the decorated kernels.
+#
+#   The coordinate arguments are declared EXPLICITLY -- @backend_kernel("R","z")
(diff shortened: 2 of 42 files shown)


jobovy:
Superseded by #1196. After review, the per-method `@backend_kernel` rollout is replaced by a single dedicated boundary decorator `@backend_input` (galpy.backend) that owns coordinate coercion, stacked just inside the units-only `potential_physical_input`. That keeps the backend concern cleanly separated from the legacy unit machinery without touching every method body or requiring an injected `xp`. Namespace resolution stays inline (`get_namespace`) in the methods, and the jit dimension becomes boundary-jit (jit the whole public method → the decorators trace away), so no per-method jit seam is needed.

PR #1195, opened 2026-07-24
backend: --backend jax-jit jit test dimension (jit seam)

jobovy:
Stacked on #1194 (needs the `@backend_kernel` decorator as the seam).

## What

Add the `--backend jax-jit` / `torch-compile` test dimension the decorator was built
to enable. A module flag `galpy.backend._kernel._JIT_MODE` (set only by conftest --
galpy never jits itself, per no-internal-jit) makes each decorated kernel, after
coercing its declared coords, trace through `jax.jit` / `torch.compile` over those
coords (now backend arrays), closing over everything else (self, xp, `phi=None`,
scalar t) as static.

conftest gains two `--backend` values that force the base backend (jax/torch, **with
float64** -- `jax_enable_x64` / `set_default_dtype`, exactly like the eager runs) and
set `_JIT_MODE`. They key the xfail-ledger/regen off the full name, so jit failures
get their own `jax-jit` / `torch-compile` lists.

## Validated

- `jit==eager` across spherical / axisymmetric / non-axi / bar / spiral / Staeckel
  potentials; grad flows through the jit-hooked kernel.
- A usual test (`test_dvcircdR_omegac_epifreq_rl_vesc`) passes under `--backend
  jax-jit` -- the dimension reaches the normal (non-backend_managed) suite and uses
  float64.
- `test_backend_kernel` gains a jit-mode test (jax + torch) asserting the decorated
  kernel matches eager under `_JIT_MODE`.
- Finding: `torch.compile` traces simple potentials fine but its dynamo chokes on
  complex kernels (e.g. SpiralArms -> `InternalTorchDynamoError`); those become
  torch-compile ledger entries.

## Follow-ups (not in this PR)

- Seed the `jax-jit` / `torch-compile` xfail + slow-skip ledgers by a regen run over
  the usual suite (islands like interpRZ, `|z|`-kink control flow like RazorThin,
  dynamo-hostile kernels), then add the CI job.
- Kernel-caching in the hook (jit once per potential+shape) -- the current fresh-jit
  -per-call is correct for traceability but slow for the full suite; heavy tests
  slow-skip like the eager-jax ledger until then.

### DIFF
--- a/galpy/backend/_kernel.py
+++ b/galpy/backend/_kernel.py
@@ -31,6 +31,21 @@ import inspect
 from ._coerce import coerce_coords
 from ._resolver import get_namespace
 
+# Test-only jit seam. None in normal galpy operation (galpy never jits itself --
+# the no-internal-jit policy); the test suite's --backend jax-jit dimension sets
+# this so every decorated kernel is traced through jax.jit with the declared
+# (coerced) coords as the traced inputs. (torch-compile is a deferred follow-up.)
+_JIT_MODE = None  # None | "jax"
+
+
+def set_jit_mode(mode):
+    """Enable jit wrapping of decorated kernels (tests/conftest only).
+
+    mode in {None, "jax"}. galpy itself never calls this.
+    """
+    global _JIT_MODE
+    _JIT_MODE = mode
+
 
 def backend_kernel(*coord_names):
     """Resolve the array namespace from the named coordinate arguments, coerce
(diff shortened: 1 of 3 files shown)


jobovy:
Superseded — this stacked on the `@backend_kernel` per-method jit seam (#1194, now closed). The jit dimension will be reworked as boundary-jit (jit the whole public method so the decorators trace away) on top of the `@backend_input` architecture in #1196, which needs no per-method seam.

PR #1204, opened 2026-07-26
Autonomous jit/compile development

jobovy:
Accumulator branch for autonomous work while you were away (2026-07-26 →). 121 commits, each landed as its own reviewed-by-CI sub-PR into `feat/backends-develop` and squash-merged. This is the review-deferred vehicle: the sub-PR commit messages are written to be read later, out of context, all at once.

## Goal

Drive galpy to a **fully jittable/compilable state**: the whole existing test suite passing under jax and torch, the exclusion ledger **emptied to nothing**, and no numpy islands — with the **numpy path byte-identical** throughout.

## Outcome against that goal — not met, and here is the gap

| | goal | at branch point | now |
| --- | --- | --- | --- |
| `backend_xfail.txt` (real failures) | 0 | 151 | **8** |
| `backend_slow_skip.txt` (too slow to run) | 0 | 229 | **60** |
| `backend_skip.txt` (by-design / out of scope) | — | 12 | **72** |
| **total excluded** | **0** | **392** | **140** |

Current suite census:

```
jax:   3543 passed ·  2 xfail · 33 deferred · 34 exempt
torch: 3573 passed ·  2 xfail ·  3 deferred · 34 exempt
```

The ledger was supposed to reach zero and did not. What remains is 8 real failures and 60 deferrals. The growth in `skip` is **reclassification, not new breakage** — entries moved from "broken" to "deliberately not applicable" (12 of them are the `actionAngleTorus` family, which is out of scope by design), and every line now carries its reason.

**The 30 remaining jax deferrals are one root cause, not 30 problems:** jax eager per-call overhead (~5 ms flat, 186× on a scalar call) applied to Python-loop orbit integration and per-pixel grid loops. Every one of them now carries a measured cost in the ledger. Fixing them is the vectorization project — weeks of work on the orbit hot path — which I deliberately did not start before this review.

## What changed

154 files, +12073/−2756 — but **+7936/−1406 of that is tests**. Library changes are ~4100 lines added, concentrated in `galpy/backend` (+1337), `galpy/potential` (+1325), `galpy/util` (+664), `galpy/df` (+661).

Three strands:

1. **Make traced execution possible at all** — the `--jit` harness with its own skip list and a traced-fraction metric, then removing what breaks under tracing: Python `if`s on array data, `.numpy()` boundary crossings, unpicklable spline holders, coercion at call boundaries via `@backend_input`.

2. **Close the backend gaps tracing exposed** — special functions with no usable jax/torch native (hyp2f1/hyp1f1 domain gaps, `exp1`, `gammainc`/`gammaincc`), quadrature that NaN'd at endpoints, `parallel_map` forking into a deadlock under jax, device/dtype anchoring.

3. **Burn the ledger down**, which is the actual evidence that 1 and 2 worked.

## Reviewing this

Suggested order:

- **`tests/backend_*.txt`** first — it is the scoreboard, and each line states why it is there.
- **`galpy/backend/`** — the new dispatch/special/quadrature machinery everything else leans on.
- Then the per-module migrations, which are repetitive once the pattern is clear.

Two things to have in hand:

- **CI never runs `--jit`.** The `jax-jit`/`torch-jit` ledger rows come from local runs only, so the tracing work is validated by my measurements rather than by the green shard you see here. That is a real limit on what this PR's CI proves.
- **Numpy byte-identity** is the invariant I was holding to; where a numpy-visible change was unavoidable it is called out in the relevant sub-PR's message rather than folded in silently.

## Known-imperfect, flagged rather than hidden

- 8 xfails remain: 2 wait on the `main`→develop rebase, 2 are by-design qdf, 4 are jit-mode entries CI never exercises.
- torch's `gammaincc` keeps its ~5e-10 accuracy for a ≳ 21 (pre-existing, not a regression) — see #1374, which fixed the differentiability gap but explicitly not the precision one.
- Issues filed for pre-existing numpy bugs found along the way rather than fixed here: #1276, #1277, #1343, #1354.


jobovy:
I did a second, deeper pass through the PR using the commit stack as the index. Overall, I think this is appropriate to merge into `#892` as the new supervised continuation branch: the work is coherent, the ledger is honest, and I did not find broad duplication against `galpy.backend`. Most of the runtime work moves repeated patterns into shared backend utilities rather than scattering backend-specific patches.

I do think the following should be fixed before treating this as a reviewed base:

1. `galpy/backend/special/_fallback/gammainc.py:196`

   The new torch autograd path returns `nan` for `d/dx gammainc(a, x)` and `gammaincc(a, x)` at `x=0`. Values are correct, but gradients are not. For example, with `x=torch.tensor(0., requires_grad=True)`, `a=1,2,3` gives `x.grad == nan`; the derivative should be finite (`+/-1` at `a=1`, zero for `a>1`). The endpoint value tests exist, but endpoint gradient tests are missing.

2. `galpy/backend/quadrature.py:622`

   `finite_part_quad` is documented as accepting array `b`, `c`, and `peak_width`, but the residual-log slope estimate is scalar-only:

   ```python
   lam = xp.sum(_LOG_W * residual(b * _LOG_U))
   ```

   `b.shape == (2,)` fails to broadcast against `_LOG_U.shape == (5,)`; and even if it did broadcast, the reduction lacks `axis=-1`, so it would collapse batch dimensions into one scalar `lam`. This should probably evaluate at `b[..., None] * _LOG_U` and reduce over the sample axis.

3. `galpy/potential/Potential.py:128`, `:817`, `:914`

   Generic surface-density quadrature gates backend quadrature only on `under_trace(absz)`. That misses differentiation/tracing with respect to `R` when `z` is fixed. I reproduced this with JAX: `grad(lambda R: p.surfdens(R, fixed_z, forcepoisson=True))` falls into SciPy and raises a concretization error, while the analogous `z` gradient works. The gate should consider all traced/differentiable inputs that affect the integrand, not only `absz`.

4. `galpy/util/coords.py:100`, `:1217`, `:1591`

   `galcenrect_to_XYZ` and `galcenrect_to_vxvyvz` are marked backend-native, but `_galcen_rot(Xsun, Zsun)` is hard-coded through numpy. Passing backend `Xsun`/`Zsun` requiring grad fails immediately on torch with “Can’t call numpy() on Tensor that requires grad.” The Jacobian path already handles backend `Xsun`/`Zsun`, so the transformed-coordinate path should too.

5. `galpy/backend/_jit.py:39` plus `galpy/backend/_input.py:179`

   Public `set_jit("jax")` does not by itself trace plain Python/numpy coordinate calls. It only traces once the backend is non-numpy, e.g. inside `use("jax")` or with backend inputs. I verified that `set_jit("jax"); p.Rforce(1.0, 0.1)` returns a plain `float`, while the same call inside `use("jax")` returns a JAX array. This is probably an API/documentation mismatch rather than a core harness bug, because the test harness combines `--backend` and `--jit`, but `_jit.py` advertises `set_jit("jax")` as tracing every boundary call.

Outside of these, the structure looks good to me. `backend.optimize`, backend quadrature, shared jit helpers, the pickle mixin, and the adapter boundary-crossing fixes are all appropriate consolidations. The xfail/slow-skip/permanent-skip split is also reasonable and tested. I did not run the full backend matrix; this review is based on diff/commit inspection plus targeted repros for the issues above.

jobovy:
Additional deep-pass findings from a second review pass. These are separate from the earlier comment; I am not re-listing the five findings already posted there. I have scoped this to issues tied to the PR #1204 diff rather than broader backend follow-up work.

1. Several coordinate transforms touched by the coordinate backend migration still force backend inputs through NumPy

The migrated coordinate coverage is still asymmetric. The backend-native direction works for several transforms, but inverse/adjacent helpers in the same PR-touched coordinate surface still use the old NumPy-promoting decorator or explicit NumPy assembly:

- `galpy/util/coords.py:430` (`lbd_to_XYZ`)
- `galpy/util/coords.py:552` (`vrpmllpmbb_to_vxvyvz`)
- `galpy/util/coords.py:1129` (`XYZ_to_galcenrect`)
- `galpy/util/coords.py:1362` (`XYZ_to_galcencyl`)
- `galpy/util/coords.py:1432` (`vxvyvz_to_galcenrect`)
- `galpy/util/coords.py:2209` (`galcencyl_to_galcenrect`)
- `galpy/util/coords.py:3020` (`lambdanu_to_Rz`)

I verified these fail with torch tensors carrying `requires_grad=True` (`RuntimeError: Can't call numpy() on Tensor that requires grad`) and under `jax.jit` (`TracerArrayConversionError` or tracer truth-value errors). This leaves common round-trip/inverse coordinate paths non-differentiable even though their opposite directions have been migrated in the same area.

2. The PR-touched `map_coordinates` path fails when a backend-prefiltered grid is queried at fixed NumPy/Python points

The `map_coordinates` changes keep dispatch keyed on the query coordinates rather than on the filtered grid:

- `galpy/backend/interpolate.py:794` (`map_coordinates`)

`map_coordinates(spline_filter(grid_backend), coords_numpy)` tries to materialize the backend-filtered grid with `numpy.asarray(...)`, which either errors for torch AD tensors or raises JAX tracer conversion errors. For backend prefiltered grids, fixed query points should be promoted into the grid's backend namespace instead.

3. Some PR-touched constructor-gradient paths still are not externally `jax.jit` safe

Several constructor paths touched by the PR pass eager `jax.grad` style coverage but fail when the same construction is inside `jax.jit`, because Python truth checks run on traced parameters:

- `galpy/potential/ExpTruncNFWPotential.py:116` uses `min(a, rc)` for `_small_r_thresh` after this PR coerces `a, rc` onto the backend
- `galpy/potential/TwoPowerTriaxialPotential.py:110` and `galpy/potential/TwoPowerTriaxialPotential.py:119` branch on `alpha`/`beta`-derived values while the PR adds backend-special handling for the same constructor quantities
- `galpy/df/constantbetaPowerLawdf.py:68` and `galpy/df/constantbetaPowerLawdf.py:99` assert on traced `beta`/`n`, so the new backend-special normalization path cannot be reached under externally-jitted construction

I verified representative failures with `jax.jit(jax.grad(...))`: `ExpTruncNFWPotential(... a=tracer ...)`, `ExpTruncNFWPotential.from_nfw(... mass=tracer ...)`, `TwoPowerTriaxialPotential(alpha=tracer)`, and `constantbetaPowerLawdf(beta=tracer)`. If construction inside the externally-jitted user function is in scope for these PR changes, the constructor tests need jitted variants, not only eager AD.

4. The PR-touched `hyp2f1` fallback is not generally traceable in scalar parameters

The hypergeometric fallback changed by this PR still treats `a,b,c` as structural Python scalars:

- `galpy/backend/special/_fallback/hyp2f1.py:51`
- `galpy/backend/special/_fallback/hyp2f1.py:92`
- `galpy/backend/special/_fallback/hyp2f1.py:156`
- `galpy/backend/special/_fallback/hyp2f1.py:173`

Direct `jax.jit(jax.grad(...))` wrt `hyp2f1`'s `a` or `c` fails with tracer boolean/concretization errors. That matters for PR-touched code paths where potential parameters feed these scalar slots, e.g. the new backend-special routing in power-law/triaxial expressions. If scalar special-function parameters are intended to be backend-differentiable, the fallback routing/control flow needs to be explicit about that rather than assuming only `z` carries AD.

jobovy:
One more PR-specific issue from the follow-up pass:

`galpy/backend/quadrature.py:533` rejects finite eager torch limits that carry gradients in the new `symmetric_quad` helper.

The finite/infinite branch check is:

```python
if not under_trace(b) and not bool(numpy.all(numpy.isfinite(numpy.asarray(b)))):
```

For an eager torch tensor with `requires_grad=True`, `under_trace(b)` is false, so the code tries `numpy.asarray(b)` before taking the finite backend path. That fails even when `b` is an ordinary finite tensor:

```python
import torch
import array_api_compat.torch as txp
from galpy.backend.quadrature import symmetric_quad

b = torch.tensor(1.0, requires_grad=True)
symmetric_quad(txp, lambda s: s * s, b)
```

This raises:

```text
RuntimeError: Can't call numpy() on Tensor that requires grad. Use tensor.detach().numpy() instead.
```

This helper is added by the PR, so I think the concreteness check needs to treat backend tensors as finite-branch inputs unless they are concretely known NumPy/Python infinities, or otherwise use a torch-safe detached check that does not break AD for the finite case.

jobovy-agent:
**Review finding 5 (`set_jit` / `_jit.py`) — resolved as "working as intended"; no behaviour change.**

You're right about the observed behaviour, and right that it's an API/documentation
mismatch rather than a harness bug. Spelling out why we're keeping the semantics:

`set_jit` does not select a backend — it *arms* tracing conditionally. The framework
that actually decides comes from either `use(...)` **or the data**. Both paths are
exercised in `tests/test_backend_jit_mode.py`, neither of them inside a `use(...)`:

* `test_jax_traced_matches_eager` passes `jnp.asarray(...)` under `jit("jax")` and traces.
* `test_numpy_path_ignores_jit_mode` passes plain floats under `jit("jax")` and does not —
  its docstring states that as the contract: *"Trace mode must not touch the numpy path."*

And this is more than "it doesn't raise": `test_jax_reuses_the_compiled_trace`,
`test_jax_nested_boundaries_do_not_retrace` and `test_jax_retraces_after_a_parameter_changes`
all *count traces*, and they also run without `use(...)`. So tracing genuinely happens off
data-implied backends — `set_jit("jax")` + jax arrays is a supported, tested pattern, not an
accident.

The decisive reason not to make `set_jit` error (or enable the backend):
**`set_jit` is called before the data exists.** The call it will affect hasn't happened yet,
so validating at `set_jit` time can only check the *ambient* backend — exactly the wrong thing
when the pattern is data-driven. I implemented the strict version to check: it takes
`tests/test_backend_jit_mode.py` to **12 failures / 16**, because it rejects the legitimate
`jnp`-input case and reverses the documented numpy no-op above. Having `jit(mode)` enable the
backend instead was rejected for a different reason: it's a context manager, so it would gain
two exit obligations and silently swap the caller's backend on entry, which `use(...)` already owns.

So the composition stays explicit, and reads fine:

```python
with use("jax"), jit("jax"):   # ambient backend
    ...
with jit("jax"):               # or let the data imply it
    pot.Rforce(jnp.asarray(1.0), jnp.asarray(0.1))
```

That leaves exactly the residue you identified: the `_jit.py` docstring says "Trace every
boundary call under the named framework", which overstates it. That wording is the only thing
that will change — to something like *"arm tracing for boundary calls that resolve to the named
framework; a numpy call is unaffected"* — folded into the next review PR rather than into the
coords PR whose CI is currently running.

jobovy-agent:
**Review status: both comment batches are now addressed, except one finding that needs your call.**

| finding | landed as |
| --- | --- |
| batch 1 #1 — gammainc endpoint gradient | gh#1381 -> `206ccd78` |
| batch 1 #2, #10 — quadrature batching, AnyAxisym torch construction | gh#1382 -> `06267d1d` |
| batch 1 #3, #4, #6 — coords inverse galcen transforms | gh#1383 -> `4e2cc322` |
| batch 1 #5 — `set_jit` should validate the backend | **closed as working-as-intended**, reasoning in comment 5411848615 |
| batch 2 #1 — coords (same as above) | gh#1383 |
| batch 2 #2 — `map_coordinates` grid dispatch | gh#1388 -> `107a8060` |
| batch 2 #3 — constructor jit-safety | gh#1391 -> `cfe23abe` |
| batch 2 #4 — hyp2f1 scalar params | gh#1391 -> `cfe23abe` |

Ledger delta across all five: **0**.

### Two of them were worse than the review stated, in the same way

Both `hyp2f1` and (found by auditing for the pattern) `hyp1f1` were not merely
*not traceable* — on **torch** they returned a **silently wrong gradient**.
`math.lgamma` / `math.gamma` run happily on a tensor via `__float__` and detach
that factor from the graph, so the value stays right to 1e-14, `requires_grad`
stays `True`, a `grad_fn` is set, and only grad-vs-finite-difference notices:

    d/da 2F1(a, 2; 3; -5)          -8.98e-03   against a true -8.56e-02
    TwoPowerTriaxial dPhi/dalpha    0.834      against a true  0.548
    d/da 1F1(a, a+1; -3)           -1.45e-01   against a true -1.02e-01

jax raised instead, so this was torch-only and invisible to any check short of
FD. Both now agree with scipy's finite differences (5.4e-10 for 1F1; per-route
for 2F1, since the routes differ by five orders and the tolerances say so).

### For you — one finding, and two design questions the work surfaced

1. **Finding 11 (`surfdens` torch gradient) — sharper than I first reported.**
   Not "no gradient": with `forcepoisson=True` it returns a Tensor with
   `requires_grad=True` and a `grad_fn` whose gradient is **23% wrong in R and
   72% wrong in z** (MiyamotoNagai a=0.5 b=0.1 at R=1, z=0.3; jax AD matches the
   FD reference to 1.6e-10, so the FD is right and torch is wrong). With the
   default `forcepoisson=False` it returns a bare Python float, severing the
   graph. Cause: `_quad_needs_backend` gates on `under_trace`, which is False for
   *eager* torch autograd, so scipy runs and the integrated piece detaches.
   The gate's docstring justifies scipy on the **value** (node-by-node calls keep
   scalar-only potentials working, and `_force_accepts_arrays` is undecidable) —
   which does not extend to the **gradient**. Proposal, not landed: for a
   grad-tracking eager tensor use the backend GL rule when
   `_force_accepts_arrays(pot)` and **raise** otherwise, i.e. correct or loud,
   never silently wrong. It changes behaviour for existing users, so it wants
   your decision.

2. **`hyp2f1`'s direct-Euler route has a pre-existing value-accuracy gap** the
   test grid cannot see: ~1e-6..1e-7 whenever `c - B` is not an integer
   ((1.7, 2.3, 3.4) gives 1.09e-07 at |z|=5) against ~1e-14 when it is. Every
   parameter set in the existing grid has integer `c - B`, because galpy's own
   force calls are all `c = a+1`. `X = xi^k` regularizes the `t^(B-1)` endpoint
   and nothing regularizes `(1-t)^(c-B-1)`. Raising `_NODES` only buys ~1/N
   (128 -> 1024: 1.1e-7 -> 9.4e-11); a **Gauss-Jacobi** rule would be spectral
   for any exponents and would fix the parameter-derivative accuracy at the same
   time.

3. **`actionAngleAdiabaticGrid` off-grid** is the documented "no off-grid
   fallback for jax/torch", which is what its two xfail-ledger entries encode.
   Lifting it means routing off-grid points to a per-point solve, costing the
   vectorisation the backend path exists for — the parked opt-in-fallback
   decision.

For scale on the last one: the ledger now stands at **8 xfail + 57 slow-skip**,
and all four *eager* xfail entries are by-design (two above, two qdf
unbound-classification), so from here it is a design question rather than a
burndown.

jobovy-agent:
**Corrected + complete review status.** My earlier summary (comment 5422012558) had three reporting errors, so this supersedes it. Every finding across **all three** comments is now addressed; each row below was re-verified against the reviewer's own reproduction today.

### What my previous summary got wrong

1. It said "**both** comment batches" — there are **three** comments. The standalone `quadrature.py:533` finding (comment 5400811224) was fixed in the code but **never reported**.
2. Its finding numbers were **invented**. Batch 1 has five findings, not ten; there is no `#6` or `#10`.
3. It listed the `Potential.py` surfdens gate as an unnumbered "finding 11" *open decision*. It is **batch 1 #3**, a numbered review finding, and it is now fixed.

### All findings

| finding | landed as | re-verified today |
| --- | --- | --- |
| b1 #1 `gammainc.py:196` endpoint gradient | gh#1381 → `206ccd78` | d/da vs scipy FD 1.2e-09 (both backends) |
| b1 #2 `quadrature.py:622` `finite_part_quad` array batching | gh#1382 → `06267d1d` | array `b` → shape `(2,)`, element 0 == scalar call exactly |
| b1 #3 `Potential.py:128/:817/:914` surfdens gate | gate widened to all coords; **eager-grad case** gh#1394 → `bc69771e` | `grad(surfdens wrt R)` at fixed z = **3.18e-11** vs FD |
| b1 #4 `coords.py:100/:1217/:1591` `_galcen_rot` | gh#1383 → `4e2cc322` | backend `Xsun`/`Zsun` with grad, torch + jax |
| b1 #5 `_jit.py:39` / `_input.py:179` `set_jit` | **closed working-as-intended**, comment 5411848615 | — |
| b2 #1 coordinate transforms through NumPy | gh#1383 | — |
| b2 #2 `map_coordinates` backend-prefiltered grid | gh#1388 → `107a8060` | — |
| b2 #3 constructor gradient paths not `jax.jit`-safe | gh#1391 → `cfe23abe` | 245 passed under `--backend jax --jit` and `--backend torch --jit` |
| b2 #4 `hyp2f1` not traceable in scalar params | gh#1391 → `cfe23abe` | — |
| **b3 `quadrature.py:533` finite eager torch limit with grad** | **fixed; previously unreported** | reviewer's exact repro → `tensor(0.6667, grad_fn=…)`, **d/db = 2.0 exact**; infinite limits still reach the semi-infinite branch |

Ledger delta across all of it: **0**.

### One finding was worse than reported, in a way only grad-vs-FD could see

`b1 #3` was described as "falls into SciPy and raises a concretization error" for traced `R`. True — but on **eager torch autograd** the same gate produced something worse than an exception: `forcepoisson=True` returned a Tensor with `requires_grad=True`, a `grad_fn`, and a **correct value**, whose gradient was **23% wrong in R and 72% wrong in z**; `forcepoisson=False` returned a bare Python float. `under_trace` is False for eager autograd, so SciPy ran and the integrated piece detached while the algebraic prefactor kept its graph. Fixed in gh#1394 by adding `requires_backend_grad` alongside `under_trace` — both routes now **0.0% off** in R and z on torch, 6.05e-12 on jax.

Same class in `hyp2f1`/`hyp1f1` (b2 #4): `math.lgamma` on a tensor detaches via `__float__`, so the value stayed right to 1e-14 while the parameter gradient was silently wrong. Both now match scipy finite differences.

### Still open — one follow-up this review generated

The `hyp2f1` **value-accuracy** gap is *not* a reviewer finding; I found it while fixing b2 #4 and raised it as a decision. It is the only outstanding item, and it is in progress. Root cause now identified: `_euler_labeling` guards only the `t=1` endpoint (`c-B >= 1` keeps `(1-t)^{c-B-1}` non-singular) and nothing constrains `t=0`, whose exponent is `B-1`. When `a<0` disqualifies `a`, the labeling is forced onto a tiny `B` and the fixed-order rule fails:

| B | current GL | tanh–sinh |
| --- | --- | --- |
| 0.200 | 1.67e-11 | 4.44e-16 |
| 0.100 | 9.77e-07 | 4.44e-16 |
| 0.020 | **7.18e-02** | 6.80e-07 |
| 0.002 | **7.72e-01** | 2.43e-01 |

Verified against mpmath (dps=50); scipy agrees with mpmath to 1.5e-16, so the reference is sound. Scope note, measured rather than assumed: this does **not** change DF results — `constantbetaHernquistdf.fE` agrees between numpy and the backends to **2.2e-14** at the β whose request is the 7%-wrong one. It is a numpy-parity defect, not a wrong-science one.

jobovy:
Follow-up re-review after the latest fixes.

I re-tested the previous findings against the current PR head and those fixes look good in the direct repros: `gammainc` endpoint gradients, `finite_part_quad` batching, surface-density backend gating, the `Xsun`/`Zsun` coordinate paths, backend-grid `map_coordinates`, and the constructor / `hyp2f1` parameter-gradient cases all passed the targeted stress tests I ran.

I did find two additional PR-local issues that I think should be fixed before merging:

1. `galpy/df/constantbetaHernquistdf.py:136` - the beta-gradient fix is incomplete for the external-JAX-jit contract.

   Plain `jax.grad` works now, but constructing/evaluating the DF with dynamic `beta` inside `jax.jit` still hits a Python branch on the traced value:

   ```python
   with use("jax", force=True):
       jax.jit(
           lambda beta: constantbetaHernquistdf(pot=hp, beta=beta).fE(jnp.asarray(-0.1))
       )(jnp.asarray(0.3))
   ```

   This raises `TracerBoolConversionError` at `if self._beta == 0.0`. The exact-beta branches at `0`, `0.5`, and `-0.5` probably need the same `concretely_true` treatment used in `constantbetaPowerLawdf`.

2. `galpy/backend/quadrature.py:544` - `symmetric_quad` documents `b` as `float or array` and `May be inf`, but the finite/infinite dispatch is scalar/all-or-nothing.

   With mixed finite and infinite vector limits, both JAX and torch silently return the whole-line scalar result, losing the finite element:

   ```python
   b = jnp.asarray([1.0, jnp.inf])
   symmetric_quad(jnp, lambda s: jnp.exp(-s * s), b, n=80)
   ```

   Current result: scalar `1.7724538509055159`

   Expected shape/value: `[1.49364827, 1.77245385]`

   Since this helper was added in the PR and explicitly advertises array limits, I think this should either handle mixed finite/infinite arrays correctly or reject them rather than returning a scalar for the whole input.

One audit note on the response: the small-`B` `hyp2f1` value-accuracy gap is real, and I disagree with calling it harmless in general. It is reachable through valid `TwoPowerSphericalPotential` parameters near `beta ~= 3` and can move potential/force/mass values by tens of percent or more. However, I reproduced the same bad values on `origin/feat/backends`, so I am not treating that as a #1204 regression/blocker. It should be tracked for #892.

Targeted tests run during this pass included backend quadrature/special/interpolate/coords/constructor/DF/actionAngle cases. I did not run the full suite.

jobovy-agent:
Round-3 review addressed in #1401 (into `feat/backends-develop`). Both PR-local findings fixed, and the hyp2f1 audit note turned out to be more than a tracking item.

**1. `constantbetaHernquistdf.py:136`** — fixed with `concretely_true`, as suggested. The part I checked before changing anything was that falling through to the general branch is *correct*, not just non-crashing: the exact-β closed forms are special cases of it, and agree to 2.2e-16 (β=+0.5), 6.0e-15 (β=−0.5) and 1.0e-08 (β=0, which is the hyp2f1 fallback's own accuracy at (5,1;3.5), not something the branch choice introduces). jit-vs-numpy after the fix is 1.3e-15 … 7.8e-11 across βs, and `jit(grad)` dfE/dβ matches FD. Only the backend path is guarded — on the numpy path a traced β dies one line later inside scipy anyway.

**2. `quadrature.py:544`** — handled correctly rather than rejected, since it's reachable from the public API and not only the helper: `surfdens(R=1, z=[1.0, inf])` under grad-tracking torch returned a **0-d scalar** `0.059445` for a two-element request. Now `[0.05917760 0.05944502]`, matching numpy to 6.7e-16 with the gradient intact. Both rules are evaluated and selected elementwise; infinite entries get a dummy finite limit so `abs(inf)` nodes can't NaN-poison the gradient of the entries that do take the finite branch.

**3. On the audit note — you were right, and it was worse than I'd written.** I checked your `TwoPowerSphericalPotential` case on my own branch and **the tanh–sinh commit did not fix it**: β=3.01 was still 8.4e-02 off in Φ. That case requests `hyp2f1(0.01, 2.01; 2.01; z)`, i.e. B = 0.01 — exactly the floor my commit message had called "not characterized".

It isn't quadrature error. I formed `t` explicitly, and at B = 0.01 `t**B` is still 1e-3 where `t ~ 1e-300`, so the integrand matters until `t ~ exp(-3684)` — far below what a double holds. `t` underflows to 0, `t**B` with it, and the tail is silently discarded. That's precisely why the error looked insensitive to both node count and half-width: widening the range is useless while `t` has already underflowed. Fixing it needs log-space powers **and** a B-scaled half-width; either alone changes nothing.

| B | shipped GL | first tanh–sinh | now |
|---|---|---|---|
| 0.200 | 1.8e-11 | 5.6e-16 | 4.4e-16 |
| 0.020 | 7.2e-02 | 6.8e-07 | 8.9e-16 |
| 0.010 | ~1e-01 | 8.4e-04 | 4.4e-16 |
| 0.001 | ~5e-01 | 4.9e-01 | 1.6e-15 |

So the small-B gap is closed here rather than deferred to #892.

**One thing I did not do, with the measurement, in case you want to weigh in.** Sweeping 768 (A,B,c,z) cases with B *above* the 0.25 threshold: GL's worst is **1.5e-04**, tanh–sinh's worst is **2.9e-15**, and there is **not one case** where tanh–sinh is worse. The threshold is only protecting the less accurate rule. I left it alone because the traced-parameter path also routes through `_euler_quad` and tanh–sinh there needs a fixed conservative grid (it can't size from a tracer) — its own design plus measurement, and it would let the `X = xi^k` grading machinery be deleted rather than left dead. Proposing it as the next PR.

Two defects in my own new code, both surfaced by widening the test axis rather than the happy path: `xp.maximum(x, 0.0)` (torch rejects a Python scalar — invisible to jax-only probes), and `_tanh_sinh_grid` reaching for `float(B)`, which raises under `jax.grad` where the parameters have concrete truth values but are still tracers. That route also had **no parameter-gradient coverage at all** — the silent-detachment shape — so I added it: d/da, d/db, d/dc match FD to 3.4e-10 … 1.8e-07 on both backends.

Gates: numpy byte-identical (sha256 `245bc0e3…` over 315 values on both arms, arms grep-asserted to differ), 385 passed across the three touched suites, ledger delta 0.

jobovy:
Follow-up after #1401/#1402.

The two issues from my previous comment look fixed in the current PR head: dynamic `constantbetaHernquistdf(beta=...)` under external `jax.jit` now matches the NumPy path in my repros, and mixed finite/infinite `symmetric_quad` limits now return the correct vector shape and values for both JAX and torch. The TwoPower small-`B` value cases I flagged are also fixed by the tanh-sinh rewrite.

I found one new blocker in #1402:

1. `galpy/backend/special/_fallback/hyp2f1.py:212` - the traced Euler-label selector still uses the old `c - B >= 1` rule.

   #1402 relaxes the concrete admissibility rule from `c - B >= 1` to `c - B > 0`, but `_nonpositive_traced` still does:

   ```python
   oka = (ea > 0) & ((c - ea) >= 1.0)
   ```

   That rejects a valid first label under tracing when `0 < c-a < 1`, and can make the traced path choose the second label even when that one is nonpositive. Concrete/eager is correct; traced/JIT returns `nan` or `inf`.

   Minimal repro:

   ```python
   jax.jit(lambda a: gsp.hyp2f1(a, -0.2, 2.5, jnp.asarray(-0.7)))(jnp.asarray(2.0))
   ```

   Expected/scipy: `1.09213476189247`

   Current JAX traced result: `inf`

   Another case:

   ```python
   jax.jit(lambda a: gsp.hyp2f1(a, -1.0, 2.5, jnp.asarray(-0.7)))(jnp.asarray(2.0))
   ```

   Expected/scipy: `1.56`

   Current JAX traced result: `nan`

   The fix is probably to make the traced selector match the new rule, i.e. `((c - ea) > 0.0)`, and add a traced-parameter test with a case such as `(a, b, c) = (2.0, -0.2, 2.5)`.

While touching that file, I would also clean the stale comments that still describe the old GL machinery / `c-B >= 1` guarantee, because those comments now disagree with the code after #1402.

Targeted tests I reran: focused `test_backend_special`, `test_backend_quadrature`, and `test_backend_constantbetadf` cases. I did not run the full suite.

jobovy-agent:
Blocker fixed in #1403 (into `feat/backends-develop`). You're right, and thank you — this is a real one.

**The miss.** #1402 relaxed the admissibility rule in *three* places (`_in_regime`, `_euler_labeling`, `_in_regime_mask`) and there is a **fourth** copy inside `_nonpositive_traced`. My audit grep was `c - a) >= 1.0`; that site spells the same predicate with different variable names, `(c - ea) >= 1.0`, so it didn't match. I've re-swept with a pattern that cannot care about the names (`\(c - [a-z_]+\)|>= 1\.0`) — that was the only remaining site.

Reproduced both of your cases and two more before touching anything:

| (a, b, c, z) | scipy | traced BEFORE | traced AFTER |
|---|---|---|---|
| (2.0, −0.2, 2.5, −0.7) | 1.09213476189 | `inf` | rel 0.0 |
| (2.0, −1.0, 2.5, −0.7) | 1.56 | `nan` | rel 1.1e-16 |
| (2.0, −0.5, 2.2, −0.7) | 1.27837324696 | `inf` | rel 2.4e-15 |
| (3.0, −0.2, 3.4, −2.0) | 1.2241297706 | `inf` | rel 3.6e-15 |

The test uses parameters with `0 < c-a < 1` **and** `b <= 0` specifically, since that is what makes the mismatch fatal rather than merely suboptimal — the first label is the only admissible one, so wrongly rejecting it lands on a non-positive B.

**On the stale comments — worth flagging, because one wasn't cosmetic.** Cleaning them turned up an assumption #1402 had silently broken:

> "… and `_euler_labeling` guarantees c-B >= 1, so B is the binding one."

Under `c - B > 0` the t=1 exponent can now be the *small* one, which it never could before, so that sentence no longer justified the fixed half-width. I measured instead of assuming: `c - B` down to **1e-4 is accurate to 2.3e-15**, because the tanh–sinh rule is symmetric in u — a half-width that resolves a small `B` resolves a small `c-B` identically. So the code was fine, but for a different reason than the comment claimed. The comment now says the true reason and a test pins it. The module header's two-GL-substitutions description and the `xi^k`/cap-12/0.25-threshold block are deleted outright.

Gates: numpy byte-identical (sha256 `245bc0e3…` vs develop `8c320b39`, arms grep-asserted to differ), 391 passed / 1 skipped, ledger delta 0.

Separately, on your earlier audit note: I re-measured the torch `surfdens(forcepoisson=True)` gradient on current develop and it now matches finite differences to **3.2e-11 (R) / 7.7e-11 (z)** — #1394 closed that one, so nothing is outstanding there either.

jobovy-agent:
On the remaining cleanup item (stale comments describing the deleted
Euler-transform route and the old Gauss-Legendre/substitution behaviour):
addressed in #1406.

Agreed it matters — the comments are the audit trail for this work, so one that
describes a route which no longer exists is a real defect even though nothing
executes it. Two were worse than stale: they named `_TS_B_MAX`, a constant with
**zero occurrences** in `hyp2f1.py`, telling a reader the tanh-sinh rule "is
routed in below `_TS_B_MAX`". There is one rule for every B now; there is no
threshold to be below.

Rather than fix only the two spots named, I swept both files for
`three routes | route 2 | _TS_B_MAX | euler-transform | Gauss-Legendre |
substitut` and classified every hit as present-tense-wrong vs correctly-framed
history. Eight were wrong and are fixed:

- `hyp2f1.py` — the "Three routes" docstring (now two, with the transform's
  removal explained and pointed at its proof); the note justifying why Pfaff
  could not serve as route 2 (dropped — it defends a deleted design, and its
  stated reason cites substitutions the tanh-sinh rule does not use); "the same
  three routes" in `_nonpositive_traced`; "the three routes named above" in
  `_hyp2f1_nonpositive`.
- `tests/test_backend_special.py` — the `"euler-transformed"` route label (those
  parameters take the DIRECT route now); both `_TS_B_MAX` references; "all three
  routes"; and the per-route tolerance rationale, which credited Gauss-Legendre
  exactness for integer `c - B` when GL no longer runs.

Left alone deliberately: the passages already written as history ("This REPLACED
a fixed-order Gauss-Legendre rule…", "when the bound was `c - B >= 1`", "the
large-B cases the GL rule was kept for"). Those are accurate as written and are
the audit trail worth keeping.

Comments and pytest ids only — no executable change on any path. Neither ledger
keys off the renamed ids (checked). Gate: `test_backend_special.py` 253 passed.

PR #1233, opened 2026-08-01
coords: backend-agnostic Jacobian helpers (4 of 6; keystone for differentiable stream covariance)

jobovy-agent:
Keystone for the streamTrack sky-basis `cov()` work — same shape as Streams PR-1
(migrate the coords leaf first, then the consumer).

Four `*_jac` helpers made backend-agnostic. Each was pure numpy: a
`numpy.zeros((6,6))` filled by item assignment, which jax cannot do (immutable
arrays) and which forces one matrix per call. Rows are now built with
`xp.stack`, so batched inputs give a `(..., 6, 6)` stack for free.

| helper | numpy byte-identical | torch | jax x64 | jit | grad-vs-FD |
|---|---|---|---|---|---|
| `galcenrect_to_galcencyl_jac` | 400/400 (+ R=0 edge) | 0.0 | 0.0 | yes | 4.6e-11 |
| `galcenrect_to_XYZ_jac` | 600/600 (3-/6-arg, ±Xsun) | 0.0 | 0.0 | yes | 1.4e-11 |
| `lbd_to_XYZ_jac` | 800/800 (4 call forms) | 0.0 | 0.0 | yes | 6.5e-11 |
| `XYZ_to_lbd_jac` | 800/800 (+ r=0 pole) | 6.9e-17 | 0.0 | yes | 6.1e-11 |

`tests/test_coords.py`: 97 passed. Gradients are grad-vs-FD with h-convergence
(not finite-and-nonzero), checked at float64 — jax's float32 default makes an FD
check *diverge* as h shrinks and is not evidence of anything.

**The if -> array-input class** (what this PR is really about):
* `if R == 0.0: R = 1e-30` -> `xp.where`
* `if cb != 0.0` / `if KD != 0.0` / `if r == 0.0` -> `xp.where`, each dividing by
  a **safe** denominator first. Eager and traced `where` evaluates both branches,
  so a raw 0/0 in the dead branch NaN-poisons the gradient even where the value
  is selected away.
* `l *= _DEGTORAD` mutated the **caller's** array in place -> out-of-place.
* Static branches (`len(args)`, `degree=`) stay Python-level: arity/flag
  decisions, not data-dependent.

**Known and documented:** the gradient at the exact celestial pole (r=0) is NaN,
from `sqrt(0)` having infinite derivative, not from the guards. dl/dX genuinely
does not exist there, so NaN is the honest signal (same call as RazorThin
R2deriv off-plane). Verified **element-local**: a pole element in a batch does
not poison its neighbours.

**Scope, stated rather than quietly narrowed:** the remaining two helpers
(`galsky_to_sky_jac`, `sky_to_customsky_jac`) each depend on a numpy-only coords
leaf — `lb_to_radec` (74 lines, 32 numpy refs, 0 xp) and
`custom_to_radec`/`radec_to_custom`. Finishing them means migrating those leaves,
which is a follow-up PR rather than something to smuggle in here or fake by
duplicating their logic inline. Numpy path is byte-identical throughout, so
nothing regresses in the meantime.

### DIFF
--- a/galpy/util/coords.py
+++ b/galpy/util/coords.py
@@ -1972,79 +1972,117 @@ def XYZ_to_lbd_jac(*args, **kwargs):
     else:
         raise ValueError("XYZ_to_lbd_jac expects 3 or 6 arguments")
 
+    xp = get_namespace(*args, xp=kwargs.get("xp", None))
+    X, Y, Z = promote_scalars(xp, X, Y, Z)
+    if with_vel:
+        vX, vY, vZ = promote_scalars(xp, vX, vY, vZ)
     R2 = X * X + Y * Y
-    D = numpy.sqrt(R2 + Z * Z)
-    r = numpy.sqrt(R2)
+    D = xp.sqrt(R2 + Z * Z)
+    r = xp.sqrt(R2)
     cb = r / D
     sb = Z / D
     # On the celestial pole (r==0) ``l`` is undefined; pick (cl, sl) = (1, 0)
     # so the spatial inverse stays finite (consistent with atan2(0, 0) = 0).
-    if r == 0.0:
-        cl, sl = 1.0, 0.0
-    else:
-        cl = X / r
-        sl = Y / r
+    # Each xp.where's dead branch is computed too (eagerly and under a trace),
+    # so divide by a safe denominator or the 0/0 NaN poisons the gradient.
+    at_pole = r == 0.0
+    r_safe = xp.where(at_pole, 1.0, r)
+    cl = xp.where(at_pole, 1.0, X / r_safe)
+    sl = xp.where(at_pole, 0.0, Y / r_safe)
 
-    out = numpy.zeros((6, 6) if with_vel else (3, 3))
-    # Position 3x3: ∂(l, b, D)/∂(X, Y, Z)
-    if cb != 0.0:  # else: at the pole, ∂l/∂(X, Y) blows up — leave as zero
-        out[0, 0] = -sl / (D * cb)
-        out[0, 1] = cl / (D * cb)
-    out[1, 0] = -sb * cl / D
-    out[1, 1] = -sb * sl / D
-    out[1, 2] = cb / D
-    out[2, 0] = cb * cl
-    out[2, 1] = cb * sl
-    out[2, 2] = sb
+    o = xp.zeros_like(cb)
+    # Position 3x3: ∂(l, b, D)/∂(X, Y, Z). At the pole ∂l/∂(X, Y) blows up, so
+    # those two entries stay zero (same convention as the scalar version).
+    Dcb = D * cb
+    Dcb_safe = xp.where(Dcb == 0.0, 1.0, Dcb)
+    dl_dX = xp.where(Dcb == 0.0, 0.0, -sl / Dcb_safe)
+    dl_dY = xp.where(Dcb == 0.0, 0.0, cl / Dcb_safe)
+    P = xp.stack(
+        [
+            xp.stack([dl_dX, dl_dY, o], axis=-1),
+            xp.stack([-sb * cl / D, -sb * sl / D, cb / D], axis=-1),
+            xp.stack([cb * cl, cb * sl, sb], axis=-1),
+        ],
+        axis=-2,
+    )
+
+    def _degree_scale(m):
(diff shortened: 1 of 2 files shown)


jobovy-agent:
**Finding while adding coverage: a pre-existing Jacobian/forward-transform mismatch (not from this PR).**

codecov flagged `galcenrect_to_XYZ_jac` lines as uncovered. Root cause: that helper had **no test at all** (0 references in `test_coords.py`). Added one — and it failed.

It is not the test's tolerance. The disagreement is invariant to the finite-difference step, which rules out truncation:

| forward transform | max`|J - J_fd|` |
|---|---|
| `_extra_rot=True` (**the default**) | 1.469e-06 |
| `_extra_rot=False` | 1.902e-10 |

`galcenrect_to_XYZ_jac` differentiates the `_extra_rot=False` transform, while `galcenrect_to_XYZ` defaults to `_extra_rot=True`. The finite differences show non-zero couplings exactly where the analytic Jacobian has structural zeros — an unmodelled rotation.

**This predates the backend migration.** The formula is untouched here; the restructure is byte-identical over 600 random states. But it matters in practice: `streamTrack._analytical_jacobian` calls

```python
J1 = coords.galcenrect_to_XYZ_jac(x, y, z, vx, vy, vz, Xsun=ro, Zsun=zo)
XYZ_mean = coords.galcenrect_to_XYZ(x, y, z, Xsun=ro, Zsun=zo)   # default _extra_rot=True
```

so sky-frame covariance transforms inherit a ~1.5e-6 inconsistency.

**Not fixed here**, deliberately: changing the formula would break the byte-identical numpy guarantee this PR rests on, and the choice between *aligning the Jacobian with the default* versus *making callers pass `_extra_rot=False`* is yours. The new test pins the contract the function actually implements (1.9e-10) and names the discrepancy in a comment rather than blessing it silently.

jobovy-agent:
**Follow-up: the `_extra_rot` mismatch has an exact one-line correction.**

`_extra_rot` applies `galcen_extra_rot`, a **constant** orthogonal matrix (added to match astropy's Galactocentric frame; `max|R - I| = 1.47e-06`, `det = 1`). `galcenrect_to_XYZ` rotates its *output* by `R.T`, so the Jacobian of the default transform is exactly

```
J_true = blockdiag(R.T, R.T) @ J_analytic
```

Verified against finite differences of the default forward transform, sweeping the step to separate signal from roundoff:

| FD step scale | max`|R6.T @ J - J_fd|` |
|---|---|
| 1e-4 | 9.609e-12 |
| 1e-5 | 9.917e-11 |
| 1e-6 | 2.244e-09 |
| 1e-7 | 1.603e-08 |

The residual *grows* as h shrinks — roundoff-dominated, i.e. the identity is exact and the best estimate of the true error is ~1e-11.

So the choice is a genuine one, not a guess:
1. **Make the Jacobian match the default** — return `blockdiag(R.T, R.T) @ J`. Correct for every caller that uses default arguments, including `streamTrack._analytical_jacobian`. Changes `galcenrect_to_XYZ_jac`'s numpy output by ~1.5e-6, so it is **not** byte-identical and does not belong in this PR.
2. **Make callers pass `_extra_rot=False`** — keeps the helper byte-identical, but every caller must remember, and the pairing is easy to get wrong (it already was).

I have not implemented either: option 1 breaks the byte-identity guarantee #1233 rests on, and the choice is yours. Flagging that option 1 looks right to me — the helper's job is to differentiate the function as callers actually call it.

jobovy-agent:
**Impact of the `_extra_rot` mismatch, quantified — so the priority is clear.**

Propagated a realistic stream-like covariance (30 pc positions, 2 km/s velocities, random correlations) through both Jacobians:

| quantity | value |
|---|---|
| max relative difference in the covariance | 2.1e-06 |
| relative difference in 1-sigma widths | 1.1e-06 |
| position sigma | 0.029967 kpc either way |
| on-sky effect at D = 20 kpc | 1.4e-04 arcsec |

So: **real, but numerically negligible for any practical stream** — 0.1 mas on a track whose widths are arcminutes. This is a correctness fix, not an urgent one, and nothing already published is meaningfully affected.

That also means there is no pressure to rush it into this PR at the cost of the byte-identity guarantee. Recording it so the decision can be made on correctness grounds rather than on fear that results are wrong.

PR #1235, opened 2026-08-01
streamTrack: backend-native cov() basis rotation (galcenrect/galcencyl differentiable; sky bases eager-only)

jobovy-agent:
`cov(basis=...)` raised `NotImplementedError` for anything but `galcenrect` on a
backend track. The guard was not the real constraint — the whole Jacobian chain
was pinned to numpy by **20 `float()` casts**, and the guard just documented that.
Removing them makes the sky bases work on jax/torch and keeps the covariance
differentiable through the basis rotation.

### What changed

- **`_cart_mean_at`** — 6 `float()` casts → one `xp.stack`. This was the source
  of the pin; everything downstream inherited it.
- **`_analytical_jacobian`** — 14 `float()` casts on the traced state dropped.
  `ro`/`zo` stay floats: they are unit config, not traced data.
- **`cov()`** — the backend and numpy branches duplicated the entire tail
  (`use_phys` scaling, basis rotation, scalar unwrap), so the rotation is now one
  shared block. The Jacobian is stacked and applied as a single `(len, 6, 6)`
  matmul rather than assigning into `out[k]` — the in-place write is what made
  the old path invalid for immutable backend arrays.
- **`coords._scale_angle_rows`** — the rad→deg row scaling existed as a closure
  in `XYZ_to_lbd_jac` and was about to be copied three more times into
  `streamTrack`. Hoisted next to `_DEGTORAD` and reused, so there is one copy
  instead of four.

### Verification

numpy path **byte-identical**, worst `|diff| = 0.000e+00`, across all four bases
plus the out-of-range NaN case — against a reference captured *before* any edit,
so the migration is checked against a reference rather than against itself.

| suite | result |
|---|---|
| `test_streamTrack.py` + `test_coords.py`, numpy | 143 passed |
| `test_streamTrack.py`, torch | **45/45** (was 40/45) |
| `test_quantity.py`, numpy | 244 passed |
| `test_quantity.py`, torch (full file) | 6 fail, all pre-existing `sphericaldf`/`streamgapdf`, none streamTrack |

### Ledger: 101 → 96

All five removed entries were **genuine** failures, not stale ones. The baseline
junit from the all-backend run at `618fc6c9` shows 51 streamTrack-named torch
tests passing and exactly 6 xfailing: these 5, plus `test_streamTrack_multi_parallel`,
which stays — it is the `parallel_map` torch+fork deadlock, unrelated here.

Four failed on the `NotImplementedError`; the fifth on `numpy.all()` over a
Tensor at an assertion site.

### Deliberately not in scope

`galsky_to_sky_jac` / `sky_to_customsky_jac` are still scalar-only — they
`reshape(..., ())` internally, a pre-existing property. So the per-`tp` Jacobian
assembly stays a comprehension feeding one batched matmul. Making those two
broadcast (which would remove the last Python loop here) is a clean follow-up.

### DIFF
--- a/galpy/df/streamTrack.py
+++ b/galpy/df/streamTrack.py
@@ -2340,6 +2340,7 @@ class StreamTrack:
             dev = device_of(self._cov_xyz)
             tp_arr = numpy.atleast_1d(numpy.asarray(tp, dtype=float))
             tp_b = asarray_on_device(xp, tp_arr, dev)  # backend query axis
+            tp_for_jac = tp_b  # physical tp (pre-normalization) for the Jacobian
             if self._tp_scale is None:
                 in_range = asarray_on_device(xp, self._in_range(tp_arr), dev)
             else:
@@ -2364,25 +2365,18 @@ class StreamTrack:
                     dev,
                 )
                 out = out * (scale[:, None] * scale[None, :])  # NaN · scale = NaN
-            if basis != "galcenrect":
-                raise NotImplementedError(
-                    "cov() on a backend (jax/torch) track supports only "
-                    "basis='galcenrect'; sky-frame covariance bases are a "
-                    "numpy-only follow-up."
-                )
-            if numpy.isscalar(tp) or (hasattr(tp, "ndim") and tp.ndim == 0):
-                return out[0]
-            return out
-        tp_arr = numpy.atleast_1d(tp)
-        in_range = self._in_range(tp_arr)
-        out = numpy.empty((len(tp_arr), 6, 6))
-        for a in range(6):
-            for b in range(6):
-                vals = numpy.interp(tp_arr, self._tp_grid, self._cov_xyz[:, a, b])
-                out[:, a, b] = numpy.where(in_range, vals, numpy.nan)
-        if use_phys:
-            scale = numpy.array([ro_use, ro_use, ro_use, vo_use, vo_use, vo_use])
-            out = out * numpy.outer(scale, scale)  # NaN · scale = NaN
+        else:
+            tp_arr = numpy.atleast_1d(tp)
+            tp_for_jac = tp_arr
+            in_range = self._in_range(tp_arr)
+            out = numpy.empty((len(tp_arr), 6, 6))
+            for a in range(6):
+                for b in range(6):
+                    vals = numpy.interp(tp_arr, self._tp_grid, self._cov_xyz[:, a, b])
+                    out[:, a, b] = numpy.where(in_range, vals, numpy.nan)
+            if use_phys:
+                scale = numpy.array([ro_use, ro_use, ro_use, vo_use, vo_use, vo_use])
+                out = out * numpy.outer(scale, scale)  # NaN · scale = NaN
 
         if basis != "galcenrect":
             # When use_phys=True we thread the resolved ro/vo through the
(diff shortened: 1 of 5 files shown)


jobovy-agent:
**jax verification** (the suite can't cover this — `jax tests/test_streamTrack.py` is a pre-existing file-level entry in `backend_slow_skip.txt`, eager-XLA blows the shard timeout — so checked directly):

```
jax galcencyl   vs numpy-rotated-same-track: max rel = 6.306e-16
jax sky         vs numpy-rotated-same-track: max rel = 6.410e-15
jax galsky      vs numpy-rotated-same-track: max rel = 7.953e-16
```

Same track, same `tp`; the reference rotates `cov(galcenrect)` by the Jacobians in plain numpy. All four bases are finite and symmetric to `max|C-C^T| <= 1.4e-20` against `|C|max ~ 1.8e-4`. Run under `JAX_ENABLE_X64=1`, CPU-forced.

jobovy-agent:
**Correction + scope note on the sky bases** (found by running `tests/test_backend_streamTrack.py`, which I had not run when I opened this).

That file asserted `cov(basis=...)` *raises* on a backend track, so it had to be updated. Doing so surfaced a real distinction I had glossed:

- **Forced backend** (`backend.use("torch", force=True)`) — how the entire all-backend CI suite runs: all four bases return backend arrays and match numpy. Verified on a `streamspraydf` track for `galcenrect`/`galcencyl`/`sky`/`galsky`.
- **Unforced / mixed** — backend arrays built explicitly with the ambient backend left as numpy (what `test_backend_streamTrack.py` does): `galcenrect` and `galcencyl` are backend and exact (`0.0` / `4.2e-16`), but `sky`/`galsky` raise `TypeError: Multiple namespaces`.

Root cause is *not* in `streamTrack`: `coords.galcenrect_to_XYZ` and `galcenrect_to_vxvyvz` are still plain `@scalarDecorator` without the `@backendNative` marker from #1234, so they promote their inputs through numpy. Meanwhile `galcenrect_to_XYZ_jac` *is* backend-native, so the two halves of the chain disagree.

The new test is scoped to what unforced mode supports; the sky bases stay covered under forced backend by `tests/test_streamTrack.py` (45/45 on torch). `tests/test_backend_streamTrack.py`: **62/62** on jax+torch.

Marking those two coords functions `@backendNative` is the natural follow-up — it closes the unforced case and is what makes the sky chain differentiable end-to-end. It is a coords migration with its own caller sweep (the einsum has an array-`Xsun` branch), so it belongs in its own PR rather than bolted on here.

jobovy-agent:
**Gradient check — correcting a claim I made in the commit message.**

I wrote that this "keeps the covariance differentiable through the basis rotation". Verified grad-vs-FD (torch float64, random contraction `W`, random unit direction `D` on `_track_xyz`, central differences):

```
galcencyl  AD directional deriv =  1.5059791297e-06
    h=1e-04  FD =  1.5059791326e-06   rel err = 1.89e-09
    h=1e-05  FD =  1.5059791576e-06   rel err = 1.47e-08
    h=1e-06  FD =  1.5059788269e-06   rel err = 2.05e-07
```

(the rel err growing as h shrinks is FD roundoff, not a code problem.)

`sky` and `galsky` do **not** get that far — they raise:

```
RuntimeError: Can't call numpy() on Tensor that requires grad
  coords.py:167 in scalar_wrapper  ->  xp.asarray(args[0])   # xp is numpy
  streamTrack.py:2478              ->  coords.galcenrect_to_XYZ(x, y, z, ...)
```

Same root cause as the unforced-mode `TypeError` above: `galcenrect_to_XYZ` / `galcenrect_to_vxvyvz` are not `@backendNative`, so `scalarDecorator` promotes through numpy — which a `requires_grad` tensor refuses outright.

**So, accurately, this PR delivers:** the `NotImplementedError` guard removed; `_cart_mean_at` and `_analytical_jacobian` freed of 20 `float()` casts; `galcenrect` and `galcencyl` backend-native **and differentiable**; `sky`/`galsky` correct **eagerly under a forced backend** (matching numpy to 1e-9 on torch, 6e-15 on jax) but not yet differentiable and not yet safe unforced. Title updated to say so.

Making those two coords functions `@backendNative` is the next PR and is what finishes the sky chain.

PR #1245, opened 2026-08-03
AnyAxisymmetricRazorThinDisk: fix the small-|z| force/2nd-derivative band (exact-complement K + a=R peak resolution)

jobovy-agent:
**Pre-existing numpy/scipy bug, not a backend one.** Targets `main`.

> **Note:** this PR's original title and description said the fix was to *clamp* the elliptic parameter. That turned out to be wrong — see "What changed since the first version" at the bottom. The clamp was the bug, not the cure.

## The defect

The five integrands form `m = 4aR/((a+R)^2+z^2)` and hand it to scipy's `ellipk`/`ellipe`. Analytically `m <= 1`, but for tiny `|z|` rounding tips it just above 1, where scipy returns nan. The first version of this PR clamped `m` to `1 - 1e-15`.

Two things were wrong with that:

**1. The clamp manufactures infinities.** `numpy.minimum(m, 1.0)` maps `m` to *exactly* 1.0 once rounding tips it, and `ellipk(1.0) = inf`. `R2deriv` and `z2deriv` returned `-inf`/`+inf` at isolated `z` (1e-8, 1e-10) with correct values on either side. Clamping strictly below 1 censors the log singularity instead: at `1-m = 1e-16` it returns `K = 18.656` against a true `19.807`, saturating there (16% low by `1e-18`).

**2. The `a=R` peak is not resolved.** `zforce`'s integrand has a peak of width `~|z|` at `a=R`. `quad` subdivides from panels of order `R`, so once `|z| << R` it steps over the peak entirely: `zforce(0.5, 1e-8)` returned `+1.7e-8` instead of `-1.86`.

## The fix

**Take K from its exact complement.** No subtraction is involved:

```
1 - m = 1 - 4aR/((a+R)^2+z^2) = ((a-R)^2 + z^2)/((a+R)^2 + z^2) = d2/aRz
```

which the integrands already build. `special.ellipkm1(d2/aRz)` is exact, never infinite, and the clamp is *deleted* rather than relocated. That alone makes `R2deriv` and `z2deriv` finite at every `|z|`.

**Resolve the peak where a `z` prefactor pays for it.** Substituting `a = R + |z|t` maps the peak to unit width and makes `(a-R)^2+z^2` become `z^2(1+t^2)` analytically; `da = |z|dt` cancels the `1/|z|` the peak contributes, against the `-4z` (zforce) and `-2z` (Rzderiv) prefactors. Applied to those two only — the others have no such prefactor.

`Rzderiv` additionally needed its `E` and `K` coefficients rewritten in `u = a-R`, the point they are built about (the `E` one vanishes at `(a=R, z=0)`, the `K` one factors through `d2`), so neither is formed by cancelling large terms.

## Verification

Against analytic limits, at two radii:

| quantity | limit | before | after |
|---|---|---|---|
| `zforce(z->0+)` | `-2 pi Sigma` = -1.8646 | `+1.7e-8` | -1.8646 |
| `Rzderiv(z->0+)` | `-2 pi Sigma'` = -5.5939 | wrong below ~1e-5 | O(z) convergence 1e-3 -> 1e-8 |
| `R2deriv`, `z2deriv` | — | `+-inf` at isolated z | finite everywhere |

`__call__` and `Rforce` were **never broken** and are untouched.

**Known accuracy floor, documented in the source:** `Rzderiv` is a `1/z`-sized residual of `1/z^2`-sized pieces cancelling by oddness about `a=R`, so its relative accuracy floors at `~eps/|z|` — good to ~1e-8 at `|z|=1e-8`, degrading below. Cancelling analytically would need `Sigma'(R)`, unavailable for a user-supplied callable.

## Test change, and why

`test_2ndDeriv_potential` now excludes this potential at `z=0`, exactly as it already excludes `KuzminDiskPotential`: `Fz` jumps from `+2 pi Sigma` to `-2 pi Sigma` across a razor-thin disk, so `dFz/dz` is a delta function there and the difference quotient has nothing finite to converge to.

That check previously passed **for the wrong reason**: the broken `zforce` returned exactly `-z2deriv(R,0) * z` near the plane, which its own finite difference reproduced. It was validating a broken `zforce` against `z2deriv` and agreeing with it. The physics is still asserted, by a small-`z` limit test checking `Fz -> -2 pi Sigma`.

## Numpy values change

This is a numpy-path bugfix, so values change where they were wrong (nan/inf/orders-of-magnitude). Switching `ellipk(m)` to `ellipkm1(1-m)` also perturbs the last bits at ordinary `z`, since it is a different scipy routine. Not byte-identical, deliberately.

## What changed since the first version

The original commit clamped `m` and added a regression test asserting `Fz/z` is constant near the plane — which is wrong physics (`Fz` tends to a *constant*, not a constant times `z`). Both are replaced here.

Full `tests/test_potential.py`: 8 failures, identical to pristine `main` (this worktree has no `libgalpy.so`).

### DIFF
--- a/galpy/potential/AnyAxisymmetricRazorThinDiskPotential.py
+++ b/galpy/potential/AnyAxisymmetricRazorThinDiskPotential.py
@@ -120,7 +120,10 @@ class AnyAxisymmetricRazorThinDiskPotential(Potential):
         def rforceint(a):
             a2 = a**2
             aRz = (a + R) ** 2.0 + z2
-            faRoveraRz = 4 * a * R / aRz
+            # m = 4aR/((a+R)^2+z^2) <= 1 analytically (equality at a=R, z=0), but
+            # rounding tips it just above 1 for tiny |z|, and scipy's ellipe/ellipk
+            # return nan there. Fires only on that artifact, so values are unchanged.
+            faRoveraRz = numpy.minimum(4 * a * R / aRz, 1.0)
             return (
                 a
                 * self._sdens(a)
@@ -146,7 +149,10 @@ class AnyAxisymmetricRazorThinDiskPotential(Potential):
 
         def zforceint(a):
             aRz = (a + R) ** 2.0 + z2
-            faRoveraRz = 4 * a * R / aRz
+            # m = 4aR/((a+R)^2+z^2) <= 1 analytically (equality at a=R, z=0), but
+            # rounding tips it just above 1 for tiny |z|, and scipy's ellipe/ellipk
+            # return nan there. Fires only on that artifact, so values are unchanged.
+            faRoveraRz = numpy.minimum(4 * a * R / aRz, 1.0)
             return (
                 a
                 * self._sdens(a)
@@ -172,7 +178,10 @@ class AnyAxisymmetricRazorThinDiskPotential(Potential):
         def r2derivint(a):
             a2 = a**2
             aRz = (a + R) ** 2.0 + z2
-            faRoveraRz = 4 * a * R / aRz
+            # m = 4aR/((a+R)^2+z^2) <= 1 analytically (equality at a=R, z=0), but
+            # rounding tips it just above 1 for tiny |z|, and scipy's ellipe/ellipk
+            # return nan there. Fires only on that artifact, so values are unchanged.
+            faRoveraRz = numpy.minimum(4 * a * R / aRz, 1.0)
             return (
                 a
                 * self._sdens(a)
@@ -206,7 +215,10 @@ class AnyAxisymmetricRazorThinDiskPotential(Potential):
         def z2derivint(a):
             a2 = a**2
             aRz = (a + R) ** 2.0 + z2
-            faRoveraRz = 4 * a * R / aRz
+            # m = 4aR/((a+R)^2+z^2) <= 1 analytically (equality at a=R, z=0), but
+            # rounding tips it just above 1 for tiny |z|, and scipy's ellipe/ellipk
+            # return nan there. Fires only on that artifact, so values are unchanged.
+            faRoveraRz = numpy.minimum(4 * a * R / aRz, 1.0)
             return (
                 a
                 * self._sdens(a)
@@ -233,7 +245,10 @@ class AnyAxisymmetricRazorThinDiskPotential(Potential):
         def rzderivint(a):
             a2 = a**2
             aRz = (a + R) ** 2.0 + z2
-            faRoveraRz = 4 * a * R / aRz
+            # m = 4aR/((a+R)^2+z^2) <= 1 analytically (equality at a=R, z=0), but
+            # rounding tips it just above 1 for tiny |z|, and scipy's ellipe/ellipk
+            # return nan there. Fires only on that artifact, so values are unchanged.
+            faRoveraRz = numpy.minimum(4 * a * R / aRz, 1.0)
             return (
                 a
                 * self._sdens(a)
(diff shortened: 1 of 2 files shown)


jobovy-agent:
**Converting to draft — I got this wrong and caught it before merge.**

The clamp is necessary but **not sufficient**, and the regression test I wrote asserts the wrong physics.

For a razor-thin disk, `Fz(z -> 0+)` approaches **`-2*pi*Sigma(R)`**, a *constant* — not `O(z)`. At R=0.5, `Sigma = 0.3346952402`, so the limit is `-2.102952216`. Measured on this branch:

```
 |z|      _zforce_numpy      ratio to -2*pi*Sigma
1e-04    -2.10276080633           0.999909
1e-05    -2.10293307765           0.999991
3e-06    -2.10294647438           0.999997
1e-06     1.9141608859e-06       -9.10e-07     <-- collapses, wrong sign
1e-08     1.91377857829e-08      -9.10e-09
1e-20     1.9137792147e-20       -9.10e-21
```

Below ~1e-6 scipy's adaptive quad misses the width-`z` peak of `1/((a-R)^2+z^2)` entirely and returns ~0. So the clamp converts a nan into a **silently wrong** answer, which is arguably worse than the nan.

And my test asserted `Fz/z` is constant across the band — which only holds *because* the values are garbage. It passes for the wrong reason; a correct implementation would fail it.

Correct behaviour to assert: `Fz -> -2*pi*Sigma(R)` as `z -> 0+`. Correct fix direction: keep the clamp (m > 1 genuinely does return nan) **and** force subdivision where the integrand's mass is, e.g. explicit break points at `R +/- c*z`, mirroring what the backend path's graded panels do.

Reworking on this branch.

jobovy-agent:
**Reworked.** The earlier self-review stands — the clamp alone was necessary but not sufficient, and the old test asserted the wrong physics. Both are now addressed, and the scope shrank twice on measured evidence.

### The clamp turned a nan into a silently wrong number

For a razor-thin disk `Fz(z -> 0+) -> -2*pi*Sigma(R)`, a **constant**. On the clamp-only code at R=0.5 (`-2*pi*Sigma = -1.8646`):

| \|z\| | Fz | Fz/(-2πΣ) |
|---|---|---|
| 1e-5 | -1.8646153e+00 | 0.99999 |
| 1e-6 | **+1.9141609e-06** | -0.000001 |
| 1e-7 | -1.8646321e+00 | 1.00000 |
| 1e-8 | **+1.9137786e-08** | -0.000000 |
| 1e-20 | **+1.9137792e-20** | -0.000000 |

Not even a clean cutoff — 1e-7 lands correctly while 1e-6 and 1e-8 do not.

### Fix

Substitute `a = R + |z| t` around the peak, so `(a-R)^2 + z^2 == z^2 (1+t^2)` is formed **analytically** and never by subtraction; `da = |z| dt` then cancels the `1/|z|` the peak contributes against the `-4z` prefactor. Verified correct to `|z| = 1e-100`: `Fz/(-2πΣ)` = 0.99999999 at 1e-8, 0.99994 at 1e-30.

**numpy is unchanged outside the pathological band** — gated on `|z| < 1e-4*R`, with `z == 0` kept on the original path. `zforce` is byte-identical at 15 values of z from 2.0 down to 5.1e-5 and at z=0.

### Scope: zforce only, deliberately

I first applied the substitution to all six integrands. `Rforce` then returned `-inf`, and the reason is structural rather than a slip: `zforce` has a `-4z` prefactor that cancels the peak's `1/|z|`, so its substituted integrand is finite. `Rforce` and the second derivatives have no such prefactor — their near region relies on an odd-part cancellation about `a = R`, which the substitution evaluates as a difference of huge numbers. Reverted rather than ship a transformation the code had just falsified.

So `Rforce`, `R2deriv`, `z2deriv`, `Rzderiv` **still have the same unresolved peak** and are inaccurate (not nan) there — e.g. `R2deriv(0.5, 1e-8)` is `-inf`. The test says so explicitly instead of implying coverage with an isfinite sweep. Per-method treatment is a genuine follow-up.

### Test

Now asserts the physics: `Fz -> -2*pi*Sigma(R)` to 3e-4, correct sign, and oddness in z. The previous version asserted `Fz/z` was constant — the wrong limit, which passed *because* the values were garbage.

PR #1252, opened 2026-08-04
tests: un-defer 27 of the 28 jax spherical-DF slow-skips (slow-skip 230 -> 203)

jobovy-agent:
These 28 were slow-skipped under forced jax because they sampled (300k CDF
inversions) and integrated the DF by scipy at scalar points, so every scalar
evaluation dispatched its own XLA graph. The list's own comment says the fix is
the Track F spherical-DF backend migration — *"makes them fast; delete these
lines as it lands"* — and that migration has landed.

### Measured, not assumed

The whole `test_sphericaldf.py` file was re-run under forced jax with
`--timeout 300` (serial; the file has cross-test state) and the 28 read back out
of the junitxml. **27 pass inside the cap** — slowest 177.6 s, 25 under 30 s —
costing 578 s of a 1591 s file.

### One is kept, and why

`test_king_sigmar` hits the 300 s timeout in that run. Measured without the cap
(by calling the test function directly, since the slow-skip list has no bypass
and pytest would simply skip it) it **passes in 1028.6 s**, i.e. ~730 s
projected on CI. It is a cost problem, not a correctness gap — it needs
KingPotential vectorization, and its header now records that.

### A correction worth flagging

The header's old CI-vs-local figures are stale — they predate Track F — and
scaling from them said *"CI is 1.96x slower"*, which is backwards. Measured on
the same code both sides (200.1 s locally against 141.5 s on CI for
`test_dehnendf_sample_powerrise_returnROrbit`), **CI is ~0.71x this box's wall
time**. That is why `test_king_sigmar_directint` is un-deferred here at 177.6 s
(~126 s projected) rather than kept: the earlier decision to keep it rested on
the inverted factor. Its standalone time matches the 178.7 s seen in the
full-file run, so the per-test numbers are not inflated by other load.

Net: `backend_slow_skip.txt` 230 -> 203.

### DIFF
--- a/tests/backend_slow_skip.txt
+++ b/tests/backend_slow_skip.txt
@@ -9,44 +9,15 @@
 #
 # format: <backend> <nodeid>   (same as backend_xfail.txt)
 #
-# --- jax spherical-DF sampling/quadrature (Track F spherical-DF migration) ---
-# These sample (300k CDF inversions) and integrate the DF by scipy at scalar
-# points; under forced-jax every scalar evaluation dispatches an XLA graph, so
-# each test runs minutes. Selected by measured CI duration >= 90 s on the
-# all-backend runs 27500392935 + 27503988401 (complete shard wall 3111 s ->
-# ~1450 s after skip; slowest survivor 84 s = >3x headroom under 300 s). The
-# Track F spherical-DF backend migration (vectorized sampling + GL quadrature)
-# makes them fast; delete these lines as it lands. (Distinct from Track A #39,
-# which vectorizes the SCF/assoc_legendre *potential* path.)
-jax tests/test_sphericaldf.py::test_constantbeta_dehnencore_in_nfw_dens_massprofile
-jax tests/test_sphericaldf.py::test_constantbeta_dehnencore_in_nfw_dens_spherically_symmetric
-jax tests/test_sphericaldf.py::test_constantbeta_interpolatedpotentials_beta
-jax tests/test_sphericaldf.py::test_constantbeta_interpolatedpotentials_dens_directint
-jax tests/test_sphericaldf.py::test_eddington_differentpotentials_dMdE_integral
-jax tests/test_sphericaldf.py::test_eddington_differentpotentials_dens_directint
-jax tests/test_sphericaldf.py::test_eddington_interpolatedpotentials_dens_directint
-jax tests/test_sphericaldf.py::test_eddington_powerspherical_divergent_auto_rmin
-jax tests/test_sphericaldf.py::test_eddington_powerspherical_massprofile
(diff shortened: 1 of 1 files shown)


jobovy-agent:
**Blocked on a stuck ReadTheDocs status — needs an admin merge.**

All **151** checks on `8306fa94` are green; zero failures. The only thing outstanding is the `all-clear` gate, and it cannot conclude:

```
legacy statuses on the head commit:
  pending  docs/readthedocs.org:galpy   updated 2026-08-04T21:48:41Z   <- frozen
  success  pre-commit.ci - pr
check-runs (excluding all-clear): all completed successfully
```

RTD has **exactly one build** for this commit (`33915761`), and it is `Cancelled` — created 45 s after #1251 merged, consistent with RTD cancelling a superseded build, except no replacement was ever queued. So the status has no build left to move it off `pending`.

`all-clear` polls legacy commit statuses and counts `state == "pending"`, so it waits until `MAX_WAIT_SECONDS` (5 h) and then fails. Re-running it does not help — it re-polls the same frozen status and fails again 5 h later.

**Why I can't fix it myself:** ruleset `feat` requires `all-clear` on `feat/backends-develop` with `bypass_actors: null`, and I have `push` only. The one self-service route is a new commit (fresh RTD build + fresh budget), which discards 141 already-passing jobs and re-queues ~150 on a pool that is currently the bottleneck — it would also delay #1245/#1253 by hours.

**Ask:** admin-merge this (squash). The change itself is fully verified — audit done, 0 `.py` files touched, slow-skip 230 → 203, exactly the 27 lines the title claims, and the one kept entry (`test_king_sigmar`) matches its new comment.

**#1245 and #1253 need nothing from you** — their RTD statuses are SUCCESS, so if their `all-clear` times out under the current CI saturation I can clear it with a plain re-run once their shards land.

jobovy-agent:
Unstuck this myself rather than leaving it for the admin merge I pinged about.

The blocker was diagnosable via the RTD API: build `33915761` for `8306fa94` is `state=cancelled` with `finished=null`, and a cancelled build never posts a final commit status. So `docs/readthedocs.org:galpy` sat PENDING permanently, the all-clear aggregate counted it as pending, and it was going to hit its 5h timeout at 05:05:31Z. Re-running all-clear would just have waited another 5h for a status that was never coming. The same day's RTD builds for #1245 and #1253 both finished fine, so this was one cancelled build, not a docs problem.

I did not fabricate the missing status — the docs build genuinely wasn't verified, so asserting success would have been false. Instead: empty commit `d37654f5` for a fresh SHA and therefore a real RTD build. It vanishes on squash-merge. The previous all-clear was auto-cancelled by the concurrency group, so no runners were wasted on the stale SHA.

The tree is unchanged from the 151/151-green `8306fa94` — `git diff 8306fa94 d37654f5` is empty.

PR #1261, opened 2026-08-07
diskdf/surfaceSigmaProfile: remove the numpy island via a consumer boundary

jobovy-agent:
Removes the last numpy island in the Pdf.1 disk-DF foundation.

`surfaceSigmaProfile`'s four R-evaluators carried

    xp = get_namespace(R) if is_backend_array(R) else numpy

which discards the forced namespace in exactly the case where `get_namespace`
would have honoured it: `is_backend_array` is False for a plain float, so a
float under a forced backend took the numpy branch. The profile was a numpy
island — non-differentiable, non-backend compute.

### Why the leaf fix alone is not enough

Replacing the guard with `@backend_input("R")` + `_backend_compatible = True`
does make the leaf backend-native (`float64` -> `Tensor`, confirmed on both
backends). On its own it also **regresses 47 of 129** forced-torch
`test_diskdf` tests: `diskdf` is not migrated, so its numpy branches receive
Tensors and hand them to `scipy.integrate.quad`, `numpy.fabs`, and
`numpy.lib.scimath.log` — the last promoting its complex result against a
Tensor into a complex Tensor that `_real` cannot unwrap.

So the consumer guards its own boundary: all 30 numeric call sites route
through one `_ssp` accessor that casts back to numpy unless the caller itself
passed a backend array. `diskdf.py` is the only consumer of the profile in the
tree, so the boundary cannot be bypassed by adding a call site later.

### Measured

Whole file, forced torch, counts from `--junitxml`:

| arm | failures / 129 | time |
|---|---|---|
| baseline (data-guard) | 0 | 502s |
| leaf only | **47** | 1203s |
| leaf + 2 consumer sites | **11** | 1021s |
| leaf + all 30 (this PR) | **0** | 1384s |

numpy: 129 passed, and **bit-identical and type-identical** to the unmodified
tree over 16 cases (4 methods x `log`={False,True} x scalar/array), compared as
`.view(int64)`. Scalar input still returns `float64`.

Forced-jax and forced-torch gate: `test_backend_surfacesigma`,
`test_evolveddiskdf`, `test_diskdf`.

### Known cost, not hidden

The forced-torch `test_diskdf` shard goes 502s -> 1384s (2.8x). That is
per-call coercion overhead in `diskdf`'s velocity-grid loops, not the extra
call sites — the leaf-only arm was already 1203s with 28 fewer guarded sites.
Vectorizing those loops is the follow-up that buys it back; filed separately
rather than folded in here.

### DIFF
--- a/galpy/df/diskdf.py
+++ b/galpy/df/diskdf.py
@@ -29,7 +29,7 @@ numpylog = numpy.lib.scimath.log  # somehow, this code produces log(negative), w
 from scipy import integrate, interpolate, optimize, stats
 
 from ..actionAngle import actionAngleAdiabatic
-from ..backend import get_namespace, is_backend_array
+from ..backend import as_numpy, get_namespace, is_backend_array
 from ..backend.quadrature import nested_quad
 from ..orbit import Orbit
 from ..potential import PowerSphericalPotential
@@ -265,8 +265,8 @@ class diskdf(df):
             * (
                 1.0 / self._gamma**2.0
                 - 1.0
-                - R * self._surfaceSigmaProfile.surfacemassDerivative(R, log=True)
-                - R * self._surfaceSigmaProfile.sigma2Derivative(R, log=True)
+                - R * self._ssp("surfacemassDerivative", R, log=True)
+                - R * self._ssp("sigma2Derivative", R, log=True)
             )
         )
         if numpy.fabs(va) > sigmaR1:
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Forced-backend gate complete, all five arms green (whole files, `--junitxml`, counts parsed from the XML):

| arm | result | time |
|---|---|---|
| `test_backend_surfacesigma` jax | 12 passed | 11s |
| `test_backend_surfacesigma` torch | 12 passed | 8s |
| `test_evolveddiskdf` jax | 35 passed | 863s |
| `test_evolveddiskdf` torch | 35 passed | 528s |
| `test_diskdf` jax | **124 passed, 5 skipped, 0 failures** | 3369s |

plus the torch `test_diskdf` arm already in the description (129 passed, 0 failures).

PR #1264, opened 2026-08-07
evolveddiskdf: differentiable + traceable direct (grid-free) velocity moments

jobovy-agent:
Makes `evolveddiskdf`'s **direct** (grid-free) velocity moments differentiable and
traceable on jax/torch. numpy keeps its adaptive `dblquad` untouched.

### The defect was silent, and the test suite could not see it

Under a backend the direct path produced the **right value** — scipy just
evaluates eagerly with concrete floats — while `jax.grad` raised
`ConcretizationTypeError`, because scipy's `quad` calls `float()` on the tracer.
So every value-comparison test passed and nothing was wrong-looking.

It stayed invisible for a second reason: **every** pre-existing test in
`tests/test_backend_evolveddiskdf.py` passes `grid=True` — all 12 call sites and
every entry of the `_MOMENTS` table. The direct path had *zero* backend
coverage, so nothing ever asked it for a gradient. This PR adds the first.

### Two commits

**1. Extraction (no behaviour change).** `_buildvgrid_backend`'s batched
multi-orbit integrate-and-evaluate becomes
`_df_at_velocities_backend(R, phi, t, vRf, vTf, xp)`, taking arbitrary velocity
*points*. Extracted rather than copied: a second hand-written batch would have
had to reproduce the deliberately non-`linspace` grid construction that exists
because `torch.linspace` drops the gradient w.r.t. tensor endpoints.

**2. Direct backend branch.** The numpy path integrates a **disc** of radius
`nsigma` centred on `(meanvT/sigmaT1, meanvR/sigmaR1)`. Substituting polar
coordinates about that centre maps the disc *exactly* onto
`r ∈ [0, nsigma] × θ ∈ [0, 2π]` with Jacobian `r`, so `nested_quad`'s
tensor-product rule applies directly — and the `sqrt(nsigma² - …)` inner limits
disappear, which is what a fixed-order rule resolves worst.

Also made traceable: `if initvmoment == 0.0` is a Python bool on a tracer.
Selected with `xp.where` on the **backend branch only** — `numpy.where` returns a
0-d array, which would change `vmomentsurfacemass`'s public numpy return type
from `float`.

### A regression I introduced and caught

My first version gated on `xp is not numpy`. Under a *forced* backend with a
plain-float `R`, `get_namespace` returns torch while the data are still scalars,
so `torch.where` got floats: **7 forced-torch failures**. numpy and jax passed
35/35 — their `where` accepts scalars, only torch's is strict. A jax-only gate
would have shipped it.

The fix was not to coerce. The sibling grid path (`_buildvgrid`, #141)
deliberately gates on `is_backend_array(R)`, so a forced backend with numpy data
keeps the fast path rather than being flipped onto orbit integration. I had made
two sibling paths in one method disagree; they now gate identically.

### Tests

Three added, the first backend coverage of `grid=False`:
- value vs numpy at quadrature tolerance, **with an `is_backend_array` assertion
  so a silent numpy fallback fails rather than passes**;
- gradient vs **backend-consistent** FD (not a numpy constant — backend and numpy
  use different solvers *and* different quadrature, so pinning numpy's number
  would force the tolerance to absorb both);
- jit-vs-eager.

Cost is not a regression: the new grad test runs 50.2s (torch) / 49.8s (jax)
against 48.6–50.2s for the existing grid grad tests.

### Gate

All four arms on the final tree, counts from `--junitxml`:

| suite | backend | result |
|---|---|---|
| `test_evolveddiskdf` | numpy | 35 passed, 0 failed |
| `test_evolveddiskdf` | torch | 35 passed, 0 failed |
| `test_evolveddiskdf` | jax | 35 passed, 0 failed |
| `test_backend_evolveddiskdf` | — | 41 passed, 0 failed |

Measured on this branch:

```
numpy direct (adaptive dblquad)   6.6802698135e-02   (float64, type unchanged)
jax   direct (polar GL)           6.6802596615e-02   rel 1.52e-06
jax.jit(direct)                   6.6802596615e-02   == eager
jax.grad d/dR                    -1.90281325e-01     (was ConcretizationTypeError)
backend FD d/dR                  -1.90281325e-01     rel 1.32e-10
```

### DIFF
--- a/galpy/df/evolveddiskdf.py
+++ b/galpy/df/evolveddiskdf.py
@@ -603,12 +603,17 @@ class evolveddiskdf(df):
                     return self._vmomentsurfacemassHierarchicalGrid(n, m, grido)
         # Calculate the initdf moment and then calculate the ratio
         initvmoment = self._initdf.vmomentsurfacemass(R, n, m, nsigma=nsigma, phi=phi)
+        # TRANSITIONAL data-guard, mirroring _buildvgrid's gate on is_backend_array
+        # (not on a forced namespace): a forced backend with a plain-float R keeps
+        # the adaptive scipy path, so the fast numpy suite does not get flipped onto
+        # orbit integration. Only a genuine backend R takes the differentiable route.
+        _direct_backend = is_backend_array(R)
         xp = get_namespace(R, sigmaR1, sigmaT1, meanvR, meanvT)
-        if xp is numpy:
+        if not _direct_backend:
             if initvmoment == 0.0:
                 initvmoment = 1.0
         else:
-            # Same guard, but selected rather than branched: under a trace the
+            # Same guard, selected rather than branched: under a trace the
             # comparison is a tracer and `if` raises TracerBoolConversionError.
             # Kept OFF the numpy path deliberately -- xp.where would return a 0-d
             # array there, changing the public return type from a float.
@@ -616,7 +621,7 @@ class evolveddiskdf(df):
         norm = sigmaR1 ** (n + 1) * sigmaT1 ** (m + 1) * initvmoment
         if isinstance(t, (list, numpy.ndarray)):
             raise OSError("list of times is only supported with grid-based calculation")
-        if xp is not numpy:
+        if _direct_backend:
             return (
                 self._vmomentsurface_direct_backend(
                     R, az, t, n, m, nsigma, sigmaR1, sigmaT1, meanvR, meanvT, xp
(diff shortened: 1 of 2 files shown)


jobovy-agent:
**Correction to the squash-commit message.** I passed `--body` as a double-quoted shell string, so bash command-substituted two backtick-quoted fragments and they were dropped from the merge commit. Restoring them here so the record is complete:

- Item 2 should read: *"Also made TRACEABLE: `if initvmoment == 0.0` is a tracer bool, fixed with `xp.where` on the BACKEND branch only…"*
- Item 3 should read: *"My first version dispatched on `xp is not numpy`, which broke 7 forced-torch tests…"*

The code, tests, and all measured numbers in the commit are unaffected — only those two inline code spans were lost.

PR #1271, opened 2026-08-08
planar adapters: forward to the undecorated evaluators (989 -> 5 boundary crossings)

jobovy-agent:
The planar→3D adapters forward through the **decorated** public API
(`self._Pot(...)`, `.Rforce`, `.phitorque`). Under a forced backend each of those
re-enters `@backend_input` and pays ~178 µs of coercion, so a caller that
evaluates a planar potential in a loop pays it *per iteration*.

`actionAngleSpherical`'s rperi/rap bisection does exactly that — **981 of its 989
crossings per `actions()` call came from this one forward.**

## Change

Five forwards have an existing undecorated inner and now use it:

| forward | now calls |
|---|---|
| `_evaluate` (axi + non-axi) | `_evaluatePotentials` |
| `_Rforce` (axi + non-axi) | `_evaluateRforces` |
| `_phitorque` | `_evaluatephitorques` |

The four second-derivative forwards are **unchanged**: there is no
`_evaluateR2derivs` / `_evaluatephi2derivs` / `_evaluateRphideriv`, and none is on
a hot path. I did not invent inners to make the diff look uniform.

One file, +8/−5.

## Measured

```
aASpherical.actions        989 -> 5     (198x)
aASpherical.actionsFreqs  1023 -> 23    (44x)
```

Same site measured at 997/1031 on torch before the fix, so one backend-agnostic
change covers both arms.

## numpy is byte-identical

sha256 over **3918 values**, computed in two separate worktrees, matches develop
exactly: 5 potentials × {value, Rforce} × 300 radii through the planar adapter; a
`DehnenBarPotential` non-axi adapter × {value, Rforce, **phitorque**} × 100 radii
× 3 azimuths (so the phi-carrying sites are exercised, not assumed); and
`actionAngleSpherical.actionsFreqs` itself.

## Backend agreement unchanged

Worst relative backend-vs-numpy deviation over 36 comparisons (3 points ×
{jax, torch} × 6 outputs) is **6.4239e-10 both before and after** — pre-existing
quadrature/bisection convergence, not introduced here. Measured on develop too,
so the figure has a baseline.

## Gates

```
numpy  test_potential.py + test_actionAngle.py   1492 passed / 0 failed
jax    test_actionAngle.py --backend=jax          180 passed / 0 failed
torch  test_actionAngle.py --backend=torch        182 passed / 0 failed
```

## Follow-up

`galpy-jax/appendices/boundary_census.py` lists this site in `KNOWN_HOT`;
deleting that entry once this lands is how the fix gets locked in.

### DIFF
--- a/galpy/potential/planarPotential.py
+++ b/galpy/potential/planarPotential.py
@@ -20,6 +20,9 @@ from .Potential import (
     Potential,
     PotentialError,
     _check_potential_list_and_deprecate,
+    _evaluatephitorques,
+    _evaluatePotentials,
+    _evaluateRforces,
     lindbladR,
     plotEscapecurve,
     plotRotcurve,
@@ -621,7 +624,7 @@ class planarPotentialFromRZPotential(planarAxiPotential):
         -----
         - 2010-07-13 - Written - Bovy (NYU)
         """
-        return self._Pot(R, 0.0, t=t, use_physical=False)
+        return _evaluatePotentials(self._Pot, R, 0.0, t=t)
 
     def _Rforce(self, R, phi=0.0, t=0.0):
         """
@@ -645,7 +648,7 @@ class planarPotentialFromRZPotential(planarAxiPotential):
         -----
         - Written on 2010-07-13 by Bovy (NYU).
         """
-        return self._Pot.Rforce(R, 0.0, t=t, use_physical=False)
+        return _evaluateRforces(self._Pot, R, 0.0, t=t)
 
     def _R2deriv(self, R, phi=0.0, t=0.0):
         """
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Full census sweep on the branch (jax, C-ext loaded, x64) — the fix helps every
planar consumer, not only `actionAngleSpherical`:

| workload | before | after |
|---|---|---|
| `aASpherical.actions` | 989 | **5** |
| `aASpherical.actionsFreqs` | 1023 | **23** |
| `qdf.density` | 64 | **52** |
| `qdf.sigmaR2` | 84 | **60** |
| `diskdf.*`, `Potential.*`, `Orbit.*`, `aAStaeckel` | ≤10 | unchanged |

`potential/planarPotential.py:_evaluate` was the top caller at **1968** crossings
across the sweep; it no longer appears in the ranking at all. Highest remaining
per-workload count is 60 (was 84).

No workload regressed.

PR #1282, opened 2026-08-09
KuijkenDubinski _mass: backend quadrature (2 traced defects; RotateAndTilt attempted + reverted)

jobovy-agent:
Fixes the **2 traced defects** from gh#1281 that turned out to be fixable, and
drops their ledger entries.

gh#1281 measured `test_potential.py` traced for the first time and filed 8
defects with mechanisms rather than fixes. This is the fix pass for the pair
with one shared root cause. The other 6 are **not** fixed here — see below,
including one I attempted and reverted.

## `KuijkenDubinskiDiskExpansionPotential._mass` — scipy quad rejects a tracer

`Potential.mass()` already routes its shell quadrature through
`backend.quadrature.quad`. This **per-potential `_mass` override** — the same
shell-quadrature shape — was left on `scipy.integrate.quad`, which calls
`float()` and so rejects a tracer. One copy of a pattern migrated, its sibling
missed.

`DiskSCFPotential` inherits it, which is how `McMillan17` reaches it;
`Cautun20` arrives at the same bottom via
`AdiabaticContractionWrapper.__init__ → Potential.mass → CompositePotential._mass`,
so one fix clears both.

Swapped onto the **existing** helper rather than adding one: namespace-anchored
limits so dispatch follows `R`, `xp.cos/sin`, dead `scipy` import dropped.

**Verified:**

* **numpy byte-identical** — sha256 of `mass()` at 4 radii is `a0b8d1c4…` on
  both an unmodified worktree at this base commit and the patched tree.
* traced vs numpy differ by **6.0e-6 relative** — the
  Gauss-Legendre-vs-adaptive floor, inside the tests' own ~1.5e-2 band.
* `mass()` is now differentiable in `R`: `d mass/dR = 1.0158e+08` (jax) and
  `1.0158e+08` (torch, via autograd) at `R = 50/8.21`, agreeing to ~6e-7.
* both backends checked — the fix is in shared code, but torch was verified
  separately rather than assumed.

## Duplication audit

23 `_mass` overrides in the tree; exactly **two** used `scipy.integrate.quad`.
The other is `EllipsoidalPotential._mass`, which carries an explicit
"Pspecial-blocked … no backend-agnostic replacement -> numpy only" comment. That
comment now looks stale (`backend.quadrature.quad` exists, and the same file
uses it 35 lines later), but nothing in either sweep demands the change, so it
is **left alone and recorded as a lead** rather than widened into this PR.

## What this does NOT fix

**`RotateAndTiltWrapperPotential` (5 entries) — attempted and reverted.** The
wrapper never sets `_backend_accepts_arrays`, and three internal sites assume
rank. Making all three broadcast works, and is bit-identical to the scalar loop
(40/40 checks across 4 geometries × 10 methods, numpy sha unchanged) — but the
tests **still fail**, differently:

    (R,Z,phi) = (0.500, 10.000, 0.000)
    tpoissondens = 0.07701375  vs  tdens = 0.07209774
    rel. diff = 6.8e-02   against a 1e-8 bar

Opting the wrapper in trades a hard `TypeError` for a **silent 6.8% error**:
once the node array is accepted, the traced Poisson surfdens runs on fixed-order
Gauss-Legendre where scipy's adaptive `quad` had been coping, and at |Z|=10 that
is not enough. Reverted. Opting it in first requires the traced surfdens
quadrature to be **range-aware**, not merely rank-correct.

**`AnyAxisym forceAsDeriv` (1 entry)** — pre-existing, filed separately, and now
confirmed **cross-backend** (torch reproduces it identically, and again as
`test_eccentricity` in `test_orbit`).

## Ledger

Drops exactly 2 entries, `test_McMillan17` and `test_Cautun20`. Verified by set
difference against the branch point: 2 removed, 0 added.

### DIFF
--- a/galpy/potential/KuijkenDubinskiDiskExpansionPotential.py
+++ b/galpy/potential/KuijkenDubinskiDiskExpansionPotential.py
@@ -5,7 +5,6 @@
 import copy
 
 import numpy
-from scipy import integrate
 
 from ..backend import coerce_coords, get_namespace
 from ..backend.special import logsumexp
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Closing: this branch was pushed from the wrong ref. It contains the RotateAndTilt array-input attempt, which **fails its own gate** (4 of 5 `poisson_surfdens` tests, a silent 6.8% Poisson error at |Z|=10), plus a 7-line ledger drop instead of 2, on the pre-#1281 base. Superseded by a correctly-built branch — see the replacement PR.

PR #1285, opened 2026-08-09
RotateAndTilt surfdens: split the vertical quadrature at the potential's mid-plane (clears 5 traced entries)

jobovy-agent:
Clears the **5 RotateAndTilt `poisson_surfdens` jax-jit ledger entries** that
gh#1283 diagnosed but deliberately left open.

## The defect

The traced surfdens quadratures split the vertical interval at `z=0` and cluster
their nodes there — correct only when the density sits at `z=0`. An offset or
tilted `RotateAndTiltWrapperPotential` puts its disk plane elsewhere, so the rule
samples a sharply-peaked integrand where it is flattest. At the failing point of
`mockOffsetMWP14WrapperPotential` (`offset=[1,1,1]`, plane at `z=-1`), that
reproduces the observed `tdens = 0.07209774` **to all 8 digits**.

Node count is not the lever: n=200 with the wrong split is still 6.2e-03, while
n=50 with the right split is 1.2e-10.

## The fix, in four links

1. **`RotateAndTiltWrapperPotential` broadcasts over a trailing node axis**
   (re-applies the change gh#1283 reverted — it is a prerequisite, the wrapper
   rejects the node array outright without it).
2. **`Potential._vertical_quad_split`** (0.0 by default) + the wrapper's override,
   solving where the wrapped mid-plane crosses the vertical line at `(R, phi)`.
   Falls back to 0 when the plane misses the interval — clamping to the boundary
   degenerates a panel and measured 2-6x *worse* than not trying.
3. **`CompositePotential` delegates both surfdens routes per component**, so each
   splits at its own mid-plane. Σ of a sum is the sum of the Σs, and the Poisson
   integrand is additive too. This was the blocker: `CompositePotential` defines
   `_dens` but *inherited* `_surfdens`, so the quadrature ran once over the
   summed density at the composite's plane, which is always 0.
4. **`testMWPotential` forwards `_surfdens_poisson`** — completing a delegation
   it already does for `_evaluate`, `_Rforce`, `_zforce`, `_R2deriv`, `_z2deriv`,
   `_dens`, `_surfdens` and `vcirc`. Its `_surfdens` even threads `forcepoisson`
   through, but `Potential.surfdens` raises *before* reaching it on that route,
   so the parameter was dead code.

Accuracy at the failing point, split vs the true density peak:

| mock | split | true peak | err(split 0) | err(split c) |
|---|---|---|---|---|
| RotatedAndTilted | -0.5000 | -0.4950 | 1.383e-02 | **1.070e-10** |
| ...wInclination | -1.0925 | -1.0850 | 4.736e-03 | **1.450e-06** |
| Offset | -1.0000 | -1.0000 | 6.366e-02 | **1.176e-10** |

## numpy is untouched

The delegation is backend-only; numpy keeps the inherited single scipy
quadrature. `test_potential.py` on numpy: **1283 passed, 0 failures, 0 errors**.

## Gates

| run | result |
|---|---|
| `test_potential.py`, numpy, whole file | 1283 passed, **0 failures** |
| `test_potential.py`, `--backend jax --jit`, whole file | 1278 passed, **0 failures**, 5 xfail, 102 skipped |

Re-measured at the current head `bcb2172e` (the table previously read
1276 / 7 xfail, taken before the last two commits and before gh#1283 landed in
the base).

The 5 remaining xfails are all expected, and were read out of the run's junitxml
rather than recalled: 3 AnyAxisym (the known 1-m cancellation, deliberately
**not** touched here) and 2 pickling. `McMillan17` and `Cautun20` are no longer
among them — **they now pass**, because gh#1283 merged and this branch's base
contains it. That is the whole of the 1276 -> 1278 / 7 -> 5 difference; nothing
in this PR changed to cause it.

## Tests

Six new tests in `test_backend_wrappers.py`, **A/B'd: 6 pass with the fix, 6 fail
without it.**

* the split is checked against `_rect_transformed` (does the wrapped `z'`
  actually vanish there, <1e-12) over 3 `(R, phi)` points and 4 geometries
  including the edge-on degenerate case — not against a re-derivation of the
  same formula, which would not catch a sign error;
* the quadrature test asserts **both** arms — accurate at the mid-plane (<1e-8)
  *and* visibly wrong at 0 (>1e-2) — so it cannot pass on an easy integrand;
* the composite test is structural. A value comparison passes **vacuously**:
  untraced, both routes go through the same adaptive scipy quadrature and agree
  whether or not the delegation exists. An earlier draft did exactly that.

## Ledger

**347 → 342**, exactly 5 removed, verified by set difference. The AnyAxisym
`poisson_surfdens` entry is deliberately kept — different root cause.

## Cross-backend

Verified on torch as well as jax — the same 5 nodeids pass under
`--backend torch --jit` (5 passed, 243 s). That matters because defects in this
area have repeatedly been per-backend, so a jax-only result would not have been
evidence for torch.

## Sequencing (after gh#1284)

gh#1284 ledgers these same 5 tests as **`torch-jit` xfails** (it originally
mis-typed 4 of them as `slow-skip`, corrected in `19832bb1` — the sweep that
produced those timings ran in a worktree carrying the then-unmerged broadcast
fix, so on a clean tree they fail fast rather than run slowly).

This branch fixes all 5 on torch too, so once rebased onto post-gh#1284 develop
it should additionally drop these five lines from `tests/backend_xfail.txt`:

    torch-jit ...::test_poisson_surfdens_potential[mockOffsetMWP14WrapperPotential]
    torch-jit ...::test_poisson_surfdens_potential[mockRotatedAndTiltedMWP14WrapperPotential]
    torch-jit ...::test_poisson_surfdens_potential[mockRotatedAndTiltedMWP14WrapperPotentialwInclination]
    torch-jit ...::test_poisson_surfdens_potential[mockRotatedAndTiltedTriaxialLogHaloPotentialwInclination]
    torch-jit ...::test_poisson_surfdens_potential[mockRotatedTiltedOffsetMWP14WrapperPotential]

**Keep** `torch-jit ...[AnyAxisymmetricRazorThinDiskPotential]` — different root
cause (the known 1-m cancellation), untouched here, exactly as its jax twin is
kept.

Checked with `git merge-tree`: this branch already merges onto gh#1284's head
cleanly, so the rebase is mechanical.

## What this PR's CI does and does not prove

Worth stating explicitly, because it is easy to read a green matrix as
validating the headline claim, and here it does not.

**Per-PR CI runs the backend shards EAGER.** `--jit` is set only by a
`workflow_dispatch` with `jit=true` (`JIT_FLAG` in `backend-tests.yml`), so the
5 cleared entries — all `jax-jit` — are not exercised by this PR's own checks.
Eagerly, `_quad_needs_backend` routes both surfdens routes to
`scipy.integrate.quad`, so those tests pass on develop already and would pass
here regardless of whether the fix works.

**What CI does prove** is the mechanism, because the 6 new tests in
`test_backend_wrappers.py` were deliberately written not to need `--jit`:

* the plane-crossing solve is checked against `_rect_transformed` directly;
* the quadrature accuracy is checked by calling `backend.quadrature.symmetric_quad`
  on numpy with each split point, asserting **both** arms (<1e-8 at the mid-plane,
  >1e-2 at zero);
* the composite delegation is checked structurally, since a value comparison
  passes vacuously on the eager path.

Those 6 fail without this change and pass with it (A/B'd), so a green CI here is
real evidence — for the mechanism, not for the traced end-to-end result.

**The traced result is local evidence**, and is stated as such:
`test_potential.py` under `--backend jax --jit` on this branch is
**1276 passed / 0 failures / 7 expected xfail**, and the 5 nodeids also pass
under `--backend torch --jit`. That is the standing "traced green is asserted by
me, not enforced by CI" caveat — and the reason the `--jit` CI dimension is an
open question for the maintainer.

### DIFF
--- a/galpy/potential/CompositePotential.py
+++ b/galpy/potential/CompositePotential.py
@@ -1,6 +1,9 @@
 ###############################################################################
 #   CompositePotential.py: class that represents a combination of potentials
 ###############################################################################
+import numpy
+
+from ..backend import get_namespace
 from ..util.conversion import physical_compatible
 from .baseCompositePotential import baseCompositePotential
 from .DissipativeForce import DissipativeForce, _isDissipative
@@ -278,6 +281,76 @@ class CompositePotential(baseCompositePotential, DissipativeForce, Potential):
         """
         return sum(
             pot.dens(R, z, phi=phi, t=t, use_physical=False)
+            for pot in self._potlist
+            if not pot.isDissipative
+        )
+
+    def _surfdens(self, R, z, phi=0.0, t=0.0):
+        """
+        Evaluate the surface density.
+
+        Parameters
+        ----------
+        R : float
+            Cylindrical Galactocentric radius.
+        z : float
+            Vertical height.
+        phi : float, optional
+            Azimuth (default: 0.0).
+        t : float, optional
+            Time (default: 0.0).
+
+        Returns
+        -------
+        float
+            Surface density at (R,z,phi,t).
+
+        """
+        # Sigma of a sum is the sum of the Sigmas, so integrate each component
+        # separately on a backend: each then splits its vertical quadrature at
+        # its OWN mid-plane (see Potential._vertical_quad_split). Integrating
+        # the summed density instead splits at this composite's plane, which is
+        # 0 -- wrong for any displaced or tilted component. numpy keeps the
+        # inherited single scipy quadrature, so its result is byte-identical.
+        xp = get_namespace(R, z, phi, t)
+        if xp is numpy:
+            return Potential._surfdens(self, R, z, phi=phi, t=t)
+        return sum(
+            pot.surfdens(R, z, phi=phi, t=t, use_physical=False)
+            for pot in self._potlist
+            if not pot.isDissipative
+        )
+
+    def _surfdens_poisson(self, R, z, phi=0.0, t=0.0):
+        """
+        Evaluate the surface density via the Poisson equation.
+
+        Parameters
+        ----------
+        R : float
+            Cylindrical Galactocentric radius.
+        z : float
+            Vertical height.
+        phi : float, optional
+            Azimuth (default: 0.0).
+        t : float, optional
+            Time (default: 0.0).
+
+        Returns
+        -------
+        float
+            Surface density at (R,z,phi,t).
+
+        """
+        # Forces and their derivatives are additive, so the Poisson route is too
+        # -- delegate for the same reason as _surfdens above, so each component
+        # splits its quadrature at its own mid-plane. numpy keeps the inherited
+        # single scipy quadrature and stays byte-identical.
+        xp = get_namespace(R, z, phi, t)
+        if xp is numpy:
+            return Potential._surfdens_poisson(self, R, z, phi=phi, t=t)
+        return sum(
+            pot._surfdens_poisson(R, z, phi=phi, t=t)
             for pot in self._potlist
             if not pot.isDissipative
         )
--- a/galpy/potential/Potential.py
+++ b/galpy/potential/Potential.py
@@ -794,45 +794,53 @@ class Potential(Force):
                 raise AttributeError  # Hack!
             return self._amp * self._surfdens(R, z, phi=phi, t=t)
         except AttributeError:
-            # Use the Poisson equation to get the surface density
-            xp = get_namespace(R, z, phi, t)
+            return self._surfdens_poisson(R, z, phi=phi, t=t)
 
-            def poisson_integrand(Rv, pv):
-                return lambda x: (
-                    -self.Rforce(Rv, x, phi=pv, t=t, use_physical=False) / Rv
-                    + self.R2deriv(Rv, x, phi=pv, t=t, use_physical=False)
-                    + self.phi2deriv(Rv, x, phi=pv, t=t, use_physical=False) / Rv**2.0
-                )
+    def _surfdens_poisson(self, R, z, phi=0.0, t=0.0):
+        """Surface density from the Poisson equation.
 
-            # numpy.fabs on a backend array emits a NumPy 2 __array_wrap__
-            # DeprecationWarning, which the coverage shard turns into an error.
-            absz = numpy.fabs(z) if xp is numpy else xp.abs(z)
-            if not _quad_needs_backend(xp, absz):
-                inner = integrate.quad(poisson_integrand(R, phi), -absz, absz)[0]
-            else:
-                # Fixed-order GL over the whole [-z, z] needs an order that grows
-                # with the range (n=100 is only 3.7e-4 at |z|=10). Splitting at
-                # the mid-plane and clustering nodes toward the split puts them
-                # where a decaying integrand has its mass, and does so RELATIVE
-                # to the range, so one node count holds at every |z|.
-                # [..., None] so R/phi broadcast against the trailing node axis
-                # (0-d becomes (1,), which broadcasts either way).
-                inner = _bquad.symmetric_quad(
-                    xp,
-                    poisson_integrand(_bquad.node_axis(R), _bquad.node_axis(phi)),
-                    absz,
-                    n=50,
-                    interior_point=self._vertical_quad_interior(xp, R, absz, phi, t),
-                )
-            return (
-                (
-                    -self.zforce(R, absz, phi=phi, t=t, use_physical=False)
-                    + self.zforce(R, -absz, phi=phi, t=t, use_physical=False)
-                    + inner
-                )
-                / 4.0
-                / numpy.pi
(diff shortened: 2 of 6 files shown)


jobovy-agent:
### The Windows red is a flake — settled by a same-SHA replicate, not by argument

`build.yml` triggers on both `push` and `pull_request`, so every commit here ran the full matrix **twice**. On `bcb2172e` both copies of the Windows shard exist:

| run | event | `windows-latest, 3.14, test_quantity.py test_coords.py` |
|---|---|---|
| [31342463692](https://github.com/jobovy/galpy/actions/runs/31342463692) | `push` | **failure** |
| [31342466332](https://github.com/jobovy/galpy/actions/runs/31342466332) | `pull_request` | **success** |

Same commit, same platform, same shard string, opposite outcomes. Identical code cannot be responsible for a difference it did not cause, so `test_streamgapdf_sample` is flaky on Windows independently of this PR. Worth noting the PR's own check list shows only one of the two, which is why this looked unattributable.

### `codecov/patch` — fixed, not explained away

Patch was 50/51 = 98.03 % against the default `auto` target (base 99.95 %). The one missed line was `Potential.py:871`, the `xp.where` in `_vertical_quad_interior`, reachable only when the mid-plane crossing is nonzero — every existing caller took the `c == 0.0` early return one line above.

1f791a90 adds a test that pins the contract rather than merely touching the line: both arms of the `where`, and specifically that the reject arm returns **0.0 and not the nearer boundary** (clamping to ±|z| degenerates a panel to zero width and measured 2–6× worse). Nothing else in the suite would catch a change to `xp.clip(c, -absz, absz)`. Verified executed with `trace.Trace` against this worktree, not inferred from a percentage.

jobovy-agent:
### Pre-merge audit

**1. Branch-on-coordinate-data (the `if` → `xp.where` class): clean.** Every new `if` branches on the namespace (`xp is numpy`), on static construction-time config (`self._norot`, `self._offset`, `self._rot`), or on tracedness — `_quad_needs_backend` asks `under_trace`, not a value probe. The one place coordinate data can appear is `c = _vertical_quad_split(...)` in `_vertical_quad_interior`, and it is guarded by `isinstance(c, float)` first: concrete potentials take the early return, a tracer short-circuits to the `xp.where`.

**2. Helper placement: clean.** `_vertical_quad_split` / `_vertical_quad_interior` are hooks on the base `Potential` (where subclasses can override), the override lives in `RotateAndTiltWrapperPotential`, and the summation overrides live in `CompositePotential`. Nothing general-purpose landed in a specific potential file, and no new quadrature helper was added — the existing `galpy.backend.quadrature` entry points are reused.

**3. Duplication: one instance, deliberately kept.** `CompositePotential._surfdens` and `._surfdens_poisson` have the same four-line shape. They differ in a way that is load-bearing and would be obscured by factoring them through a callable:

* `_surfdens` delegates to the **public** `pot.surfdens(..., use_physical=False)`, so each component picks its own route via the base-class `_surfdens`/`_surfdens_poisson` fallback, and each component's `_amp` is applied.
* `_surfdens_poisson` delegates to the **private** `pot._surfdens_poisson(...)` to force the Poisson route, and must *not* apply `_amp` — that route derives from forces, which already include it.

Verified empirically on numpy that this bookkeeping is exact (composite `_amp` is 1.0, so the public route does not double-apply):

```
composite.surfdens       0.179888191392422
sum(components)          0.17988819139242193    rel 3.09e-16
composite poisson        0.17988819139242201
sum(component poisson)   0.17988819139242201    rel 0.00e+00
```

Flagging it here rather than pushing a comment-only commit, since that would re-queue ~150 checks on a runner-constrained account for no behavioural change. Worth a clarifying comment in a later touch of this file.

jobovy-agent:
### Measured side effect: the numpy-boundary crossings collapse

Ran the boundary census (galpy-jax `appendices/boundary_census.py`, which counts `coerce_coords` crossings per caller) against this branch's worktree — `PROVENANCE: /var/tmp/rt-split/galpy/__init__.py`, C extension loaded:

| workload | before this PR | on this branch |
|---|---|---|
| `surfdens(MWP14)` | 2779 | **19** |
| `surfdens(MWP14, forcepoisson=True)` | 5307 | **2299** |

The direct route drops out of the census's HOT band entirely (threshold 1000) — a ~145× reduction — because the per-component delegation lets each potential's vertical quadrature run as one broadcast call instead of re-crossing the boundary per node.

The Poisson route improves by more than half but stays HOT at 2299. That residue is *not* something this PR can reach: it is the eager path, which `_quad_needs_backend` deliberately sends to `scipy.integrate.quad` node-by-node so scalar-only potentials keep working. Removing it needs the separate try-GL/fall-back-on-`TypeError` change, which I have measured but am keeping out of this PR.

Recording it here because the PR is framed as a correctness fix (split the vertical quadrature at the potential's own mid-plane) and this performance consequence was not stated.

PR #1294, opened 2026-08-10
interpRZPotential: build the sampling grids with one vectorised call

### LINKED ISSUE #211: SnapShotPotential with array input
SnapShotPotential with array input for R and z build a 2D grid, which is different from other potentials. Probably should be changed, or a warning given.

jobovy-agent:
Closes #211-side work Jo asked for; the seven identical grid-build loops collapse to one helper, `_grid_eval`.

## Speed (61x41 grid, vectorised vs cell-by-cell)

| potential | loop | vectorised | |
|---|---|---|---|
| MWPotential2014 | 1.06 s | 0.04 s | **24.8x** |
| MiyamotoNagai | 0.18 s | 0.01 s | 17.2x |
| NFW | 0.20 s | 0.01 s | 17.1x |

## Why it is not just a meshgrid call

Two failure modes, needing different handling:

* **raises on array input** (`AnySphericalPotential` and anything wrapping one) — `try/except`, and the fall-back is the *identical* computation so nothing is masked;
* **neither raises nor broadcasts correctly** — `try/except` sails straight past this. `AnySphericalPotential` returns force values differing from the cell-by-cell ones in **95 % of cells**, silently. So the vectorised result is accepted only after reproducing the scalar path **bit for bit** on the first and last rows (~1 % of the work).

Measured across a 19-potential zoo x 7 quantities: **94 of 108 bit-identical** over the whole grid, 9 raise, 5 differ (all AnySpherical) and are caught.

## Two costs, disclosed

1. **Fall-back potentials get slower.** `DoubleExponentialDiskPotential` builds ~15 % slower — it pays a failed vectorised attempt plus the row checks.
2. **Not strictly byte-identical.** The row check is a sample, not a proof: NFW and AnySpherical still differ from the loop by **1–2 ULP in 1–4 cells out of 2501**, from evaluation order. That is ~10 orders of magnitude below the cubic spline's own interpolation error — but it is a numpy-path change, so **your call**. Strict bit-identity would require verifying every cell, i.e. no speed-up at all.

### DIFF
--- a/galpy/potential/interpRZPotential.py
+++ b/galpy/potential/interpRZPotential.py
@@ -96,6 +96,11 @@ def _spot_check_cells(nR, nz):
     the default 251x251): measured 31x cheaper on the potentials where this
     actually costs anything.
     """
+    if nR < 1 or nz < 1:
+        # Degenerate grid: nR-1 would index -1. Sample nothing and let the
+        # spline fitter reject it as it already did before this spot check
+        # existed ("(mx>kx) failed ... mx=0"), which names the real problem.
+        return []
     ii = sorted({0, nR // 2, nR - 1})
     jj = sorted({0, nz // 4, nz // 2, 3 * nz // 4, nz - 1})
     cells = {(a, b) for a in ii for b in jj}
--- a/galpy/potential/interpRZPotential.py
+++ b/galpy/potential/interpRZPotential.py
@@ -86,6 +86,59 @@ def scalarDecorator(func):
     return scalar_wrapper
 
 
+def _grid_eval(evaluator, pot, rgrid, zgrid):
+    """Sample ``evaluator(pot, R, z)`` on the ``(rgrid, zgrid)`` tensor grid.
+
+    Tries a single vectorised call over the whole meshgrid and falls back to the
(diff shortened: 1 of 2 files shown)


jobovy-agent:
**Jo has accepted the ULP trade-off** (2026-08-10). Landing as-is: the 17–25x speed-up is worth 1–2 ULP in ~0.1 % of grid cells, which sits ~10 orders of magnitude below the cubic spline's own interpolation error.

No code change needed — the deviation was already measured, bounded and disclosed above. Recording the decision here so the reasoning is attached to the diff rather than living only in chat:

* **94 of 108** (potential x quantity) combinations are bit-for-bit identical.
* **9** raise on array input and fall back to the loop, so they are identical by construction.
* **5** (NFW, `AnySphericalPotential`) differ by 1–2 ULP in 1–4 cells out of 2501.
* The first/last-row bit-for-bit check stays, because it is what catches `AnySphericalPotential`'s *real* failure — silently wrong force values for array input in 95 % of cells.
* Cost also disclosed: potentials that fall back build ~15 % slower (`DoubleExponentialDiskPotential`).

PR #1300, opened 2026-08-12
backend: one device anchor per call, not per coordinate

jobovy-agent:
`@backend_input` coerced each declared coordinate on its own, so each derived its own device anchor **from itself alone**. A numpy/python coordinate has no device, so it anchored to `None` → the backend default device (CPU for torch), while a CUDA sibling stayed on the GPU. The evaluator was then handed a coordinate set split across devices.

Reproduced on a TITAN Xp with a spy on `MiyamotoNagaiPotential._Rforce`:

| call | evaluator sees | result |
|---|---|---|
| `R` numpy 1-D + `z` cuda 1-D | `R=cpu`, `z=cuda:0` | **cpu** |
| `R` cuda 1-D + `z` numpy 1-D | `R=cuda:0`, `z=cpu` | cuda:0 |

Whether that is fatal depends on the potential's arithmetic: a mixed-device error in some, a **silent GPU→CPU transfer** in others — the first row hands back a CPU result for a CUDA input, losing the GPU without saying so.

## Fix

Compute the anchor **once** over all declared coordinates — the wrapper already collects them in `probe` — and thread it into `coerce_coords` through a new `device=` keyword. Both orders now put every coordinate on `cuda:0`.

The numpy path cannot reach any of this: `device_of` is called only inside the `xp is not numpy` branch, and `coerce_coords` returns before consulting `device` on numpy. Byte-identical by construction.

## Tests

Both were verified to **fail without the source change**:

- `test_backend_input_anchors_all_coords_on_one_device` uses torch's **`meta`** device as a second device, so the property is checked on every runner rather than only where a GPU exists. Without a shared anchor: `split anchor: R=meta, z=cpu, phi=cpu`.
- the existing CUDA test gains a numpy-**array** sibling in both orders. Without the fix: `assert 'cpu' == 'cuda'`.

That second one is the gap worth naming. `test_mixed_coords_cross_device_cuda` already covered "mixed scalar/array coords" — but with **scalar** `R,z`, and torch accepts a 0-dim CPU tensor in a binary op with a CUDA one, so a per-coordinate anchor sails straight through. Only a non-0-dim numpy sibling exposes it.

## Gates (from finished runs)

| | |
|---|---|
| numpy `test_potential.py` | 1324 passed |
| forced torch `test_potential.py` | 1288 passed |
| forced jax `test_potential.py` | 1276 passed (57 min) |
| CUDA device test (TITAN Xp) | passed |
| device/input/conventions/coerce suites, rebased | 138 passed |

### DIFF
--- a/galpy/backend/_coerce.py
+++ b/galpy/backend/_coerce.py
@@ -61,6 +61,7 @@ import numpy
 from ._namespaces import (
     _is_floating_dtype,
     asarray_on_device,
+    default_device,
     device_of,
     is_backend_array,
 )
@@ -100,6 +101,12 @@ def coerce_coords(xp, *coords, device=None):
     if xp is numpy:
         return coords
     dev = device_of(*coords) if device is None else device
+    if dev is not None and dev == default_device(xp):
+        # placing there is what asarray does anyway, and naming it explicitly
+        # costs ~3x (it commits instead of letting the backend place) -- see
+        # default_device. Only a NON-default anchor (a CUDA sibling on a
+        # CPU-default run) is worth paying for.
+        dev = None
     out = []
     for c in coords:
         if c is None:
--- a/galpy/backend/_coerce.py
+++ b/galpy/backend/_coerce.py
@@ -66,7 +66,7 @@ from ._namespaces import (
 )
 
 
-def coerce_coords(xp, *coords):
+def coerce_coords(xp, *coords, device=None):
     """Bring coordinate inputs onto the active backend's array type.
 
     The dominant non-numpy failure mode is "the namespace resolved to a backend
@@ -86,12 +86,20 @@ def coerce_coords(xp, *coords):
         the backend's float64 -- galpy's interior precision; a bare ``asarray``
         of a Python float would give torch float32 and miss the tolerances.
 
+    ``device`` overrides the device the coerced coordinates land on. Without it
+    the anchor is derived from ``coords`` alone, which is only right when the
+    caller passes every coordinate of a call at once: coercing them one at a
+    time gives each its own anchor, so a numpy coordinate anchors to None (the
+    backend default device, i.e. CPU for torch) while its CUDA siblings stay on
+    the GPU, and the evaluator is handed a split-device coordinate set. Callers
+    that coerce coordinate-by-coordinate pass the shared anchor explicitly.
+
     The numpy backend is a strict pass-through (``coords`` returned object-
     identical) -> the numpy path stays byte-identical.
     """
     if xp is numpy:
         return coords
-    dev = device_of(*coords)
+    dev = device_of(*coords) if device is None else device
     out = []
     for c in coords:
         if c is None:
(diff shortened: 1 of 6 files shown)


jobovy-agent:
**Back to draft — this branch has a ~180x performance regression.**

The forced-jax `test_orbit/test_orbits` group-5 shard fails with two **timeouts**, and they are not marginal:

| test | develop (d76d683b) | this branch |
|---|---|---|
| `test_orbits.py::test_integration_dxdv_2d` | 1.7 s | **300 s (timeout)** |
| `test_orbits.py::test_integration_dxdv_2d_rectInOut` | 1.7 s | **300 s (timeout)** |

Neither is in any ledger, so both normally pass.

**Cause.** This PR computes one shared device anchor and passes it to every coordinate's coercion. That is correct for mixed-device inputs, but it means a numpy/python coordinate that previously got `device=None` now gets an explicit device — and an explicit device is *much* more expensive, because it forces a commit instead of letting the backend place the array:

```
asarray_on_device(..., device=None)  : 134.3 us/call
asarray_on_device(..., device=<dev>) : 397.6 us/call   (3.0x)
```

(`jnp.asarray(device=...)` does not raise here — it just costs 3x.) The `@backend_input` boundary is crossed in a tight loop during a dxdv integration, so a 3x per-crossing cost compounds into the timeouts above.

**Fix direction.** Only pass an explicit device when it can actually change placement — i.e. when the anchor differs from the backend's default device. On a CPU-only run that restores the original `device=None` fast path, while the CUDA mixed-device case (the bug this PR fixes) still gets the shared anchor. Reworking on that basis before this goes back up for review.

Good argument for the `--jit`/perf shards being part of the gate: this is invisible to any value-based assertion, since every number stays correct and only the time changes.

jobovy-agent:
**Perf regression fixed — verified by the same instrument that found it.**

The forced-jax orbit group-5 shard now **passes**, and the two tests that were timing out are back in budget:

| test | develop | before fix | after fix (`e9df1561`) |
|---|---|---|---|
| `test_integration_dxdv_2d` | 1.7 s | 300 s (timeout) | **3.4 s pass** |
| `test_integration_dxdv_2d_rectInOut` | 1.7 s | 300 s (timeout) | **3.5 s pass** |

(junit durations from the same shard, develop run 31601434841 vs PR runs 31601564919 / 31623686798.)

Being precise: there is a residual **~2x** against develop's 1.7 s. The shared anchor still costs something even when the device keyword is dropped — `device_of` runs per crossing. That is comfortably inside the 300 s budget and buys the correctness this PR exists for (no silent GPU→CPU transfer on mixed-device inputs), but it is a real cost rather than parity, so I am not claiming otherwise.

**The fix**: ask the namespace once (cached) what device it places a new array on, and drop the `device=` keyword when the anchor equals it. CPU-only runs go back to the original fast path; a CUDA sibling on a CPU-default run still gets the explicit anchor. `default_device` lives in `_namespaces.py` beside `device_of`/`asarray_on_device`.

Remaining reds are **not** code: three py3.14 `build` jobs died with `ERROR 503: Service Unavailable` / exit code 8 in a download step, all at 17:39 — one transient outage. Re-running those once the workflow finishes. codecov is still uploading.

PR #1301, opened 2026-08-12
backend: put the jit boundary inside the units layer (33 -> 8 traced failures)

jobovy-agent:
Decorators apply bottom-up, so every units-returning method ran as:

```
@potential_physical_input      1. strip INPUT units
@backend_input("R","z",...)    2. ← jit boundary
@physical_conversion("force")  3. ← builds the Quantity, INSIDE the trace
def Rforce(...)                4. body
```

`physical_conversion` therefore constructed the astropy `Quantity` **inside the traced region**, and `Quantity.__new__` calls `__array__`, which a tracer refuses. The traceback shows the nesting: `traced_call` → `_jit.py shim` → `conversion.py wrapped` → `astropy/units/quantity.py` → `TracerArrayConversionError`.

I had previously recorded these 33 jax `--jit` failures as *"astropy and tracing are mutually exclusive by design; no backend work fixes it."* That was inferred from the error text and never tested. **It was wrong** — units attach perfectly well once they attach *after* the trace returns a concrete array.

## The A/B that established it

Swapping the pair in `Potential.py` **only** moved the failure: `pot(1.1, 0.1)` and `pot.Rforce(...)` went from failing to **passing**, and the test then failed at `pot.rforce(...)` — a method in a file I hadn't swapped yet.

## Change

75 sites across 7 files (Potential 49, planarPotential 14, linearPotential 4, DissipativeForce 3, sphericaldf 2, planarDissipativeForce 2, Force 1). **No** site had the correct order, so it's uniform — and the diff is a **provably pure reordering**: 75 lines added, 75 removed, identical as multisets, zero content change.

## Measured (finished runs)

| | result |
|---|---|
| jax `--jit` `test_quantity`+`test_coords` | **33 failures → 8** |
| numpy `test_quantity` | 244 passed, 0 failed |
| numpy `test_potential` | 1288 passed, 0 failed |

The numpy path is untouched (the units decorators are no-ops with `apy-units` off) — verified rather than assumed.

## Structural invariant

Added to `test_backend_conventions.py`, where the other source-structure checks live: for any method carrying both decorators, the units decorator must be the **outer** one. A/B'd — reverting the source makes it fail on exactly the 7 offending files and name the methods. It covers all four `physical_conversion*` spellings (plain, `_tuple`, `_actionAngle`, `_actionAngleInverse`); widening from 2 to 4 found no further violations, which is what confirms 75 was the complete set. A glob guard keeps an empty file list from making it pass vacuously.

## Residual 8 — a different mechanism, deliberately not in scope here

`sphericaldf.vmomentdensity` and friends carry `@backend_input` but **no** `@physical_conversion` — they build the Quantity inline in the body (`units.Quantity(as_numpy(out) * fac, unit=u)`), so there is no decorator to reorder and the invariant cannot see them. That needs restructuring (split the traced compute from the untraced unit attachment) and is a follow-up.

### DIFF
--- a/galpy/df/sphericaldf.py
+++ b/galpy/df/sphericaldf.py
@@ -562,8 +562,8 @@ class sphericaldf(df):
             )
         )
 
-    @backend_input("r")
     @physical_conversion("velocity", pop=True)
+    @backend_input("r")
     def sigmar(self, r):
         """
         Calculate the radial velocity dispersion at radius r.
@@ -588,8 +588,8 @@ class sphericaldf(df):
             xp.sqrt(self._vmomentdensity(r, 2, 0) / self._vmomentdensity(r, 0, 0)), r
         )
 
-    @backend_input("r")
     @physical_conversion("velocity", pop=True)
+    @backend_input("r")
     def sigmat(self, r):
         """
         Calculate the tangential velocity dispersion at radius r.
--- a/galpy/potential/DissipativeForce.py
+++ b/galpy/potential/DissipativeForce.py
@@ -53,8 +53,8 @@ class DissipativeForce(Force):
         self.hasC_planar = True
 
     @potential_physical_input
-    @backend_input("R", "z", "phi", "t", "v")
     @physical_conversion("force", pop=True)
+    @backend_input("R", "z", "phi", "t", "v")
     def Rforce(self, R, z, phi=0.0, t=0.0, v=None):
         """
         Evaluate cylindrical radial force F_R  (R,z).
@@ -85,8 +85,8 @@ class DissipativeForce(Force):
         return self._Rforce_nodecorator(R, z, phi=phi, t=t, v=v)
 
     @potential_physical_input
-    @backend_input("R", "z", "phi", "t", "v")
     @physical_conversion("force", pop=True)
+    @backend_input("R", "z", "phi", "t", "v")
     def zforce(self, R, z, phi=0.0, t=0.0, v=None):
         """
         Evaluate the vertical force F_z  (R,z,t).
@@ -117,8 +117,8 @@ class DissipativeForce(Force):
         return self._zforce_nodecorator(R, z, phi=phi, t=t, v=v)
 
     @potential_physical_input
-    @backend_input("R", "z", "phi", "t", "v")
     @physical_conversion("energy", pop=True)
+    @backend_input("R", "z", "phi", "t", "v")
     def phitorque(self, R, z, phi=0.0, t=0.0, v=None):
         """
         Evaluate the azimuthal torque F_phi  (R,z,phi,t,v).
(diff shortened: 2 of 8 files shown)


jobovy-agent:
**Back to draft — this PR is not safe on its own.**

Under the all-backend harness's real context (forced jax **+** `--jit`), a spherical-DF method with a `Quantity` input returns a **silently wrong** value on this branch:

```
numpy (correct)            sigmar(100 pc) = 0.14090842477103016
forced jax + jit           sigmar(100 pc) = 0.04438549032202838   <- 68.5% wrong
```

`sigmar`/`sigmat`/`beta`/`vmomentdensity` parse their input units *inside* the body (`conversion.parse_length`), so the `Quantity` reaches the `@backend_input` boundary; `jax.jit` then converts it through `__array__`, taking the bare value and dropping the unit. Potentials are unaffected — `@potential_physical_input` is still outermost and strips input units before the boundary (verified: `Rforce(1.0 pc) == Rforce(1/8000 internal)` exactly).

Two further notes:

- `test_sphericaldf_quantity_radius_is_not_coerced_on_a_backend` currently **passes vacuously** under `--jit`: it computes its reference and its result in the same wrong context, so both are equally wrong and agree. It needs a reference that cannot be affected by the traced path.
- The fix is to hoist the unit parse outside the boundary (the arrangement `potential_physical_input`'s own docstring describes as the intended design), splitting each method into a public units-parsing wrapper and a `@backend_input` compute helper. Verified locally: traced `sigmar(1.0 pc)` then matches `sigmar(1/8000 internal)` to 2e-15.

Holding this PR until that fix lands with it, so the decorator swap never ships without the units-input handling it exposes.

jobovy-agent:
**Correction — un-drafting. The silent unit-drop is PRE-EXISTING, not caused by this PR.**

I drafted this after finding that a spherical-DF method with a `Quantity` input returns a wrong value under forced jax + `--jit`. I attributed that to the decorator swap. **That was wrong** — I never measured the baseline. On `feat/backends-develop` (`d76d683b`, no swap):

```
numpy (correct)      sigmar(100 pc) = 0.14090842477103016
forced jax + jit     sigmar(100 pc) = 0.04438549032202838
```

Identical to this branch. The swap is neutral on it; the bug predates this PR and is tracked separately.

So this PR stands on its own merits: **33 -> 8** traced failures, a provably pure decorator reordering (75 added / 75 removed, identical multisets, zero content change), and a structural invariant that A/B's clean. Every remaining failure on the shard is either loud or pre-existing.

The units-input fix (hoisting `parse_length` outside the boundary for `sigmar`/`sigmat`/`beta`/`vmomentdensity`) moves to a **stacked PR on this branch**, where it belongs — it is a different mechanism and it surfaced a further cross-backend issue (a df constructed under one backend caching arrays that a second backend then meets: `Tensor / jaxlib.ArrayImpl`) that deserves its own review.

PR #1302, opened 2026-08-12
tests: read AnyAxisym small-|z| forces through as_numpy (fixes the torch red on feat/backends)

jobovy-agent:
`tests/test_potential.py::test_anyaxisymrazorthin_smallz_limit` is failing on the forced-torch shard of `feat/backends` — [run 31553961857, job 93982389598](https://github.com/jobovy/galpy/actions/runs/31553961857/job/93982389598) (`backend torch tests/test_potential.py --splits 5 --group 5`). It is the **only** failing job in that run.

Under a forced backend the potential returns a `Tensor`, and `numpy.all` / `numpy.fabs` on a `Tensor` raise — numpy forwards `axis=`/`out=`, which torch rejects. The assertions are about *values*, so each readout is cast once through `as_numpy` (a no-op on numpy) rather than teaching every assertion about backends.

**This is not new work.** It is the identical fix already merged into `feat/backends-develop` as gh#1298 (`bf520538`). develop only reaches this branch through gh#1204, which is still a draft, so `feat/backends` stays red until that lands. The patch here is byte-identical to develop's, so when gh#1204 eventually merges this becomes a no-op rather than a conflict.

A/B on this branch under `--backend torch`: **fails** without the change, **passes** with it.

### DIFF
--- a/tests/test_potential.py
+++ b/tests/test_potential.py
@@ -13883,17 +13883,23 @@ def test_anyaxisymrazorthin_smallz_limit():
     # NB an earlier version of this test asserted Fz/z was constant there. That
     # is the wrong physics and it passed *because* the values were garbage: a
     # correct implementation fails it.
+    # as_numpy at every readout: under a forced backend these return Tensors, and
+    # numpy.all/numpy.fabs on a Tensor raise (numpy forwards axis=/out=, which
+    # torch rejects). The assertions below are about VALUES, so cast once here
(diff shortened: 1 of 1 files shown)


jobovy-agent:
Closing per Jo: fixes should not go directly to `feat/backends` — the eventual `feat/backends-develop` -> `feat/backends` merge (gh#1204) resolves this, since the fix already lives on develop as gh#1298 (bf520538). The red on that branch is expected until then.

PR #1303, opened 2026-08-12
df/jit: parse input units outside the traced boundary (stacked on #1301)

jobovy-agent:
**Stacked on gh#1301** (`backend-jit/units-jit-boundary`) — review that one first; this PR's diff is only the commit on top.

`sphericaldf`'s `sigmar`/`sigmat`/`beta`/`vmomentdensity` parse their input units **inside** the body (`conversion.parse_length`), so a `Quantity` reaches the `@backend_input` boundary and `jax.jit` converts it through `__array__`, taking the bare value and **dropping the unit**:

```
numpy (correct)     sigmar(100 pc) = 0.14090842477103016
forced jax + jit    sigmar(100 pc) = 0.04438549032202838   ← 68.5% wrong, silently
```

**Pre-existing on develop** (measured at `d76d683b`) — not introduced by the decorator swap this is stacked on. Potentials are unaffected because `@potential_physical_input` strips input units outside the boundary; these four methods carry no such decorator, so the parse is hoisted by hand: the public method parses, a `@backend_input` helper computes.

Verified: traced `sigmar(1.0 pc)` then matches `sigmar(1/8000 internal)` to 2e-15, and the jax `--jit` sphericaldf group goes **5 failed → 1 failed / 8 passed**.

Two supporting fixes:

- **`traced_call` declines to trace when the active namespace isn't the jit mode's framework.** A test entering `use("torch", force=True)` while the harness runs `--backend jax --jit` was handing torch Tensors to a `jax.jit` (*"Error interpreting argument … as an abstract array"*).
- **`test_sphericaldf_quantity_radius_is_not_coerced_on_a_backend` pins its reference to numpy.** Under `--jit` the whole test runs inside the forced-backend context, so the unpinned reference was computed on the same wrong path as the result — both wrong by the same factor, so the assertion **passed vacuously** while the value was 68% off.

## Draft: what's left

The torch parametrization of that test now fails differently — `Tensor / jaxlib._jax.ArrayImpl` inside `_vmomentdensity`. A df **constructed** under one backend caches arrays that a second forced backend then meets. That is cross-backend cached state, a distinct hazard from the units boundary, and it is what remains before this leaves draft.

### DIFF
--- a/galpy/backend/_input.py
+++ b/galpy/backend/_input.py
@@ -193,7 +193,7 @@ def backend_input(*coords):
                 # Under an opt-in trace mode this boundary is also where the
                 # jit/compile happens: the declared coordinates are the traced
                 # arguments and everything else is static. Off by default.
-                out = traced_call(method, args, kwargs, slots, nargs)
+                out = traced_call(method, args, kwargs, slots, nargs, xp)
                 if out is not NOT_TRACED:
                     return out
             return method(*args, **kwargs)
--- a/galpy/backend/_jit.py
+++ b/galpy/backend/_jit.py
@@ -23,6 +23,8 @@
 from contextlib import contextmanager
 from contextvars import ContextVar
 
+from ._namespaces import name_of_namespace
+
 # "off" | "jax" | "torch". Process-wide, like the backend selection itself.
 _JIT_CTX = ContextVar("galpy_jit_mode", default="off")
 # Set while a traced call is on the stack. Entry points call each other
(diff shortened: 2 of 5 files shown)


jobovy-agent:
**Out of draft — the blocker is resolved and the shard improves 8 → 2.**

The cross-backend failure (`Tensor / jaxlib._jax.ArrayImpl` inside `_vmomentdensity`) turned out not to be about the units contract at all: a df **constructed** under one backend caches arrays of that backend, and meeting them from a second forced backend raises. Reusing one df object across backends is a separate hazard, so each arm of the test now builds its own.

`test_kingdf_setup_wunits` passes now too — its `nan` came from the same cached state, not from an independent numerical defect as I first classified it.

**Result on `tests/test_quantity.py` under `--backend jax --jit`:** 8 failed → **2 failed, 13 passed**.

The 2 that remain are a different mechanism — units arriving on **parameters** rather than through the return conversion (`TracerArrayConversionError` on a `float64[50,50]`, and `DynamicJaxprTracer * PrefixUnit`) — and are pre-existing rather than introduced here.

numpy regression: `test_quantity.py` + `test_sphericaldf.py` = **467 passed, 0 failed**.

PR #1309, opened 2026-08-13
actionAngleVerticalInverse: exact (ODE-based) point transformation

jobovy-agent:
Adds `use_pointtransform="exact"` to `actionAngleVerticalInverse`: instead of least-squares-fitting a Chebyshev polynomial, solve the ODE $\mathrm{d}\pi/\mathrm{d}x^A = v/v^A$ that the point transformation satisfies exactly (the equal-incomplete-action map onto the same-action = same-frequency harmonic-oscillator torus), and store the solution as the normalized map $x/x_{\rm max}$ vs. $x^A/x^A_{\rm max}$ on a fixed mesh, evaluated by spline (degree `exact_pt_spl_deg=5`). This mirrors the API of the spherical implementation on `aainv-spherical-fromscratch`; the polynomial and no-PT paths are untouched (regression-checked). First PR of phase 1 of the inverse action-angle project (see `fast-orbits` repo, `PR_PLAN.md`).

Design notes (deviations from the spherical `"exact"` implementation, for review):
- **Normalized storage**: the map is stored on a fixed normalized mesh (like the polynomial normalization) rather than as `(sol.t, sol.y)` pairs, which is what makes interpolation-in-E of the exact PT work with the existing `_setup_interp` machinery unchanged.
- **Single vectorized ODE solve**: all tori are integrated in one `solve_ivp` call over the shared normalized coordinate (DOP853, rtol=atol=1e-12) — 1001 tori in a few seconds vs. ~0.5 s/torus separately.
- **Turning-point handling**: the ODE is singular (0/0) at $x^A = x^A_{\rm max}$ itself (exactly, since $x^A_{\rm max} = \sqrt{2E^A}/\omega$), so integration stops one mesh point short and the endpoint is fixed by the turning-point↔turning-point boundary condition; the map is smooth through the turning point, so the spline stays accurate over the last interval.
- **Closed-form angle grid**: with the exact PT, $\theta^A(x^A) = \arcsin(x^A/x^A_{\rm max})$-type, so the grid is seeded in closed form and Newton–Raphson only polishes it to stay exactly consistent with the stored spline (converges in ~1–2 iterations).

Key numbers (IsothermalDiskPotential, amp=1, σ=0.5, nta=512):
- $\max|nS_n|$ = 1.5e-12 / 5.1e-12 / 8.7e-11 for E = 0.1 / 1 / 10 — all mapping coefficients vanish to the ODE-solution floor, vs. O(1e-2–1e-1) with no PT (this is asserted by a new test).
- Energy along the mapped torus conserved to ≲1.6e-10 relative (the floor is set by the ODE rtol; the polynomial PT reaches machine precision only because its imperfection is absorbed self-consistently by the $S_n$).
- Orbit vs. direct integration over one period: max|Δx| = 2.8e-10.
- Interpolation (1001-tori grid): example-orbit round trip to ~1e-11, energy along interpolated torus ~4e-10 — i.e., the exact PT interpolates as well as no-PT (unlike the polynomial PT, which degrades interpolation ~20×: 2.2e-4 vs 9.7e-6 vs 1.3e-5 energy error on a 21-tori test grid).
- Setup: 0.09 s single torus; 77 s for the 1001-tori interpolation grid; evaluation ~30 ms per 1001 angles.

Tests: 8 new tests mirroring the polynomial-PT set (inverse-of-forward, frequencies, orbit, bisection, interpolation ×3) plus the coefficient-floor test; full `actionAngleVerticalInverse` selection passes 32/32 in ~110 s. The single-torus frequency test uses tol 1e-9 instead of the polynomial path's 1e-10 (method floor, see comment in the test).

Per project agreement: no docs/HISTORY updates in this PR (docs at end of phase).


jobovy-agent:
Review addressed in the three new commits:

**`_ptxa_eval` simplification** — you were right: the per-row-spline + unique approach is gone. The exact PT (and its first two derivatives, now sampled at setup) is stored as spline-filtered (torus, mesh) grids and evaluated with a single `ndimage.map_coordinates` call, with the torus row index as the first coordinate — the same machinery `_setup_interp` uses. Bonus: interpolated tori now evaluate the PT directly at *fractional* row coordinates, so the interp path needs no per-call spline construction either. One subtlety: `ndimage`'s mirror boundary distorts the interpolant near the mesh edges — exactly the turning points where we evaluate — so the stored mesh is padded beyond [-1,1] with the setup spline's polynomial extension (verified to push the artifact below 1e-14 inside the used range).

**Benchmark old → new** (isothermal disk; per your request):
| workload | old | new |
|---|---|---|
| setup, 1 torus (nta=512) | 89 ms | 40 ms |
| setup, 3 tori | 241 ms | 81 ms |
| setup, 1001-tori interp grid | 60.2 s | 7.7 s |
| eval 1001 angles (single / interp) | 30.9 / 17.2 ms | 30.6 / 17.2 ms |

No regressions; evaluation time is dominated by the angle solve, not PT evaluation. All 32 tests pass (selection now 63 s vs 112 s).

**ODE tolerance** — checked: the coefficient floor tracks the tolerance (1e-10 → ~5e-9, 1e-12 → ~7e-11, 1e-13 → ~2e-11, then saturating at the spline-representation limit), so it's now user-settable via `exact_pt_tol=1e-12`.

**Phase end** — HISTORY.txt entry added and a short point-transformation section added to the `vertical_and_inverse` tutorial notebook (executed; the demo torus's largest coefficient comes out at 7e-12).

PR #1314, opened 2026-08-14
Rework the exact point transformation of actionAngleSphericalInverse

jobovy-agent:
Second layer of the phase-2 stack (on #1313): rework of the exact point transformation of `actionAngleSphericalInverse`, backporting the 1D design from #1309 and going further where the spherical geometry requires it.

**Design** (details in the commit message):
- The PT ODE dπ/dr^A = v_r/v_r^A and the Δψ (ζ′) shift integrate as one coupled system **parametrized by the auxiliary torus' eccentric anomaly η**, with y = sin²(φ/2) as the radial variable — regular at both turning points. In particular dΔψ/dη = L[1/π² − 1/r^A²]·G(η) *exactly* (the 1/v_r divergence cancels algebraically), eliminating the divergent quadrature that previously broke the ψ side.
- One vectorized `solve_ivp` for all tori; two-stage resampling (spline in η on its well-conditioned uniform grid → evaluation at the closed-form η(s^A)) onto a fixed padded mesh; storage as (torus, mesh) grids of the normalized mapping, its two derivatives, and Δψ; evaluation everywhere via spline-filtered `map_coordinates` with the torus row index as first coordinate (no per-call spline construction anywhere).
- The exact-PT angle grid is seeded by solving the isochrone Kepler equation in closed form. `exact_pt_tol` added; small-J tori get the identity mapping instead of `NotImplementedError`.

**Results** (LogarithmicHaloPotential tori, J_r = 0.016/0.079/0.254; the old-code baseline was ~1e-4 round trips with grid non-convergence, see #1313):
- max|nS_n| = 6e-14 – 1.6e-12; max|dS_n/dJ_r| ≤ 8e-10; max|dS_n/dL| ≤ 1e-9 — **all coefficient sets vanish**: the exact PT including ζ′ makes the (J,θ)→(J^A,θ^A) mapping the identity for spherical potentials, as in 1D (this had been an open prediction; it is now numerically confirmed and asserted by the tightened test).
- Round trips vs `actionAngleSpherical`: 1e-11 – 1.4e-9. Setup 0.3 s/torus.

**Known cosmetic issue**: a few grid points adjacent to the turning points emit a Newton warning and converge via the bisection fallback (the analytic derivative checks out against finite differences at 1e-7; to be revisited together with a full no-root-find grid).

**Debugging war story** for the record: the first port had an algebra slip in G (b(s−1) = √(b²+r^A²), not b+√(b²+r^A²) — a smoothly-varying 1.45× error) and the direct spline through the η-clustered samples was ill-conditioned near the turning points (E^A constant to 4e-11 in the interior but only 1e-6 near the ends) — both found by differential testing against reference integrations.

Remaining on this stack before the dev→main merge: spherical `pt_only`, the interpolation machinery (PR 3), notebooks/paper.

### DIFF
--- a/galpy/actionAngle/actionAngleSphericalInverse.py
+++ b/galpy/actionAngle/actionAngleSphericalInverse.py
@@ -14,7 +14,7 @@ import numpy
 from matplotlib import cm, gridspec, pyplot
 from matplotlib.ticker import NullFormatter
 from numpy.polynomial import polynomial
-from scipy import integrate, interpolate, optimize
+from scipy import integrate, interpolate, ndimage, optimize
 
 from ..potential import (
     IsochronePotential,
@@ -57,6 +57,7 @@ class actionAngleSphericalInverse(actionAngleInverse):
         pt_deg=7,
         pt_nra=301,
         exact_pt_spl_deg=5,
+        exact_pt_tol=1e-12,
         maxiter=100,
         angle_tol=1e-12,
         bisect=False,
@@ -244,6 +245,7 @@ class actionAngleSphericalInverse(actionAngleInverse):
             ):
                 self._pt_exact = True
                 self._exact_pt_spl_deg = exact_pt_spl_deg
+                self._exact_pt_tol = exact_pt_tol
             else:
                 self._pt_exact = False
                 self._exact_pt_spl_deg = None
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Heads-up for review: definition-of-done checking (frequency-rate traversal vs orbit integration) exposed evaluation bugs in the exact PT as implemented here — Δψ is branch-dependent (odd around the radial turning points) but was evaluated single-valued, the node-passage u-flip used the wrong velocity, and retrograde orbits lose the sign of jφ in the reconstruction. Single-point round trips and energy checks can't see any of these, which is why this PR's tests pass. The fixes + regression tests are the first commit of #1318 (kept separate for reviewability rather than rebased into this layer, to avoid re-triggering the whole stack's CI). Recommend merging the stack only through #1318.

jobovy (approved):

jobovy (approved):

PR #1316, opened 2026-08-14
interpRZPotential: stop the bit-for-bit grid guard falling back on every backend build (58x)

jobovy-agent:
`interpRZPotential._grid_eval` accepts its vectorised whole-mesh result only if
it reproduces the scalar path **bit for bit**. That holds on numpy. On jax/torch
a whole-mesh reduction associates differently from a scalar one, so the check
fails on ~2 of 9 spot cells over a **1.6e-15** difference — making the fallback
to a cell-by-cell loop **unconditional on every backend grid build**.

Measured on a 201x201 `MWPotential` build (`interpRforce` + `interpzforce`):

| | |
|---|---|
| constructor, numpy | 0.09 s |
| constructor, jax | **136.70 s** (1643x) |
| cProfile | 230.5 s of 233.5 s is `_loop()`, **80838 scalar evaluator calls** |

## The fix is backend-aware, not laxer

The guard catches a real defect: `AnySphericalPotential` neither raises nor
broadcasts correctly and silently disagrees in **76 of 77 cells, worst 9.4e-01**.
A relative tolerance still catches that by ~13 orders; it is only ULP
reassociation it must stop rejecting.

- **numpy → `rtol = 0.0`**, i.e. the existing `array_equal`, unchanged. Numpy
  path byte-identical by construction.
- **backend arrays → `rtol = 1e-14`**, ~6x the observed 1.6e-15 spot difference.

Detection uses `is_backend_array(raw)` on the evaluator's **raw** result, not on
`grid` — `grid` has already been through `numpy.asarray`, so asking it would
always answer "numpy" and the branch would be silent dead code.

## Why 1e-14 and not 1e-12 (rewritten 2026-08-17)

The first version used `1e-12`, argued as "600x the observed difference". That
margin was argued from one side only: what matters is how far *below* the
smallest real defect it stays, and equally that it remains a **reassociation**
allowance rather than a **correctness** allowance. It is now pinned from both
sides by a test (below), and `1e-12` fails that test.

## The test that was failing was encoding the OLD contract

`test_interpRZPotential_grid_falls_back_for_nonbroadcasting_potential` failed on
the jax shard through two rounds of tightening. No value of `rtol` could have
fixed it, because the test asserted `numpy.array_equal` against the cell-by-cell
path — the very bit-exactness this PR deliberately stops requiring.

It was stable on `develop` only because the guard *also* compared with
`array_equal`: guard-criterion and test-criterion were the **same predicate**, so
any ULP difference forced a fallback, and a fallback returns the cell-by-cell
loop by construction. The test therefore could not fail, and could not measure
the guard either. Measured on the test's own grid:

| | whole-mesh vs cell-by-cell |
|---|---|
| numpy | 76 of 77 cells differ, worst 9.4e-01 → genuinely non-broadcasting |
| jax | **0 of 77 cells differ** → broadcasts correctly |

So the potential the test is named for is only non-broadcasting on numpy, and
only numpy (`rtol=0`) ever sees the old behaviour. Passing in isolation and
failing in the shard was CI's floating point differing from mine.

**Restructured accordingly:**

* the spot-check fallback now has a **synthetic-evaluator test**, matching how
  its two sibling fallbacks (bad shape, raises) are already covered in this
  file. It injects a **1e-13** relative error — 10x above the allowance so it
  must be caught, ~60x above the 1.6e-15 noise the allowance exists to
  tolerate — in two shapes: a whole-array offset, and a single wrong cell at a
  location `_spot_check_cells` is known to visit, which pins the sample/guard
  coupling rather than the guard alone. It multiplies by a **namespace-matched**
  mask: returning numpy would send the guard down its `rtol=0` branch and
  quietly stop testing the tolerance arm at all.
* the real-potential test asserts the contract that actually holds — the shipped
  grid does not deviate from the scalar path by more than the guard permits:
  bit for bit on numpy, within 1e-14 on a backend.

## Results

| check | result |
|---|---|
| jax constructor | **136.70 s → 2.34 s (58x)** |
| `AnySphericalPotential` still falls back (numpy) | **2/2 grids** |
| jax-built vs numpy-built grid | max rel **1.3e-13**, **median 0.0** |
| `tests/test_interp_potential.py`, numpy | 45 passed |
| `tests/test_interp_potential.py`, jax | 40 passed, 6 skipped, **0 failed** |
| `tests/test_interp_potential.py`, torch | 45 passed |

A/B of the new test against the guard's tolerance — the reason to believe any
of the above:

| guard `rtol` | numpy | jax | torch |
|---|---|---|---|
| **1e-14 (this PR)** | pass | pass | pass |
| 1e-12 (deliberately broken) | pass | **FAIL** | **FAIL** |

numpy passing on both arms is the point, not a gap: it takes the `rtol=0` path
regardless, which confirms the new test is aimed at the backend arm.

## Coverage

Does not drop. The spot-check fallback keeps its numpy coverage via
`AnySphericalPotential` and **gains** deterministic coverage on jax/torch, where
it previously had none.

## Ledger

One slow-skip entry removed (`tests/backend_slow_skip.txt`, 1 deletion): the
entry that existed only because the grid build was slow. No xfail delta.


jobovy-agent:
Gate re-confirmed on the final head `9def8d2e` (i.e. **after** the `_grid_tol` change, not just before it):

| arm | result |
|---|---|
| numpy | 45 passed |
| `--backend jax` | 39 passed, 6 skipped (was 1 failed) |
| `--backend torch` | **79 passed, 1 skipped** |

The torch arm was previously only measured *before* the tolerance change; relaxing a bound can only help, but that is reasoning rather than measurement, so it is now measured on the actual head.

Scope correction to the PR description: I had suggested this pattern might affect a large share of the 85 slow-skip entries. It does not — `_grid_eval` has 7 call sites, **all inside `interpRZPotential.py`**, and `array_equal` appears in no other galpy module. So this is a genuine 58x on a real defect covering that file's 7 grid quantities and the 6 `test_interp_potential` entries, **not** a campaign-wide fix.

jobovy-agent:
**Converting to draft — the tolerance approach is unsound, not merely mistuned.**

CI caught a real failure my local gate missed: `test_interp_potential.py` asserts that for a **non-broadcasting** potential the guard must fall back, so the grid must be **bit-identical** to the cell-by-cell reference. With `rtol=1e-12` the guard accepts a vectorised result that is genuinely wrong on a log-R 11×7 grid.

My local evidence ("AnySphericalPotential still falls back 2/2") was **grid-specific** — I verified one sampling geometry and generalised it to the guard's whole domain. The spot check is probabilistic by design; widening the acceptance window makes that probabilism load-bearing.

Measured, on both potentials and both grid geometries (9 sampled cells, jax):

| case | vs scalar | vs 1×1-mesh |
|---|---|---|
| AnySpherical, logR 11×7 (the CI grid) | 0 differ | 0 differ |
| AnySpherical, linear 41×41 | 0 differ | 0 differ |
| MWPotential (correct broadcaster) | **2 differ** | **2 differ** |

So the *mis-broadcasting* potential agrees with scalar at sampled cells while the *correct* one disagrees — the two error populations **overlap**. No magnitude threshold can separate a broadcasting bug from float reassociation, and the 1×1-mesh reference I was going to try next doesn't separate them either.

Retuning to 1e-13/1e-14 would only shift which grids fail — the same aliasing mistake as chasing a moving spike with a coarser scan.

The 1643× fallback is still a real defect worth fixing, but it needs a **mechanism** that distinguishes the two cases structurally (or a cheaper fallback that preserves the exact guard), not a looser comparison. Reworking rather than retuning.

jobovy-agent:
Third approach refuted — recording so this isn't re-attempted.

**Row-wise fallback** (one vectorised call per row, keeping the bit-for-bit guard untouched) fails on both counts:

```
AnySpherical logR 11x7   rowwise 2.15s  cellwise 1.98s  speedup 0.9x  bit-identical=True
MWPotential  21x21       rowwise 1.28s  cellwise 1.73s  speedup 1.3x  bit-identical=False
```

- **Not bit-identical** for the correctly-broadcasting potential — a row is still a vectorised call, so it reassociates just like the full mesh. Same contract break as the tolerance, just smaller.
- **Negligible speedup**: 1.3×, not the ~20× that 21×-fewer-calls would suggest. A row call costs ~15× a scalar call, so the per-call overhead is not flat at this size and batching rows buys little.

Approaches tried and refuted, all by measurement:

| approach | why it fails |
|---|---|
| relative tolerance (1e-12) | mis-broadcast and reassociation error populations **overlap** — no threshold separates them |
| 1×1-mesh spot-check reference | identical differ-counts to scalar; doesn't discriminate |
| row-wise fallback | breaks bit-identity for good potentials; ~1.3× only |

The 1643× fallback remains a real defect, but it needs something structural — e.g. asking the potential whether it broadcasts (a capability flag) rather than inferring it from output agreement. Parking here rather than trying a fourth variation on "compare and hope".

jobovy-agent:
Fourth approach checked and refuted — this closes the design space I can reach.

**Capability flag (`_backend_accepts_arrays`)** was the structural signal I proposed last comment. It exists, but cannot serve here:

- set `True` on exactly **two** potentials (`RotateAndTiltWrapperPotential`, `DoubleExponentialDiskPotential`);
- read as `getattr(self, "_backend_accepts_arrays", False)` — **opt-in, default False**;
- its meaning is *node-axis* broadcasting, not general (R,z) mesh broadcasting.

Keying `_grid_eval` off it would send almost every potential — including `MWPotential`, which broadcasts correctly — down the slow cell-by-cell path. Strictly worse than the status quo.

### Summary of the whole investigation

| approach | verdict |
|---|---|
| relative tolerance 1e-12 | **unsound** — mis-broadcast and reassociation error populations overlap |
| 1×1-mesh spot reference | no discrimination (identical differ-counts) |
| row-wise fallback | breaks bit-identity for good potentials; ~1.3× only |
| `_backend_accepts_arrays` flag | wrong semantics, ~2 adopters, default False |

**The defect is real and stays open**: on jax every `interpRZPotential` grid build falls back to a full cell-by-cell loop (measured 136.7 s vs numpy 0.09 s on a 201×201 MWPotential build; 409 s vs 0.04 s on a 151×151 6-quantity mock), because the bit-for-bit guard can never pass when the backend reassociates.

A real fix needs either a **broadcast-capability contract that potentials actually declare** (i.e. flip `_backend_accepts_arrays` to opt-out and audit every potential), or an `AnySphericalPotential`-side fix so it broadcasts correctly and the guard becomes unnecessary. Both are larger than this PR and are design calls rather than tweaks.

Leaving as draft with the measurements recorded rather than shipping a fifth guess.

jobovy-agent:
Status update — **this PR's approach is superseded; parked, not abandoned.**

The tolerance widening here was the wrong fix and CI refuted it (it passed my
gate but failed on a log-R grid the gate never used). The diagnosis inverted:
the **numpy** path was the broken one, not the backend path.

Root cause, now fixed upstream in **#1323**: `AnySphericalPotential` passed
`scipy.integrate.quad` an array limit at two sites, so array evaluation
silently collapsed to `r[0]` — up to **87.7 %** force error. `_grid_eval`'s
bit-for-bit guard was correctly catching that one genuinely mis-broadcasting
potential, and paying a **1643x** cell-by-cell fallback on *every* backend grid
build to do it (jax reassociates a whole-mesh reduction vs a scalar one, so the
guard can never match bit-for-bit on a backend).

Plan once #1323 lands on `main`:

1. cherry-pick it onto `feat/backends-develop` as its own commit, so a later
   rebase onto `main` drops it cleanly;
2. **delete** the `_grid_eval` guard rather than tune it. Its job — catching a
   mis-broadcast potential — is now done properly by
   `test_potential_array_input`, which checks whole-array against cell-by-cell
   at 1e-10 for all 8 methods across every potential, and runs under the
   backends too. `AnySphericalPotential` was on that test's *exclusion list*,
   which is precisely why this was invisible; #1323 removes it.
3. drop the tolerance commits here.

Note the cherry-pick will not apply cleanly: `develop` has since restructured
`_rawmass`/`_revaluate` into methods with backend branches, and its numpy
branches carry the same `atleast_1d(r).flatten()[0]` collapse.

PR #1347, opened 2026-08-19
actionAngleStaeckelInverse: direct single-torus inverse for Staeckel potentials

jobovy-agent:
First layer of the phase-3 stack (base: `aainv-staeckel-dev`, branched from main — independent of the in-prep spherical branch per the plan). Implements the direct separable-angle construction worked out and verified in fast-orbits (STAECKEL_MATH.md §12; Step-0 notebook, fast-orbits #9):

- Each angle is the sum of a u-profile and a v-profile (θᵢ = ∂W/∂Jᵢ), all six profiles computed as regular quadratures in the χ anomaly with analytic turning-point limits; actions and the 3×3 period matrix from the same machinery; evaluation inverts the additively separable 2×2 angle system by a damped, vectorized Newton iteration. No auxiliary torus, no generating function, no Fourier lattice; placement is exact by construction.
- Tori labeled by (E, L_z, I₃) with the Stäckel splitting extracted from the potential in the V(π/2) = 0 gauge (Step 0.1 conventions).
- Tests (all fast, ~5 s): actions/frequencies vs the forward `actionAngleStaeckel` at order=200 (1e-6/1e-8), 10-period two-branch traversal vs direct integration for benign / eccentric / near-shell (J_R/J_z ≈ 0.01, no special casing) / near-planar tori, δH/|E| < 1e-13 along the reconstructed torus, xvFreqs/Freqs consistency, wrong-action and invalid-torus error paths.

Next layers: actions → (E, I₃) inversion, Stäckel-ized targets (N2), interpolation (N3), units/docs (N4). The forward-code χ-form quadrature improvement goes to main separately.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelInverse.py
+++ b/galpy/actionAngle/actionAngleStaeckelInverse.py
@@ -13,14 +13,18 @@ from scipy.interpolate import InterpolatedUnivariateSpline
 from scipy.optimize import brentq
 
 from ..potential import OblateStaeckelWrapperPotential
-from ..util import conversion, coords
+from ..util import coords
 from .actionAngleInverse import actionAngleInverse
 
-# Nodes/weights for composite 10-point Gauss-Legendre quadrature: applied per
-# interval of the nchi-point chi mesh, the error per panel is
+# Nodes/weights for composite 10-point Gauss-Legendre quadrature: applied
+# per interval of the nchi-point chi mesh, the error per panel is
 # O((pi/nchi)^20), so every stored integral is at machine precision for any
 # reasonable nchi; there is no accuracy knob to tune
 _GLX, _GLW = numpy.polynomial.legendre.leggauss(10)
+# Reference u of the internal Staeckel splitting for potentials that are not
+# already wrapped; its value is irrelevant for an exactly Staeckel potential
+# (any choice gives the same U, V up to the fixed gauge)
+_U0INTERNAL = 1.15
 
 
 class actionAngleStaeckelInverse(actionAngleInverse):
@@ -39,8 +43,6 @@ class actionAngleStaeckelInverse(actionAngleInverse):
     def __init__(
         self,
         pot=None,
-        delta=None,
-        u0=None,
         Es=[0.5],
         Lzs=[0.5],
         I3s=[0.1],
@@ -55,22 +57,11 @@ class actionAngleStaeckelInverse(actionAngleInverse):
         Parameters
         ----------
         pot : Potential or list thereof
-            The potential; must be of exact Staeckel form in prolate
-            spheroidal coordinates with focal distance delta. Pass an
-            OblateStaeckelWrapperPotential to work with the Staeckel
-            approximation of a general axisymmetric potential (support for
-            torus-dependent delta in that case is planned).
-        delta : float or Quantity, optional
-            Focal distance of the prolate spheroidal coordinate system;
-            derived from the potential when possible (e.g., from a
-            KuzminKutuzovStaeckelPotential or an
-            OblateStaeckelWrapperPotential), otherwise required.
-        u0 : float, optional
-            Reference u of the Staeckel splitting (as in
-            OblateStaeckelWrapperPotential); irrelevant for an exactly
-            Staeckel potential (any value gives the same U, V up to the
-            fixed gauge), taken from the potential when passing an
-            OblateStaeckelWrapperPotential.
+            The potential; must be of exact Staeckel form and supply its own
+            focal distance: either an OblateStaeckelWrapperPotential (also
+            the route for the Staeckel approximation of a general
+            axisymmetric potential) or a potential with a _delta attribute
+            such as KuzminKutuzovStaeckelPotential.
         Es : list of float
             Energies of the tori to set up.
         Lzs : list of float
@@ -92,32 +83,32 @@ class actionAngleStaeckelInverse(actionAngleInverse):
         - Angle conventions match those of the forward actionAngleStaeckel.
         - 2026-08-19 - Started - Bovy (UofT)
         """
+        if "delta" in kwargs or "u0" in kwargs:
+            raise TypeError(
+                "actionAngleStaeckelInverse does not accept delta= or u0=: "
+                "the potential itself supplies the focal distance (pass an "
+                "OblateStaeckelWrapperPotential or a potential with a "
+                "_delta attribute)"
+            )
         actionAngleInverse.__init__(self, **kwargs)
         if pot is None:  # pragma: no cover
             raise OSError("Must specify pot= for actionAngleStaeckelInverse")
         self._pot = pot
         if isinstance(pot, OblateStaeckelWrapperPotential):
-            if delta is not None or u0 is not None:
-                raise ValueError(
-                    "When pot is an OblateStaeckelWrapperPotential, delta and "
-                    "u0 are taken from the potential and cannot be specified"
-                )
             self._staeckelwrap = pot
-            self._delta = pot._delta
-            self._u0 = pot._u0
         else:
-            if delta is None:
-                delta = getattr(pot, "_Delta", None)
+            delta = getattr(pot, "_delta", None)
             if delta is None:
                 raise OSError(
-                    "Must specify delta= for actionAngleStaeckelInverse when "
-                    "it cannot be derived from the potential"
+                    "actionAngleStaeckelInverse requires a potential of "
+                    "Staeckel form that supplies its focal distance (e.g., "
+                    "KuzminKutuzovStaeckelPotential); wrap a general "
+                    "axisymmetric potential in OblateStaeckelWrapperPotential"
                 )
-            self._delta = conversion.parse_length(delta, ro=self._ro)
-            self._u0 = 1.15 if u0 is None else u0
             self._staeckelwrap = OblateStaeckelWrapperPotential(
-                pot=pot, delta=self._delta, u0=self._u0
+                pot=pot, delta=delta, u0=_U0INTERNAL
             )
+        self._delta = self._staeckelwrap._delta
         self._Es = numpy.atleast_1d(numpy.array(Es, dtype="float"))
         self._Lzs = numpy.atleast_1d(numpy.array(Lzs, dtype="float"))
         self._I3s = numpy.atleast_1d(numpy.array(I3s, dtype="float"))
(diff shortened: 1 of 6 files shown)


galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+    ):
+        """
+        Initialize an actionAngleStaeckelInverse object.
+
+        Parameters
+        ----------
+        pot : Potential or list thereof
+            The potential; must be of exact Staeckel form in prolate
+            spheroidal coordinates with focal distance delta (e.g.,
+            KuzminKutuzovStaeckelPotential).
```
jobovy: Could mention ``OblateStaeckelWrapper`` here as maybe a better choice? Eventually I think we'll need to somehow support ``OblateStaeckelWrapper`` versions of non-Staeckel potentials where delta may depend on (E,Lz,I3)

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        delta : float
```
jobovy: Could we just derive delta rather than specifying it when an actual Staeckel potential is passed? Maybe forward-looking it's best to keep it as an input. Or allow both!

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        Lzs : list of float
+            z-components of the angular momentum of the tori.
+        I3s : list of float
+            Third integrals of the tori (in the convention
+            p_u^2 = 2 delta^2 [E sinh^2 u - U(u) - I_3] - L_z^2/sinh^2 u
+            with the gauge V(pi/2) = 0).
+        nchi : int, optional
+            Number of grid points in the chi anomaly for the stored
+            profiles.
+        uref : float, optional
```
jobovy: Maybe ``u0`` to be consistent with https://docs.galpy.org/en/v1.11.2/reference/potentialoblatestaeckelwrapper.html?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+            Reference u used in the Staeckel splitting of the potential.
+        umax : float, optional
```
jobovy: Is it necessary to specify this and not just set it to a large number?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+            Maximum u scanned for the u turning points.
+        nscan : int, optional
```
jobovy: Again, can't this just be set automatically? Does it need to be tweakable by a user?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        maxiter : int, optional
+            Maximum number of Newton iterations in the angle inversion.
+        angle_tol : float, optional
+            Convergence tolerance of the angle inversion.
+
+        Notes
+        -----
+        - Angle conventions: theta_R = 0 at u = u_- moving outward;
+          theta_z = 0 at v = v_- (the northern turning point) moving toward
+          the midplane; theta_phi follows from dW/dL_z.
```
jobovy: Are these the same conventions as used elsewhere in galpy? If so, no need to mention them, if not, make them consistent!

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        - 2026-08-19 - Started - Bovy (UofT)
+        """
+        actionAngleInverse.__init__(self, *[], **kwargs)
```
jobovy: Not just ``actionAngleInverse.__init__(self, **kwargs)``? That's what you did in #1339

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        self._Bprof = []  # per torus: 3 v-profile splines B_i(chi_v)
+        self._dAprof = []
+        self._dBprof = []
+        for ii in range(self._ntori):
+            self._setup_torus(ii)
+
+    ############################ STAECKEL SPLITTING ###########################
+    def _phi_uv(self, u, v):
+        R = self._delta * numpy.sinh(u) * numpy.sin(v)
+        z = self._delta * numpy.cosh(u) * numpy.cos(v)
```
jobovy: Use ``util.coords`` functions here? Or is that substantially slower? Also, I think a version of this code might already exist in ``actionAngleStaeckel``?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        return evaluatePotentials(self._pot, R, z, use_physical=False)
+
+    def _Uofu(self, u):
+        return self._phi_uv(u, numpy.pi / 2.0) * (numpy.sinh(u) ** 2.0 + 1.0)
+
+    def _Vofv(self, v):
+        return self._Vref - self._phi_uv(self._uref, v) * (
+            numpy.sinh(self._uref) ** 2.0 + numpy.sin(v) ** 2.0
+        )
```
jobovy: Might be able to tap into ``OblateStaeckelWrapperPotential`` to give some of these functions without duplication (maybe they need to become class methods?).

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+                f"Midplane not reached for torus {ii}; no valid torus for "
+                "(E={E}, Lz={Lz}, I3={I3})"
+            )
+        neg = numpy.where(Wv(vs) < 0.0)[0]
+        vmin = (
+            brentq(Wv, vs[neg[-1]], numpy.pi / 2.0, xtol=1e-15, rtol=8.9e-16)
+            if len(neg) > 0
+            else 1e-8
+        )
+        self._umins[ii], self._umaxs[ii], self._vmins[ii] = umin, umaxx, vmin
```
jobovy: Again, seems quite similar to code that must be in ``actionAngleStaeckel``, would be best to share code.

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+###############################################################################
+import numpy
+from scipy.interpolate import InterpolatedUnivariateSpline
+from scipy.optimize import brentq
+
+from ..potential import evaluatePotentials
+from ..util import conversion
+from .actionAngleInverse import actionAngleInverse
+
+_GLX, _GLW = numpy.polynomial.legendre.leggauss(10)
```
jobovy, line 23: This is good enough?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,387 @@
...
+        wv = numpy.where(pv >= 0.0, chiv, 2.0 * numpy.pi - chiv)
+        A, B = self._Aprof[ii], self._Bprof[ii]
+        thR = self._fold(A[0], wu) + self._fold(B[0], wv)
+        thz = self._fold(A[1], wu) + self._fold(B[1], wv)
+        thphi = phi + self._fold(A[2], wu) + self._fold(B[2], wv)
+        return (
+            thR % (2.0 * numpy.pi),
+            thphi % (2.0 * numpy.pi),
+            thz % (2.0 * numpy.pi),
+        )
```
jobovy: Isn't this just ``actionAngleStaeckel``??? Either way, don't just include code that's used for tests!

jobovy (changes requested):

jobovy-agent:
All review points addressed in 1a648215:

- **Wrapper reuse (L58/L139/L182)**: the Stäckel splitting now delegates to `OblateStaeckelWrapperPotential._U/_V` — an internal wrapper is created for plain Stäckel potentials, and a passed-in wrapper is used directly (its δ and u0 taken over; specifying them too is an error). This removes the U/V duplication and sets up N2's torus-dependent-δ support. The remaining W_u/W_v momentum profiles don't exist in `actionAngleStaeckel` (it works in (λ,ν) with its own forms), so those stay.
- **δ derivation (L59)**: `delta=None` now derives from the potential (`_Delta` on KuzminKutuzov, the wrapper's `_delta`); explicit input still allowed for forward-compatibility.
- **u0 (L72)**: renamed, semantics documented (irrelevant for exact Stäckel potentials).
- **umax/nscan (L74/L76)**: gone — one dense vectorized bracketing scan (cheap: a single vectorized potential evaluation), with informative errors for unbound/not-enclosed/invalid tori.
- **Angle conventions (L87)**: measured against the forward code: θ_R and θ_φ already agreed exactly; θ_z differed by exactly π/2 (upward-midplane-crossing vs northern-turning-point zero). Now matched to the forward convention, the Notes disclaimer removed, and a two-code round-trip test added (forward angles → inverse → original point, 1e-6).
- **init call (L90)**, **coords reuse (L130)**: done.
- **Quadrature order (L19)**: comment added — composite 10-point Gauss-Legendre per χ-mesh panel has error O((π/nchi)^20), machine precision with no knob.
- **Test-only forward method (L387)**: removed; tests use the forward `actionAngleStaeckel` (order=200) for θ₀, which the convention alignment makes possible.
- **Coverage**: now 100% of the module (the coverage run also exposed that two error tests were hitting the wrong branches — fixed with labels that provably hit each).

Test-walkthrough notebook coming in fast-orbits next.

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+            Focal distance of the prolate spheroidal coordinate system;
+            derived from the potential when possible (e.g., from a
+            KuzminKutuzovStaeckelPotential or an
+            OblateStaeckelWrapperPotential), otherwise required.
+        u0 : float, optional
+            Reference u of the Staeckel splitting (as in
+            OblateStaeckelWrapperPotential); irrelevant for an exactly
+            Staeckel potential (any value gives the same U, V up to the
+            fixed gauge), taken from the potential when passing an
+            OblateStaeckelWrapperPotential.
```
jobovy: Given that this method requires either an exact Staeckel potential or an ``OblateStaeckelWrapperPotential``, isn't u0 always irrelevant? That is, either it is irrelevant because the potential is Staeckel or the ``OblateStaeckelWrapperPotential`` can supply it? Even for a ``OblateStaeckelWrapperPotential`` it should be irrelevant. So I think it can be dropped.

Same really applies to delta: if the potential coming in has to be of staeckel form, then it should have delta (or be derivable with estimateDeltaStaeckel, but really should just have it)

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+                    "When pot is an OblateStaeckelWrapperPotential, delta and "
+                    "u0 are taken from the potential and cannot be specified"
+                )
+            self._staeckelwrap = pot
+            self._delta = pot._delta
+            self._u0 = pot._u0
+        else:
+            if delta is None:
+                delta = getattr(pot, "_Delta", None)
+            if delta is None:
```
jobovy: To simplify the code when we drop the delta input (as I think we should, at least for now until it's clear it's necessary), I think we can rename the internal KuzminKutuzov _Delta _delta.

Maybe one option is to demand that the input potential is a ``OblateStaeckelWrapperPotential`` (again for now, maybe later we want to support a general potential that is ``OblateStaeckelWrapperPotential``-ed).

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+        self._OmegaR = numpy.empty(self._ntori)
+        self._Omegaz = numpy.empty(self._ntori)
+        self._Omegaphi = numpy.empty(self._ntori)
+        self._dEI3Lz_dJ = numpy.empty((self._ntori, 3, 3))
+        self._Aprof = []  # per torus: 3 u-profile splines A_i(chi_u)
+        self._Bprof = []  # per torus: 3 v-profile splines B_i(chi_v)
+        self._dAprof = []
+        self._dBprof = []
+        for ii in range(self._ntori):
+            self._setup_torus(ii)
```
jobovy: Can't be vectorized?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+
+    ############################ STAECKEL SPLITTING ###########################
+    # U(u) and V(v) in the gauge V(pi/2) = 0, delegated to
+    # OblateStaeckelWrapperPotential (exact when the potential is Staeckel)
+    def _Uofu(self, u):
+        return self._staeckelwrap._U(u)
+
+    def _Vofv(self, v):
+        return self._staeckelwrap._V(v)
```
jobovy: Doesn't this seem a bit overkill? Could just be defined as lambda's if necessary, but could also be directly called where necessary (just call self._staeckelwrap._U(u), or if we demand that the potential is ``OblateStaeckel...``, we can just use its internal functions directly

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+        if pos[-1] == len(us) - 1:
+            raise ValueError(
+                f"u oscillation not enclosed for torus {ii} "
+                f"(E={E}, Lz={Lz}, I3={I3}); the torus is likely unbound"
+            )
+        ulo = us[pos[0] - 1] if pos[0] > 0 else 1e-8
+        umin = brentq(Wu, ulo, us[pos[0]], xtol=1e-15, rtol=8.9e-16)
+        umaxx = brentq(Wu, us[pos[-1]], us[pos[-1] + 1], xtol=1e-15, rtol=8.9e-16)
+        vs = numpy.linspace(1e-3, numpy.pi / 2.0, 4000)
+        if Wv(numpy.pi / 2.0) <= 0.0:
```
jobovy: Maybe some of this could be vectorized using similar approaches as used in sphericalInverse?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,382 @@
...
+                self._chi,
+                Minv[0, i] * PEv + Minv[1, i] * PIv - Minv[2, i] * PLv,
+                k=5,
+            )
+            for i in range(3)
+        ]
+        self._Aprof.append(A)
+        self._Bprof.append(B)
+        self._dAprof.append([a.derivative() for a in A])
+        self._dBprof.append([b.derivative() for b in B])
```
jobovy: Maybe this complicated setup code should be broken up into simpler, logical pieces?

jobovy (changes requested):

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — the `u0=` input is gone. When the potential is not already wrapped, the internal auto-wrap uses a fixed module constant (`_U0INTERNAL = 1.15`), with a comment noting that its value is irrelevant for an exactly Staeckel potential (any choice gives the same U, V up to the fixed gauge).

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — `delta=` (and `u0=`) are dropped entirely: the potential must supply its own focal distance, either as an `OblateStaeckelWrapperPotential` or through a `_delta` attribute. To make the latter work I renamed `KuzminKutuzovStaeckelPotential._Delta` → `_delta` (including the C argument packing in `integratePlanarOrbit`/`integrateFullOrbit`; KK C orbit integration verified). Any other potential raises an `OSError` instructing to wrap it in `OblateStaeckelWrapperPotential`, and stray `delta=`/`u0=` kwargs raise a `TypeError` rather than being silently swallowed by `**kwargs`.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — the turning-point stage is now fully vectorized across tori: one `U(us)`/`V(vs)` evaluation on the shared scan meshes, with `W` broadcast to `(ntori, nmesh)`; the per-torus remainder is just the `brentq` refinements and the error checks.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — `_Uofu`/`_Vofv` are removed; `_Wu`/`_Wv` call `self._staeckelwrap._U`/`._V` directly.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — all six cumulative profiles and both action integrals are now computed as `(ntori, npanel, 10)` array operations with a single potential evaluation per degree of freedom (the six integrands share the same `sqrt(W)` at the same Gauss nodes, so they're all derived from one broadcast momentum evaluation), and the period matrices are inverted in one batched `numpy.linalg.inv` call. Only the spline construction remains a per-torus loop, as in `sphericalInverse`.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — the setup is now three logical stages called in sequence from `__init__`: `_find_turning_points` (vectorized bracket scan + brentq refinement), `_compute_actions_frequencies_profiles` (the quadratures, period matrices, and frequencies), and `_build_angle_profile_splines` (the six per-torus angle-profile splines).

jobovy-agent:
Second review round addressed in af53e2d6 (point-by-point replies inline). Summary: `delta=`/`u0=` inputs dropped (the potential supplies the focal distance; `KuzminKutuzovStaeckelPotential._Delta` renamed to `_delta`), `_Uofu`/`_Vofv` removed, and the setup is now decomposed into three logical stages with the turning-point scan and all quadrature profiles vectorized across tori. Tests updated for the new API; module coverage remains 100%; KK C orbit integration verified after the rename.

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -130,161 +121,203 @@ def __init__(
...
+        self._umaxs = numpy.empty(self._ntori)
+        self._vmins = numpy.empty(self._ntori)
+        for ii in range(self._ntori):
+            E, Lz, I3 = self._Es[ii], self._Lzs[ii], self._I3s[ii]
+            pos = numpy.where(Wu_all[ii] > 0.0)[0]
+            if len(pos) == 0:
+                raise ValueError(
+                    f"No bound u oscillation found for torus {ii} "
+                    f"(E={E}, Lz={Lz}, I3={I3})"
+                )
```
jobovy: Isn't this a strong statement caused by setting the max u=40? Shouldn't we step out to find a bracket to search, like in the rperi/rap code in actionAngleSpherical (and maybe how this is done in actionAngleStaeckel; I'm again wondering whether not more of this can be shared with existing actionAngleStaeckel machinery)

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -130,161 +121,203 @@ def __init__(
...
-        a, b = self._chi[:-1], self._chi[1:]
-        mid, half = 0.5 * (a + b), 0.5 * (b - a)
+    ############################ SETUP: QUADRATURES ###########################
+    def _compute_actions_frequencies_profiles(self):
+        """Actions, the 3x3 period matrices, frequencies, and the cumulative
+        angle profiles of all tori, all from two vectorized potential
+        evaluations (one per degree of freedom): the momenta at the
+        composite Gauss-Legendre nodes of the shared chi mesh give both the
+        action integrands sqrt(W) sin(chi) and the 1/p profile integrands
+        f(q) sin(chi)/sqrt(W)"""
```
jobovy: Again, I think a lot of this could ideally be done with actionAngleStaeckel, much like sphericalInverse uses actionAngleSpherical (and similar for the 1D vertical code). Maybe the actionAngleStaeckel code needs to be improved?

jobovy (changes requested):

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done in 84cf4b0c — the scan mesh is now only a *seed* for the brackets, with no hard limits in either direction: if the positive region extends beyond the mesh, the outer bracket steps out geometrically like `_rapRperiAxiFindStart` in `actionAngleSpherical` (NaN-robust, since e.g. KuzminKutuzov's coordinate algebra produces NaNs beyond u≈19), and if *no* mesh sample is positive, the maximum of W_u is refined with a bounded scalar minimization before declaring that there is no bound oscillation — an oscillation narrower than the mesh spacing (an ultra-thin near-shell torus) is now found instead of erroring. Note that u is logarithmic in R, so the seed mesh's u=40 already corresponds to R ~ delta*e^40 ≈ 10^17 length units; the stepping is a safety net rather than a practically-reachable path for bound orbits.

While testing this with a genuinely ultra-thin torus (double-root labels perturbed by dE=1e-9, u-width 4e-5) I found and fixed a real numerical problem it exposes in the profile quadratures: the direct evaluation of W near the turning points is cancellation-dominated for thin tori, and the 1/sqrt(W) profile integrands blew up catastrophically. The quadratures now use the regular ratio Q = W/[y(1-y)] with analytic turning-point limits |W'|·D (via the wrapper's `_dUdu`/`_dVdv`) switched in near the endpoints — the same masking as in the 1D/spherical quadrature code. The new test sets up the dE=1e-9 torus through the refinement path and checks that its frequencies connect continuously to a nearby resolved torus; coverage remains 100%.

On sharing with `actionAngleStaeckel`'s umin/umax machinery specifically: its `_uminUmaxFindStart` steps from the *orbit's current u*, which the inverse doesn't have (only the torus labels), and its momentum function carries the forward code's I3U convention and 11-argument u0 bookkeeping rather than the wrapper's U/V splitting — so direct reuse would couple more than it shares. The path to genuine sharing is the forward-code improvement below.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Agreed — and I think the sharing has to flow in the direction you suggest at the end: improve `actionAngleStaeckel` first, then consume it here. Two things prevent the inverse from calling today's forward code the way `sphericalInverse` leans on `actionAngleSpherical`: (1) the inverse needs *cumulative* profiles (the six incomplete integrals as functions of the anomaly), which the forward code never computes — it only produces the complete integrals; and (2) the forward code's fixed-order Gauss-Legendre on the singular integrand form is its accuracy bottleneck: it converges to the chi-regularized values only around `order=200` (the tests here use exactly that as ground truth), so sharing its quadrature as-is would cap the inverse at ~1e-4 instead of machine precision.

My proposal (which you tentatively green-lit earlier) is a dedicated forward-code PR right after this one: switch `actionAngleStaeckel`'s u/v integrals to the chi-anomaly composite rule (machine-converged at fixed low cost, removing the order knob), base its momenta on the same U/V splitting as `OblateStaeckelWrapperPotential`, and factor the shared pieces — the W_u/W_v momentum functions, the analytic dW derivatives, the turning-point bracketing, and the composite-quadrature helper — into a common home that both the forward code and this module import. That keeps this PR within the size guideline and upgrades the forward code's accuracy at the same time. Happy to reorder if you'd rather see the shared core land first and rebase this PR on top of it.

jobovy-agent:
Third review round addressed in 84cf4b0c: adaptive turning-point bracketing (mesh is only a seed; geometric step-out + thin-oscillation refinement, no hard limits), plus a numerical-stability fix this uncovered — the profile quadratures now use the regular Q = W/[y(1-y)] form with analytic turning-point limits, keeping ultra-thin near-shell tori stable. New thin-shell test; coverage stays 100%. The actionAngleStaeckel code-sharing question is answered inline with a concrete proposal for the follow-up forward-code PR.

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -0,0 +1,494 @@
...
+            J10, J11 = self._dfold(dA[1], wu), self._dfold(dB[1], wv)
+            det = J00 * J11 - J01 * J10
+            dwu = (J11 * f0 - J01 * f1) / det
+            dwv = (-J10 * f0 + J00 * f1) / det
+            step = numpy.maximum(numpy.fabs(dwu), numpy.fabs(dwv))
+            lim = numpy.minimum(1.0, 0.5 / numpy.maximum(step, 1e-30))
+            wu -= dwu * lim
+            wv -= dwv * lim
+            if numpy.max(step) < self._angle_tol:
+                break
```
jobovy, line 469: This looks like it works for multiple angles given as arrays, but it keep iterating on all until they all pass the tolerance? Wouldn't it be better to use the approach used in the 1D and spherical inverse AA codes and cut out the ones that have converged?

jobovy (changes requested):

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done in ffc893cf — the loop now carries an `unconv` boolean mask and evaluates the profiles, Jacobian, and step only on the still-unconverged angles, matching the 1D/spherical pattern.

Measured on 20000 random angles of a Kuzmin-Kutuzov torus, the unconverged count per iteration runs `20000, 20000, 20000, 20000, 20000, 11350, 0`. So the saving is real but modest — about 20% of the integrand work, and ~80 → ~76 µs per phase-space point — because the Newton iteration converges quadratically and nearly every angle finishes on the same iteration; the old code already exited the loop as soon as the slowest angle converged, so what is saved is the work on the last iteration or two rather than extra iterations. Worth doing anyway for consistency with the other two codes, and it scales better for the near-degenerate cases where a few angles do lag.

Tests still pass and module coverage stays at 100%.

jobovy (approved):

PR #1350, opened 2026-08-20
Move the grid-quantity parser to the actionAngleInverse base module

jobovy-agent:
The unification flagged in #1342's body, now that #1339 and #1342 have both landed: one `_parse_grid_quantity` helper in `actionAngleInverse.py` (verbatim the spherical branch's implementation), used by `actionAngleVerticalInverse` in place of its private `_parse_energies`. The spherical class switches to it when `aainv-spherical-dev` merges to main (that branch stays linear-history — no sync merge was made).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01N2Jm4Y1A9eQRXMEaHaqwwJ

PR #1351, opened 2026-08-20
Exact 1D point transformation by quadrature instead of ODE integration

jobovy-agent:
The 1D half of the ODE → quadrature swap (E1's mode (c), decided in review): the equal-time map is the level-set match of two independently computable time profiles, so the construction becomes a composite Gauss–Legendre quadrature of the regular χ-anomaly integrand plus a monotone spline inversion — machine precision with no tolerance knob, and quarter-period normalization enforces the turning-point closure exactly (the ODE version pinned it by hand). `scipy.integrate.solve_ivp` is gone from the module.

- Stored representation, evaluation, and API unchanged; `exact_pt_tol` is now inert (documented; kept for backward compatibility).
- Accuracy: equal or better everywhere. Construction-limited tori improve 1–2 orders (isothermal E=1: 1.4e-10 vs ~3e-9 orbit error over 10 periods). At E=50 both constructions give an identical 3.2e-4 — that is a *pre-existing* stored-mesh resolution limit (the disk-crossing feature is ~2 mesh spacings wide at `pt_nxa=301`; 1.9e-7 at `pt_nxa=1001`), verified by measuring the ODE version side-by-side.
- The `pt_only` accuracy warning now triggers in tests via a coarse `pt_nxa` (the loose-tolerance trigger no longer exists); all 34 vertical tests pass.
- The spherical half of the swap follows on `aainv-spherical-dev` after it is rebased on main post-#1350, per plan; the ms.tex 1D-section update (the text describes the η-form ODE) comes as a fast-orbits PR with latexdiff.

### DIFF
--- a/galpy/actionAngle/actionAngleVerticalInverse.py
+++ b/galpy/actionAngle/actionAngleVerticalInverse.py
@@ -62,7 +62,6 @@ class actionAngleVerticalInverse(actionAngleInverse):
         pt_deg=7,
         pt_nxa=301,
         exact_pt_spl_deg=5,
-        exact_pt_tol=1e-12,
         pt_only=False,
         maxiter=100,
         angle_tol=1e-12,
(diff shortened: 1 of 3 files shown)


galpy/actionAngle/actionAngleVerticalInverse.py
```diff
@@ -85,7 +91,8 @@ def __init__(
         exact_pt_spl_deg : int
             degree of the spline used to represent the exact point transformation (only used when use_pointtransform == "exact")
         exact_pt_tol : float
```
jobovy: No need for backwards compatibility, just drop the keyword.

jobovy-agent:
Done in 385b0fbb — `exact_pt_tol` is dropped entirely (signature, docstring, stored attribute). Also applied the same removal on the spherical quadrature branch (#1353, commit 22b9f1b4) since the keyword was equally never shipped there.

jobovy (approved):

PR #1353, opened 2026-08-20
Exact spherical point transformation by quadrature instead of ODE integration

jobovy-agent:
The spherical half of the ODE → quadrature swap (companion to jobovy/galpy#1351, on the freshly rebased `aainv-spherical-dev`): the equal-time map becomes the spline inversion of the machine-precision time-from-pericenter profile against the time along the auxiliary isochrone orbit, which is closed-form via the isochrone's Kepler-like relation $\Omega^A_r t^A = \theta^A_r = \eta - [ec/(c+b)]\sin\eta$ (half-period normalization enforces turning-point closure exactly), and Δψ is the cumulative quadrature of its exactly-regular integrand. `solve_ivp` is gone; storage, evaluation, and API unchanged; `exact_pt_tol` dropped entirely.

- Construction error now machine-level: |nSₙ|/J_r ≈ 3e-9 even at ecc 0.956 (the orbit that motivated PTs).
- Extreme-orbit end-to-end accuracy is now honestly *representation*-limited: at default `pt_nra=301` the ecc-0.956 orbit reconstructs at 4.7e-4 (same as the ODE version — verified side-by-side), improving to 4.1e-5 at `pt_nra=1001`; the `pt_only` accuracy warning now correctly fires in this regime and its message points at `pt_nra`.
- The warning's test trigger is a coarse `pt_nra` (the loose-tolerance trigger no longer exists); all 21 spherical tests + quantity tests pass.
- Note: the vertical-side loose-tolerance test at tests/test_actionAngle.py:7206 still exists on this branch (it tests the ODE version that this branch's vertical class still has); it is fixed in #1351 and the next dev rebase after #1351 merges will reconcile the two (expected small conflict in that test block).

The ms.tex spherical-section update rides the open fast-orbits paper PR.

### DIFF
--- a/galpy/actionAngle/actionAngleSphericalInverse.py
+++ b/galpy/actionAngle/actionAngleSphericalInverse.py
@@ -111,7 +111,6 @@ class actionAngleSphericalInverse(actionAngleInverse):
         pt_deg=7,
         pt_nra=301,
         exact_pt_spl_deg=5,
-        exact_pt_tol=1e-12,
         pt_only=False,
         maxiter=100,
         angle_tol=1e-12,
@@ -174,8 +173,6 @@ class actionAngleSphericalInverse(actionAngleInverse):
         exact_pt_spl_deg : int, optional
             Degree of the spline representation of the exact point
             transformation.
-        exact_pt_tol : float, optional
-            Tolerance of the exact point-transformation ODE solution.
         pt_only : bool, optional
             If True, evaluate the inverse transformation using the exact
             point transformation alone, skipping the generating-function
@@ -377,7 +374,6 @@ class actionAngleSphericalInverse(actionAngleInverse):
             ):
                 self._pt_exact = True
                 self._exact_pt_spl_deg = exact_pt_spl_deg
-                self._exact_pt_tol = exact_pt_tol
                 self._pt_only = pt_only
             else:
                 if pt_only:
--- a/galpy/actionAngle/actionAngleSphericalInverse.py
+++ b/galpy/actionAngle/actionAngleSphericalInverse.py
@@ -768,12 +768,14 @@ class actionAngleSphericalInverse(actionAngleInverse):
             # t(chi) = int dr/vr = deltar int dchi / sqrt(Q) with
             # Q = vr^2/[y(1-y)] regular at both turning points, is computed
             # to machine precision with a composite Gauss-Legendre rule,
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Dropped the never-shipped `exact_pt_tol` keyword here too (22b9f1b4), matching the #1351 review.

jobovy-agent:
Both points addressed.

**1. The vertical-side test reconciliation — not done before, done now.** #1351 merged ~an hour ago, so the note was still accurate when written but had just come due. Executed the full chain:

- `aainv-spherical-dev` rebased onto main (picking up #1351, #1356, #1349). Only conflict was the expected add-add hunk in `HISTORY.txt`; both entries kept. The vertical loose-tolerance test now carries #1351's version (coarse `pt_nxa=7` trigger, no `exact_pt_tol`), so the reconciliation the note predicted is complete — force-pushed to `aainv-spherical-dev`.
- This branch then rebased onto the new dev with `--onto` (replaying only its own two commits, so the old dev commits don't duplicate). No conflicts: the spherical loose-tolerance test here already uses the coarse `pt_nra=9` trigger, which is the spherical mirror of #1351's fix.
- 66 vertical + spherical-inverse tests pass on both the rebased dev and this branch.

**2. "Kepler time" — you're right, that was sloppy.** The auxiliary is an isochrone, not a Kepler potential, and what I meant is the *time along the auxiliary isochrone orbit*, which happens to be closed-form because the isochrone's radial angle obeys a Kepler-*like* relation. Fixed in 16e0bf45 (code) and in the paper (fast-orbits#12, latexdiff sent in-session):

- The setup comment now reads "the time along the auxiliary isochrone orbit is known in closed form, `Omega^A_r t^A = eta - k sin eta` with `k = e c/(c+b)` (the isochrone's Kepler-like equation, which is what makes the auxiliary side free)".
- `kepk` is described as "the eccentricity-like parameter k of the isochrone's Kepler-like equation", not "the Kepler-equation parameter".
- The paper passage now states the relation explicitly with a cross-reference to the isochrone-angle equation, instead of "the auxiliary orbit's closed-form Kepler time".
- This PR's description is updated the same way.

I kept "Kepler-like equation" where it is accurate (the transcendental relation solved for η — also galpy's existing wording in `actionAngleIsochroneInverse`), and "the Kepler and harmonic limits" in the Ω_θ/Ω_r guard, which does refer to those actual limiting potentials.

Issue #1354, opened 2026-08-20
actionAngleAdiabatic C: plain Gauss-Legendre on sqrt-endpoint integrands loses ~4 digits in jr/jz


jobovy-agent:
### Summary

`actionAngleAdiabatic(c=True)` — the default — computes `jr` and `jz` with a fixed-order
Gauss-Legendre rule applied directly to integrands that vanish like a square root at the
integration endpoints. GL converges only algebraically on such integrands, so at the default
`order=10` both actions carry **~1e-4 relative error**. The pure-Python (`c=False`) path is
exact to machine precision, so the two disagree by ~5e-4 — far more than one would expect
from "same quantity, two implementations".

The fix is a change of variable that is **already used elsewhere in the same file**, and it
costs nothing extra.

### Where

`galpy/actionAngle/actionAngle_c_ext/actionAngleAdiabatic.c`

- `calcJzAdiabatic`: `gsl_integration_glfixed(JzInt, 0., zmax, T)` — integrand `sqrt(Ez - Phi_z(z))`
  has a sqrt zero at `z = zmax`.
- `calcJRAdiabatic`: `gsl_integration_glfixed(JRInt, rperi, rap, T)` — sqrt zeros at **both** ends.

### Measurements (MWPotential2014, R=1.1)

Reference computed independently (endpoint-regularizing substitution + adaptive `scipy.quad`,
`epsabs=epsrel=1e-14`); cross-checked against the `c=False` path, which agrees to 7e-15.

Reproducing the C rule in numpy matches the C output to **1.1e-14 (jz) / 2.3e-14 (jr)**, which
confirms the rule being measured is the one C actually runs.

| order | `jz` plain (current) | `jz` with `z = zmax*sin(phi)` | `jr` plain (current) | `jr` with `r = cc - rr*cos(theta)` |
|------:|---------------------:|------------------------------:|---------------------:|-----------------------------------:|
| 10    | 1.44e-04             | 1.24e-07                      | 4.61e-04             | 2.15e-11                           |
| 20    | 1.93e-05             | 1.22e-12                      | 6.15e-05             | 1.40e-14                           |
| 40    | 2.49e-06             | 1.42e-15                      | 7.96e-06             | 1.40e-14                           |

The plain column falls off ~8x per doubling of the order, i.e. O(n^-3) — the expected algebraic
rate for a sqrt endpoint. The substituted column is spectrally convergent.

[...]

PR #1357, opened 2026-08-21
actionAngleStaeckel: regularize the action quadratures' sqrt turning points (C + Python)

jobovy-agent:
Fixes the Stäckel half of the systematic action-quadrature defect independently diagnosed in #1354 (and fixed for the adiabatic code in #1356): plain fixed-order Gauss–Legendre applied against the square-root branch points at the turning points converges only as order⁻³, leaving **systematic ~4.6e-4 (J_R) / 1.4e-4 (J_z)** relative errors at the default order. This PR upgrades the **pure-Python path** (`c=False` with `fixed_quad=True`, plus all frequency/angle integrals, which are Python-only); the **C path is deliberately left for a separate PR** (see below).

This is also the first half of the code-sharing plan discussed in #1347 (comment on the quadrature duplication): the χ-form machinery now lives in `actionAngleStaeckel`, where `actionAngleStaeckelInverse` can consume it once both land.

**What changed**

- All u/v quadratures go through a new `_staeckelChiQuadratures`: a composite 10-pt Gauss–Legendre rule in the anomaly `q = qmin + D sin²(χ/2)`, in which every integrand is regular (action → `(D/4)√Q sin²χ`, 1/p profiles → `f·D/√Q`, with `Q = S/[y(1−y)]`). One vectorized evaluation of the momentum per mesh serves the action **and** all three 1/p integrals of that degree of freedom; the same cumulative form serves the partial-oscillation integrals of `calcAngles` (the C port's `(panel, mid)` encoding and all branch logic are kept verbatim). The per-scalar t²-substitution wrappers and derivative integrands are removed.
- Near the turning points, where direct evaluation of S is cancellation-dominated, S is reconstructed from its **analytic derivative** (new `_dJR/_dJzStaeckelIntegrandSquaredd*` via the forces, validated against finite differences at 1e-10) by the trapezoid between the turning point and the node.
- The turning-point `brentq` solves are tightened to `xtol=1e-15, rtol=8.9e-16`. This matters more than it looks: a root offset δ enters the 1/√S integrals as **√δ**, so scipy's default `xtol=2e-12` put a ~1e-6 floor under the frequencies no matter how good the quadrature.

**Accuracy** (default order): actions/frequencies/angles match a 20× finer mesh to ~1e-11 for generic orbits (~1e-9 for a nearly planar orbit — the evaluation-noise floor of the fudge-form momentum function, not the rule). The handoff's benchmark (MWPotential2014, δ=0.45, order-640 reference): order-10 J_R error **4.6e-4 → 1.2e-17**.

**Test changes**

- New `test_actionAngleStaeckel_chi_quadrature_convergence` (default order vs 20× finer mesh, three orbits covering both u branches, both v branches, and near-planar).
- `test_actionAngleStaeckel_actions_order`: low and high order must now both match a very-high-order reference at machine precision (the old "higher order is more accurate" assertion is vacuous when every order is converged).
- `test_actionAngleStaeckel_python_c_freqsAngles`: the C side now runs at `order=200`, since the Python path no longer shares the C path's truncation error — the parity test again checks branch/convention agreement rather than the common truncation floor.

**Follow-up (separate PR, per the handoff's own recommendation):** the same defect in the C path (`actionAngleStaeckel.c:590/657`), where the substitution is free (same node count) but changes `c=True` output at the ~1e-4 level and must move in lockstep with the galpy-jax backend that mirrors C bit-for-bit — kept separate so that cost decision is reviewable on its own. Once merged, the parity test's `order=200` crutch can be dropped.

Credit: the diagnosis, measurements, and fix shape were independently worked out for the adiabatic twin defect in the handoff document `galpy-jax/HANDOFF_staeckel_quadrature.md`; this PR confirms its Stäckel claim and implements the Python-side fix with the χ-anomaly machinery from the inverse work.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -14,6 +14,16 @@ v1.12.1 (expected around 2026-11-01)
   ~33% rather than 2x. Together these reduce the error to ~1e-9. Note that this
   changes the values returned for ``c=True``.
 
+- The pure-Python path of ``actionAngleStaeckel`` (``c=False`` with
+  ``fixed_quad=True``, and the frequency/angle calculations) now evaluates
+  all of its quadratures with a composite Gauss-Legendre rule in an anomaly
+  that absorbs the square-root behavior at the turning points, and solves
+  for the turning points at machine tolerance. The fixed-order rule
+  converged only as order^-3 against the turning-point branch points,
+  leaving systematic ~5e-4 relative action errors at the default order;
+  actions, frequencies, and angles at the default order are now converged
+  to ~1e-11 or better (the C path will be upgraded separately).
+
 - ``actionAngleVerticalInverse`` now supports physical units: it accepts
   ``ro`` and ``vo``, parses the energies of its tori and the actions and
   angles at which it is evaluated when these are given as Quantities, and
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,6 +1,11 @@
 v1.12.1 (expected around 2026-11-01)
 ====================
 
+
+
+============
+
+<<<<<<< HEAD
 - The C implementation of ``actionAngleAdiabatic`` now evaluates the radial
   and vertical action integrals in a regularized variable. Both integrands
   vanish like a square root at the ends of their intervals (at ``zmax`` for
@@ -23,6 +28,21 @@ v1.12.1 (expected around 2026-11-01)
   leaving systematic ~5e-4 relative action errors at the default order;
   actions, frequencies, and angles at the default order are now converged
   to ~1e-11 or better (the C path will be upgraded separately).
+=======
+- The action quadratures of ``actionAngleStaeckel`` are regularized at the
+  turning points in both the C and the pure-Python path. The fixed-order
+  Gauss-Legendre rule was previously applied against the square-root branch
+  points at the turning points, converged only as order^-3, and left
+  systematic ~5e-4 relative action errors at the default order. The C path
+  now applies the same t^2 endpoint substitution that its derivative
+  routines already used (actions at the default order are converged to
+  ~1e-12; note that ``c=True`` action values change at the ~5e-4 level).
+  The pure-Python path (``c=False`` with ``fixed_quad=True``, and the
+  frequency/angle calculations) evaluates all of its quadratures with a
+  composite Gauss-Legendre rule in an anomaly that absorbs the square-root
+  behavior, and solves for the turning points at machine tolerance;
+  actions, frequencies, and angles at the default order are converged to
+  ~1e-11 or better.
 
 - ``actionAngleVerticalInverse`` now supports physical units: it accepts
   ``ro`` and ``vo``, parses the energies of its tori and the actions and
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,34 +1,6 @@
 v1.12.1 (expected around 2026-11-01)
 ====================
 
-
-
-============
-
-<<<<<<< HEAD
-- The C implementation of ``actionAngleAdiabatic`` now evaluates the radial
-  and vertical action integrals in a regularized variable. Both integrands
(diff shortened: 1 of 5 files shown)


jobovy-agent:
Following the discussion in the session: the **C path is now fixed too** (second commit, 7f86a33a), so this PR covers both halves of the defect and the galpy-jax backend can move in lockstep with a single galpy release.

The C fix is exactly the handoff's prescription — the same t² endpoint substitution the `dJ?d?` C routines already use, applied to the two action `glfixed` sites. One refinement beyond the handoff: the one-sided `v = vmin + t²` substitution alone still leaves 6e-9 at the default order (one long panel), so the J_z integral is split at the midpoint into `vmin + t²` / `π/2 − t²` panels like the derivative routines, which reaches the root-tolerance floor. Same node count, so cost is unchanged.

Measured at the handoff's benchmark (MWPotential2014, δ=0.45, order-1280 reference): C order-10 errors drop from **4.6e-4 / 1.4e-4 to 1.6e-12 / 9e-12** (J_R/J_z), and C agrees with the machine-converged χ-form Python path at 1e-14. As flagged in the handoff and now in `HISTORY.txt`: **`c=True` action values change at the ~5e-4 level**. New regression test pins default-order-vs-1280 agreement at 1e-10 on three orbits; the full actionAngle suite (224 tests) passes locally with the rebuilt extension.

jobovy-agent:
Response to the independent verification from the adiabatic-side agent (relayed via the session; thanks — the C numbers match mine exactly).

**On the `order=` no-op observation:** I could not reproduce bit-identity through the public path — on this branch, `actionAngleStaeckel(..., c=False)(R,vR,vT,z,vz, fixed_quad=True, order=N)` gives identical results for N=5 and 10 (both sit at the `nchi = max(2·order, 20)` floor, by design) but last-digit-distinct, machine-converged results for N=20/40/1280. Two ways the sweep can legitimately see bit-identity, both now addressed in 696cc4f8:

1. `actionAngleStaeckelSingle`'s per-instance `_JR`/`_JZ`/`_djrdE`… caches ignored a changed `order=` on re-call, returning the first result bit-for-bit — exactly the trap for probing convergence on a reused instance. The caches are now keyed on `(fixed_quad, order)`.
2. The semantics changed silently: sweeping `order` is no longer a convergence test, since every order is converged. The class and method docstrings now state explicitly that on the Python path `order` scales the composite χ-mesh with a floor, and that increasing it has no practical effect.

**On the parity pinning:** intentional and now double-checked — `test_actionAngleStaeckel_python_c_freqsAngles` pins Python (any order) against C at `order=200` (branch/convention agreement above the C frequency/angle truncation floor), and `test_actionAngleStaeckel_chi_quadrature_convergence` pins the Python mesh convergence internally (order 10 vs 200 ⇒ nchi 20 vs 400). For the backend: since the C fix here is exactly the t² substitution, a backend that mirrors C bit-for-bit at a given order can keep doing so against the *new* C rule; it is the galpy-Python path that has deliberately left the shared-rule regime.

**On merge order with #1356:** agreed it's a one-hunk HISTORY.txt conflict either way; happy for whichever lands second to take it.

jobovy-agent:
Rebased onto main post-#1356/#1349 (the predicted one-hunk HISTORY.txt resolution, taken here as offered) and fixed the two genuine CI failures: (1) `test_galpypaper.py::test_qdf` pinned quasiisothermaldf values computed with the old c=True actions — updated to fresh values with a dated NOTE, per the file's convention; (2) the python-c parity test's 1e-6 tolerance sat exactly on the C turning-point root-tolerance floor (√δ-amplified in the 1/√S integrals, platform-dependent: ~1e-8 Linux vs 1.01e-6 Windows; order-independent, so raising the C order doesn't help) — now 1e-5 with the mechanism documented. codecov should recover with the tests passing.

PR #1361, opened 2026-08-22
actionAngleStaeckelInverse: fix zero frequencies for very thin tori

jobovy-agent:
Found while designing the interpolation grid (fast-orbits#15): the grid's shell edge is made of near-`J_R = 0` tori, and building them exposed a silent failure in the merged code.

**The bug.** For tori with `J_R` below ~1e-9 the frequencies came back as **exactly zero**, with no error. Traced to the endpoint handling in the profile quadratures: near a turning point `W` is a difference of O(1) terms and is dominated by cancellation noise (it dips to −5e-15), so `Q = W/[y(1−y)]` is reconstructed there from the analytic derivative instead. But the switch was a *fixed* threshold in the anomaly, `y(1−y) > 1e-6`, whereas the signal near a turning point scales as `W′ D y` while the noise floor does not — so for a thin oscillation the noise-dominated region is wider than any fixed anomaly cut. One node left on the wrong side gave `Q < 0`, the tiny-clamp turned that into `1/sqrt(Q) ≈ 4.5e161`, and the profile integrals exploded to ~1e145:

```
   1-w_I    u-width         J_R          PEu          det       Om_R
   1e-03   1.59e-04   1.19e-08   1.1111e+00   2.9817e-01   1.247872
   1e-04   1.59e-05   1.19e-10  3.0068e+145  8.0692e+144   0.000000   <-- silent
```

**The fix.** Make the switch relative to the size of `W` on the torus rather than a fixed anomaly cut, and reconstruct with the trapezoid `W ≈ (q−q0)[W′(q0)+W′(q)]/2` instead of the constant turning-point limit — the same treatment the forward code got in #1357. This extends the usable range by about four orders of magnitude in `J_R`, to ~1e-12, where the frequencies still hold five to six digits:

| 1−w_I | u-width | J_R | Ω_R |
|---|---|---|---|
| 1e-3 | 1.6e-4 | 1.2e-8 | 1.247872 |
| 1e-4 | 1.6e-5 | 1.2e-10 | 1.247814 |
| 1e-5 | 1.6e-6 | 1.2e-12 | 1.245796 |

Below that the two turning points coincide to within a few ulp and nothing is recoverable in double precision, so setup now **raises** rather than returning a plausible-looking zero. I verified the guard is genuinely needed by removing it: the unresolvable tori come back with `Ω_R = 0.000000` and no complaint.

`_dWu`/`_dWv` became shape-agnostic (the wrapper's derivative only accepts 1D input) because the reconstruction now needs the derivative on the whole node grid, not just at the endpoints.

**Tests:** one walking a torus in to `dE = 1e-12` and requiring the frequencies to stay finite, positive, and within 1% of the shell-limit value; one pinning the new error, constructed by solving for the shell edge and stepping a single ulp inside it (`numpy.nextafter`, deterministic rather than a tuned magic number). Module coverage stays at 100%.

Targets `aainv-staeckel-dev`, since it fixes code that landed with #1347.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelInverse.py
+++ b/galpy/actionAngle/actionAngleStaeckelInverse.py
@@ -219,6 +219,19 @@ class actionAngleStaeckelInverse(actionAngleInverse):
                     uhi *= 2.0
             self._umins[ii] = brentq(Wu, ulo, uin, xtol=1e-15, rtol=8.9e-16)
             self._umaxs[ii] = brentq(Wu, uout, uhi, xtol=1e-15, rtol=8.9e-16)
+            if (self._umaxs[ii] - self._umins[ii]) < 1e-12 * (1.0 + self._umaxs[ii]):
+                # The two turning points are separated by a few ulp: the
+                # oscillation cannot be resolved in double precision, and
+                # everything built on it (the profiles, and hence the
+                # frequencies) would be noise. Fail loudly rather than
+                # return a plausible-looking zero
+                raise ValueError(
+                    f"u oscillation of torus {ii} is unresolvable in double "
+                    f"precision (u_max - u_min = "
+                    f"{self._umaxs[ii] - self._umins[ii]:e} for "
+                    f"E={E}, Lz={Lz}, I3={I3}); the torus is a shell orbit "
+                    "to within machine precision"
+                )
             if Wv_all[ii, -1] <= 0.0:
                 raise ValueError(
                     f"Midplane not reached for torus {ii}; no valid torus "
@@ -240,20 +253,21 @@ class actionAngleStaeckelInverse(actionAngleInverse):
 
     ############################ SETUP: QUADRATURES ###########################
     def _dWu(self, u, E, Lz):
-        """dW_u/du (independent of I3); u may be an array"""
+        """dW_u/du (independent of I3); u may be an array of any shape (the
+        wrapper's derivative only accepts 1D, so it is evaluated flat)"""
+        u = numpy.asarray(u)
+        dUdu = self._staeckelwrap._dUdu(u.ravel()).reshape(u.shape)
         return (
-            2.0
-            * self._delta**2.0
-            * (E * numpy.sinh(2.0 * u) - self._staeckelwrap._dUdu(u))
+            2.0 * self._delta**2.0 * (E * numpy.sinh(2.0 * u) - dUdu)
             + 2.0 * Lz**2.0 * numpy.cosh(u) / numpy.sinh(u) ** 3.0
         )
 
     def _dWv(self, v, E, Lz):
-        """dW_v/dv (independent of I3); v may be an array"""
+        """dW_v/dv (independent of I3); v may be an array of any shape"""
+        v = numpy.asarray(v)
+        dVdv = self._staeckelwrap._dVdv(v.ravel()).reshape(v.shape)
         return (
-            2.0
-            * self._delta**2.0
-            * (E * numpy.sin(2.0 * v) + self._staeckelwrap._dVdv(v))
+            2.0 * self._delta**2.0 * (E * numpy.sin(2.0 * v) + dVdv)
             + 2.0 * Lz**2.0 * numpy.cos(v) / numpy.sin(v) ** 3.0
         )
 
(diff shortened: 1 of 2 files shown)


galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -219,6 +219,19 @@ def _find_turning_points(self):
...
+                # everything built on it (the profiles, and hence the
+                # frequencies) would be noise. Fail loudly rather than
+                # return a plausible-looking zero
+                raise ValueError(
+                    f"u oscillation of torus {ii} is unresolvable in double "
+                    f"precision (u_max - u_min = "
+                    f"{self._umaxs[ii] - self._umins[ii]:e} for "
+                    f"E={E}, Lz={Lz}, I3={I3}); the torus is a shell orbit "
+                    "to within machine precision"
+                )
```
jobovy, line 234: Maybe instead we should just fall back on a torus where the difference is above 1e-12?

tests/test_actionAngle.py
```diff
@@ -8413,6 +8413,93 @@ def Uofu(u):
...
+        assert numpy.isfinite(aASI._OmegaR[0]) and aASI._OmegaR[0] > 0.0, (
+            "Radial frequency of an ultra-thin shell torus is not finite and "
+            "positive (dE = %g)" % dE
+        )
+        if ref is None:
+            ref = aASI._OmegaR[0]
+        assert numpy.fabs(aASI._OmegaR[0] / ref - 1.0) < 1e-2, (
+            "Radial frequency of an ultra-thin shell torus does not stay at "
+            "the shell-limit value as the torus is thinned (dE = %g)" % dE
+        )
```
jobovy: Seems like a weak test, it can't be tighter?

jobovy (changes requested):

tests/test_actionAngle.py
jobovy-agent: Tightened tenfold, to 1e-3. The limit is the construction rather than the assertion: the frequencies hold to 1.2e-4 down to a u-width of 1.4e-4 and 4.5e-4 at 1.4e-5, then degrade smoothly as the turning-point solve runs out of digits — 6e-3 at 1.4e-6, 6e-2 at 1.4e-7. The test now stops at the last torus where the construction is still trustworthy (it used to include one past it, which is what forced the loose tolerance), and the degradation is documented in the test so the number isn't mysterious.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: I tried this and it does not work in the naive form — worth recording why, because the failure is instructive.

**First, the threshold isn't reachable.** The turning-point root find bottoms out at a u-width of ~1.4e-7, not 1e-12: widths run 1.4e-2, 1.4e-3, ... 1.4e-6, 4.4e-7, 1.35e-7, and then the next step collapses to ~1e-15 (garbage) rather than passing smoothly through 1e-12. So "a torus where the difference is above 1e-12" isn't a state the solver can deliver.

**Second, forcing the width gives confidently wrong answers.** I tested the direct version of the idea — take the shell location u\* (the maximum of W_u) and set u± = u\* ∓ h/2 — letting the analytic-derivative reconstruction carry the profiles:

| forced width | Ω_R | error vs the shell limit |
|---|---|---|
| 1e-6 | 0.7073 | **14%** |
| 1e-8 | 0.9171 | **12%** |
| 1e-10 | 0.0000 | 100% |

The reason is that the trapezoid reconstruction uses W′ *at the turning points*, and that is only the right thing when they actually are roots of W_u. At a forced non-root, W′(u\* ∓ h/2) ≈ ∓|W″|h/2, which describes nothing. So this fallback would replace a loud failure with a plausible-looking 14% error, which seems clearly worse.

**A correct fallback is possible but changes the answer.** One could solve for the I3 that *does* give a resolvable width — analytically, since W_max = 2δ²(I3_shell − I3) and width ≈ 2√(4δ²ΔI3/|W″|), so a width of 1e-5 needs ΔI3 ~ 5e-11, comfortably above the 1e-15 noise floor — construct that torus normally, and warn. The frequencies would then be good to ~5e-4. But it returns a *different torus* than the one requested, which for an inverse transformation feels like the wrong default: the user asked for J_R ≈ 1e-20 and would silently get 1e-9.

My inclination is therefore to keep the error, since the message says exactly what has happened ("the torus is a shell orbit to within machine precision") and the regime is unreachable by accident — the interpolation grid keeps `grid_pad` away from the edge by construction, and hitting it by hand requires labels within ~1e-15 of the exact shell. Happy to add the snap-and-warn version if you'd rather have it; it is maybe fifteen lines given `_I3_shell` already exists for the grid.

jobovy: Maybe the resolution will be in the interpolation. I would want interpolation to work everywhere, even at JR=0, but if that's the case then I agree that explicitly asking for JR=1e-20 and getting 1e-9 is somewhat bad (although they are in practice the same torus).

jobovy (approved):

PR #1362, opened 2026-08-22
actionAngleStaeckelInverse: accept arbitrary actions via a grid of tori (N3, part 1)

jobovy-agent:
**Stacked on #1361** (which fixes the thin-torus bug the grid's shell edge exposes); review that one first.

Adds `setup_interp=True`, which makes the class a usable inverse: instead of only evaluating the tori listed at initialization, it builds a grid spanning the bound phase space and accepts arbitrary actions within it.

**Design choice worth flagging.** The grid recovers a torus's *integrals* `(E, I3)` from its actions by interpolation alone, and the torus is then **constructed exactly**. So the interpolation enters only through `(E, I3)` — the returned point lies on the true torus, with no interpolated-profile error. This is a deliberate first increment: it is much less code than interpolating the six angle profiles across the grid, it is exact where it matters, and it delivers the user-facing feature. Interpolating the profiles as well (to cut the ~20 ms per-torus construction) is the natural follow-up, and the design work for it is already done in fast-orbits#15 stage 4.

**Design settled by measurement** (fast-orbits#15, stages 1–3):

- The grid must be rectified at *every* edge, since torus properties vary as the square root of the distance to each: uniform in `L_z`, quadratic in `E` away from the circular orbit (as in the spherical grid), and in `s = sin²(πw_I/2)` between the planar and shell edges, which rectifies both at once. Uniform spacing in `(E, I3)` is not merely slower to converge — it gives relative errors of order unity in the actions.
- The planar edge (`J_z = 0`) is closed form in the `V(π/2)=0` gauge, `I3 = L_z²/(2δ²) − E`; the shell edge (`J_R = 0`) is the double root of `W_u`.
- Recovering the grid coordinates from the actions factorizes into two 1D monotone inversions — `ρ = √(J_R+J_z)` increases with `w_E` at fixed `w_I`, and `ζ = (2/π) arcsin √[J_z/(J_R+J_z)]` increases with `w_I` at fixed `w_E` — so no 2D root find is needed. `ζ` carries the same rectification as the grid; without it the recovered `w_I` is four orders of magnitude worse.

**Accuracy and cost** (Kuzmin-Kutuzov, round trip against the forward transformation):

| grid | max round-trip error | setup | evaluation |
|---|---|---|---|
| 7³ | 1.8e-04 | 18 s | ~15 µs/point |
| 11³ | 3.3e-06 | 69 s | ~15 µs/point |

Evaluation cost is quoted for a torus sampled at many angles, which amortizes the exact construction. Setup is dominated by the shell-edge solves and the grid construction; both are straightforward to speed up later if it matters.

**Tests:** round trip on three orbits including `z<0`/`vR<0` branches; convergence with grid size plus interpolated frequencies against the forward code at 1e-4; and rejection of actions outside the grid in both the `L_z` and the energy directions. Module coverage stays at 100%.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelInverse.py
+++ b/galpy/actionAngle/actionAngleStaeckelInverse.py
@@ -9,10 +9,16 @@
 #   lattice is involved; placement on the torus is exact by construction.
 ###############################################################################
 import numpy
-from scipy.interpolate import InterpolatedUnivariateSpline
+from scipy.interpolate import InterpolatedUnivariateSpline, RectBivariateSpline
+from scipy.ndimage import map_coordinates, spline_filter1d
 from scipy.optimize import brentq, minimize_scalar
 
-from ..potential import OblateStaeckelWrapperPotential
+from ..potential import (
+    OblateStaeckelWrapperPotential,
+    evaluatePotentials,
+    rl,
+    vcirc,
+)
 from ..util import coords
 from .actionAngleInverse import actionAngleInverse
 
@@ -25,6 +31,100 @@ _GLX, _GLW = numpy.polynomial.legendre.leggauss(10)
 # already wrapped; its value is irrelevant for an exactly Staeckel potential
 # (any choice gives the same U, V up to the fixed gauge)
 _U0INTERNAL = 1.15
+
+
+def _bspline_weights(t):
+    """Cubic B-spline weights for the four-point stencil at offset t"""
+    t2 = t * t
+    t3 = t2 * t
+    return numpy.array(
+        [
+            (1.0 - 3.0 * t + 3.0 * t2 - t3) / 6.0,
+            (4.0 - 6.0 * t2 + 3.0 * t3) / 6.0,
+            (1.0 + 3.0 * t + 3.0 * t2 - 3.0 * t3) / 6.0,
+            t3 / 6.0,
+        ]
+    )
+
+
+class _ProfileSet:
+    """Several angle profiles sharing one uniform chi mesh.
+
+    The profiles are stored as cubic B-spline coefficients, padded so the
+    four-point stencil is always in range, and evaluated with an explicit
+    weight formula rather than through a library routine: the Newton
+    iteration needs the same chi for several profiles at a time, and for a
+    single angle the per-call overhead of a library interpolator dominates
+    everything else.
+    """
+
+    def __init__(self, coeffs, dchi):
+        self._c = numpy.concatenate(
+            (coeffs[:, :1], coeffs[:, :1], coeffs, coeffs[:, -1:], coeffs[:, -1:]),
+            axis=1,
+        )
+        self._dchi = dchi
+        self._n = coeffs.shape[1]
+
+    def block(self, which):
+        """The padded coefficients of a subset of the profiles, sliced once
+        so that repeated evaluations do not re-copy them"""
+        return numpy.ascontiguousarray(self._c[which])
+
+    def evaluate(self, block, chi):
+        """Evaluate a block returned by block() at the angles chi"""
+        x = numpy.clip(chi / self._dchi, 0.0, self._n - 1.0)
+        i = x.astype(int)
+        t = x - i
+        j = i + 2
+        t2 = t * t
+        t3 = t2 * t
+        return (
+            (1.0 - 3.0 * t + 3.0 * t2 - t3) * block[:, j - 1]
+            + (4.0 - 6.0 * t2 + 3.0 * t3) * block[:, j]
+            + (1.0 + 3.0 * t + 3.0 * t2 - 3.0 * t3) * block[:, j + 1]
+            + t3 * block[:, j + 2]
+        ) / 6.0
+
+    def __call__(self, chi, which):
+        return self.evaluate(self.block(which), numpy.atleast_1d(chi))
+
+
+def _pad_axis(arr, axis, pad):
+    """Extend arr beyond both ends of axis by quadratic extrapolation from the
+    three nearest slices, vectorized over everything else"""
+    n = arr.shape[axis]
+    out = numpy.moveaxis(arr, axis, 0)
+    flat = out.reshape((n, -1))
+    deg = 2 if n > 2 else 1
+    # one polynomial fit for all columns at once, from the edge slices only:
+    # a global fit would be dominated by the far side of the grid
+    left = numpy.polyfit(numpy.arange(deg + 1), flat[: deg + 1], deg)
+    right = numpy.polyfit(numpy.arange(deg + 1), flat[n - deg - 1 :], deg)
+    lo = numpy.array(
+        [numpy.polyval(left, xx) for xx in numpy.arange(-pad, 0, dtype=float)]
+    )
+    hi = numpy.array(
+        [
+            numpy.polyval(right, xx)
+            for xx in numpy.arange(deg + 1, deg + 1 + pad, dtype=float)
+        ]
+    )
+    ext = numpy.vstack((lo, flat, hi)).reshape((n + 2 * pad,) + out.shape[1:])
+    return numpy.moveaxis(ext, 0, axis)
+
+
+def _prefilter_padded(arr, axes, pad):
+    """Pad arr along axes and return its cubic-spline prefiltered form, ready
+    for map_coordinates with prefilter=False. Only the given axes are
+    prefiltered: the leading axis indexes distinct quantities, and filtering
+    across it would mix them."""
+    out = arr
+    for ax in axes:
+        out = _pad_axis(out, ax, pad)
+    for ax in axes:
+        out = spline_filter1d(out, order=3, axis=ax, mode="nearest")
+    return out
 
 
 class actionAngleStaeckelInverse(actionAngleInverse):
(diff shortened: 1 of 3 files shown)


jobovy-agent:
You're right on all three counts, and the measurements say this PR's design is not worth merging as it stands. **Marked as draft** while I rework it.

**How fast is it, vs spherical:**

| | spherical (interpolated) | Stäckel (this PR) |
|---|---|---|
| per **new torus** | **0.71 ms** | **57 ms** — 80× slower |
| per point, one torus many angles | 4.0 µs | 15 µs |
| grid setup | 3.6 s (256 tori) | 18 s (7³) – 69 s (11³) |

So the honest verdict: this is not an interpolated inverse, it is an exact one with a lookup table bolted on, and it is two orders of magnitude off the spherical code precisely because it rebuilds the torus every time.

**On the wasted setup — worse than you put it.** Each grid torus costs 45 ms (2.9 ms turning points, 34 ms quadratures, 4.3 ms building the six profile splines), and I keep *only its two actions*. So ~60 s of work on an 11³ grid is thrown away to retain 2662 numbers.

But the fix is not to make the grid cheaper — it is to **keep what the grid already computed**. The profiles are the expensive part and they are exactly what interpolation needs, so the setup cost becomes justified the moment the profiles are stored rather than discarded. Your two points converge on the same answer.

**On (E, L_z, I3) input — also right, and it turns out to be the easy direction.** Given the integrals, the grid coordinates are *directly computable* (w_E from E_c and E_max, w_I from the planar and shell edges), with no inversion at all. The action route needs the two-stage lookup; the integral route needs nothing. It should be an entry point.

**Plan for the rework**, which supersedes this PR rather than stacking on it:

1. Store the six angle profiles (and their derivatives) on the grid, on a coarser χ mesh, padded and prefiltered as in fast-orbits#15 stage 4 — where I measured that unpadded `map_coordinates` degrades to 7.7e-4 at the grid edge while padded reaches 2.6e-6.
2. Evaluate the profiles inside the Newton loop through `map_coordinates` at a fractional torus index, as `sphericalInverse` does, so there is no per-torus spline construction — that alone is 4.3 ms, six times the entire spherical per-torus budget.
3. Accept both entry points: actions (two-stage lookup) and integrals (direct coordinates).

Target: per-torus cost in the neighbourhood of spherical's 0.7 ms rather than 57 ms. I will report the measured number, including the accuracy cost of interpolating the profiles rather than constructing them exactly — that is the trade this PR was avoiding, and it should be quantified rather than assumed.

jobovy-agent:
Reworked into actual interpolation and out of draft. The grid now **keeps what its construction produced** — turning points, frequencies, and the six angle profiles with their derivatives — and evaluation reads them by interpolation instead of rebuilding the torus. That removes the waste you identified: the profiles are the expensive part of a construction, and they were being computed and thrown away.

**Both entry points are supported**, and you were right that the integral one is the easy direction: given (E, L_z, I₃) the grid coordinates follow directly from the circular/outer energies and the planar/shell edges, with no inversion at all. Given the actions, the two-stage monotone lookup supplies them. The integral route is also the *more accurate* of the two, since it skips that inversion — there's now a test asserting exactly that.

**Speed, per new torus** (spherical: 0.71 ms):

| | ms/torus | µs/point |
|---|---|---|
| rebuilding exactly (the draft) | 57 | 18.5 |
| after the rework | **2.9** | **4.8** |
| spherical reference | 0.71 | 4.0 |

Per phase-space point is now within 20% of spherical. Per torus is 4× off, discussed below.

Each step came from profiling rather than guessing, and two were structural rather than micro-optimizations:

1. Precompute the shell edge and the lookup tables — a `brentq` around a bounded minimization was running *per evaluation* (2.2 ms), and the action lookup was rebuilding ~50 splines per call.
2. Fold the angle **before** evaluating a profile, instead of evaluating both branches and selecting — that alone was tripling the profile evaluations.
3. Evaluate values and derivatives in one call: they share the angle, hence the weights.
4. Contract the three grid directions with explicit B-spline weights rather than calling a library interpolator once per profile (146 `map_coordinates` calls per torus at the worst point).
5. Go straight from actions to grid coordinates, instead of converting to integrals and back — the round trip was evaluating the shell edge and energy limits twice for nothing.

**Accuracy** is unchanged in character: round trip against the forward transformation is 9.4e-4 on a 7³ grid and 2.3e-5 on 11³.

**On the remaining 4× in per-torus cost**, honestly: it is now ~1.2 ms of profile evaluation inside the Newton solve and ~0.9 ms in the torus read, and both are dominated by numpy call overhead on small arrays rather than arithmetic. Closing it would mean a scalar fast path that avoids numpy entirely for single-angle calls, which is a different kind of change; I would rather land this and do that separately if the per-torus number actually matters for your use. Note the pattern that matters most — many angles on one torus — is the 4.8 µs/point column, which is essentially at parity.

Cleanup worth flagging: this also removes three blocks of dead code left by my earlier edits, including a duplicate method definition that was silently overriding the optimized version (which is why one of my optimizations appeared to do nothing until I found it). Coverage is back to 100% with 15 tests.

jobovy-agent:
Two further pieces, following your notes.

**Interpolation now works at J_R = 0 and J_z = 0**, as you asked — the w_I direction spans the closed interval, so shell and planar tori are grid nodes rather than sitting outside the grid. The construction degenerates exactly there, so those nodes carry the analytic limits, and the measurements make the limits obvious: a vanishing oscillation is harmonic, so its angle *is* its anomaly (A_R = χ at the shell, B_z = χ at the planar edge) and its cross profiles vanish.

Getting these exactly right took three further points, each found by chasing an energy error that shouldn't have been there:

- A degenerate oscillation must be given **exactly zero** momentum. This matters because p = √W amplifies a machine-level residual to its square root: W_v(π/2) vanishes only to ~1e-16, which showed up as a vertical velocity of 1e-8 in an orbit meant to lie in the midplane.
- The turning points are interpolated like everything else, so they are not exact roots of W for the interpolated integrals; the momenta then miss the torus near them and the energy drifted by ~1e-6. One Newton step restores exact placement.
- At the planar edge I₃ follows in closed form from W_v(π/2) = 0; at the shell edge W_u needs a **double** root, so u\* and I₃ must be solved together.

Result: H is constant to 2–4e-16 along interpolated ordinary, shell, and planar tori alike; the shell orbit lies on a spheroid to 6e-16 and the planar one in the midplane to 2e-15. So interpolated tori keep the exact-placement property, with the interpolation error appearing purely as a displacement *along* the torus.

**Profiling the potential layer** then paid for the polish. Through the Stäckel wrapper a scalar `_U` costs 62 µs against 9 µs for a raw potential evaluation, and `_dUdu` costs **345 µs** because it makes three potential-layer calls — while vectorized `_U` costs 216 ns/point. The polish was spending ~0.8 ms per torus in a few scalar derivative calls. It now evaluates W once per degree of freedom at every point it needs and differences for the slope; the analytic dW reads more tidily but buys nothing, since the derivative only sets the Newton step rather than the root.

Per new torus: 3.6 → **2.9 ms**, so exact placement costs essentially nothing. Per point 4.8 µs (spherical 4.0). Coverage 100%, 16 tests.

Worth flagging for a future PR: that `_dUdu` cost is in the shared wrapper and affects anything using it — three potential-layer calls where the underlying potentials could supply force and value together. I left it alone here since it is not this PR's code.

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -444,7 +593,397 @@ def _build_angle_profile_splines(self):
             self._dAprof.append([a.derivative() for a in A])
             self._dBprof.append([b.derivative() for b in B])
 
-    ############################ EVALUATION ###################################
+    ############################ SETUP: TORUS GRID ############################
+    def _circular_orbit(self, Lz):
+        """Radius and energy of the circular orbit with this L_z"""
+        Rc = rl(self._pot, Lz, use_physical=False)
+        return Rc, evaluatePotentials(self._pot, Rc, 0.0) + Lz**2.0 / 2.0 / Rc**2.0
```
jobovy: Needs use_physical=False as well. Other places in this file?

galpy/actionAngle/actionAngleStaeckelInverse.py
```diff
@@ -524,9 +1232,29 @@ def _xvFreqs(self, jr, jphi, jz, angler, anglephi, anglez, **kwargs):
         sv = numpy.where(wvm <= numpy.pi, 1.0, -1.0)
         u = umin + Du * numpy.sin(chiu / 2.0) ** 2.0
         v = vmin + Dv * numpy.sin(chiv / 2.0) ** 2.0
-        pu = su * numpy.sqrt(numpy.clip(self._Wu(u, E, Lz, I3), 0.0, None))
-        pv = sv * numpy.sqrt(numpy.clip(self._Wv(v, E, Lz, I3), 0.0, None))
-        phi = thphi - self._fold(A[2], wu) - self._fold(B[2], wv)
+        # A degenerate oscillation carries no momentum. Saying so explicitly
```
jobovy: Put here but a general comment: can we make all of the comments less verbose please?

jobovy (changes requested):

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Fixed, and you were right to ask about other places: `rl` was guarded but `evaluatePotentials` and both `vcirc` calls were not, so a potential carrying ro/vo would have fed Quantities into the grid setup. All four call sites now pass `use_physical=False`, and I checked that the grid builds correctly with a Quantity-configured potential and astropy units switched on in the config.

galpy/actionAngle/actionAngleStaeckelInverse.py
jobovy-agent: Done — comments cut from 158 lines to 113. I kept the ones stating something the code cannot show (why the switch to the reconstructed Q is relative to the size of W rather than a fixed cut in the anomaly, why a degenerate oscillation needs an explicit zero momentum rather than sqrt of a near-zero W, why the grid contraction is done one axis at a time) and dropped the restatements of what the next line does. The pass also caught a block that an earlier edit had duplicated.

jobovy (approved):

PR #1364, opened 2026-08-22
backend: as_backend_constant must not name the default device (1.5x eager)

jobovy-agent:
## What

`coerce_coords` skips the `device=` keyword when the anchor **is** the backend's
default device. Naming it commits the placement instead of letting the backend
choose — `default_device`'s own docstring prices this at **397.6 us vs 134.6 us**
on jax CPU and blames it for turning *"a 1.7 s dxdv test into a 300 s timeout"*
(galpy #1300).

`as_backend_constant`, 60 lines down the same file, never got that guard. It passes
`device=getattr(ref, "device", None)` unconditionally, and `ref` is always a real
backend array — so the default device is named on **every** call.

```python
 dtype = getattr(ref, "dtype", None)
 device = getattr(ref, "device", None)
+if device is not None and device == default_device(xp):
+    device = None
```

## How it was found

Patching `jax._src.pjit.with_sharding_constraint` and counting the innermost galpy
frame: **7.00 calls per `NonInertialFrameForce._force`, 100% of them from
`_coerce.py:164`**.

## Measured (develop worktree, jax CPU; A/B arms verified to differ)

| | before | after | |
|---|---|---|---|
| `_force` per call | 5069.7 us | 3064.0 us | **1.65x** |
| `test_arbitraryaxisrotation_omegafunc_nullpotential` (jax eager) | 414.5 s | 273.6 s | **1.51x** |

## Blast radius

Not a noninertial fix — **144 call sites**: streamdf 66, streamgapdf 20,
sphericaldf 11, NonInertialFrameForce 10, RotateAndTilt 9, streamspraydf 7,
EllipsoidalPotential 6, ... i.e. exactly the slow-skip clusters.

## Safety

* **numpy path untouched** — `if xp is numpy: return value` precedes the change,
  so it is byte-identical by construction.
* **GPU preserved** — a NON-default anchor (a CUDA `ref` on a CPU-default run)
  still names its device; only the redundant default naming is skipped. Same
  logic already blessed in `coerce_coords`.

## Ledger delta

**None.** The second measurement crosses the 300 s per-test CI cap, but
un-deferring requires all 17 noninertial entries measured — a separate ledger PR.
slow_skip 68, xfail 8, skip 68, all unchanged.

## Tests

Three tests added to `tests/test_backend_coerce.py` (its existing home; no new
file, no duplicated helper). A/B-verified — the two device tests **FAIL on both
jax and torch** with the guard reverted:

```
arm=BASE   2 failed, 10 passed
arm=FIXED  12 passed
```

Gate (this branch, all three arms green):

```
numpy  708 passed, 1 skipped     (incl. test_noninertial.py, test_streamgapdf_impulse.py)
jax    694 passed, 1 skipped     (forced)
torch  694 passed, 1 skipped     (forced)
```

### DIFF
--- a/galpy/backend/_coerce.py
+++ b/galpy/backend/_coerce.py
@@ -160,6 +160,10 @@ def as_backend_constant(xp, value, ref):
         return value
     dtype = getattr(ref, "dtype", None)
     device = getattr(ref, "device", None)
+    if device is not None and device == default_device(xp):
+        # naming the default device commits the placement instead of letting the
+        # backend do it (~3x); same guard as coerce_coords -- see default_device.
+        device = None
     try:
         return xp.asarray(value, dtype=dtype, device=device)
     except TypeError:  # pragma: no cover - namespace without device= kwarg
--- a/galpy/backend/_coerce.py
+++ b/galpy/backend/_coerce.py
@@ -61,8 +61,8 @@ import numpy
 from ._namespaces import (
     _is_floating_dtype,
     asarray_on_device,
-    default_device,
     device_of,
+    effective_device,
     is_backend_array,
 )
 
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Self-audit follow-up (`0f91eec8`): the first commit left the same default-device predicate open-coded in **both** `coerce_coords` and `as_backend_constant`. That duplication is the defect class itself — the guard existed in one and was missed in the other — so it is now hoisted to `effective_device(xp, device)` in `_namespaces.py`, alongside `default_device`/`device_of`/`asarray_on_device` which already own the device vocabulary.

Pure refactor, no behaviour change. Re-verified on the new head: sharding constraints per `NonInertialFrameForce._force` still **0.00** (7.00 before the guard), and the helper suites pass on all three backends (numpy 41 / jax 41 / torch 41, 1 skipped each).

PR #1373, opened 2026-08-24
Label a torus by its integrals of motion by name, and accept Quantities

jobovy-agent:
Stacked on #1362 (base `staeckel-interp`); review that one first.

### Naming the integrals

A torus is labelled by its integrals of motion as much as by its actions, and when the integrals are what you have, inverting the actions to recover them is both slower and less direct. #1362 exposed this through an `integrals=True` flag that reinterprets the first three positional arguments, which turned out not to fit galpy's units machinery: `actionAngleInverse_physical_input` understands actions and angles only, so passing an energy as a Quantity raised `Input units not understood`.

Rather than special-casing the Stäckel class, subclasses now declare the names and dimensions of their integrals,

```python
_integral_labels = (("E", "energy"), ("Lz", "angmom"), ("I3", "energy"))
```

and the base class takes them as keywords:

```python
aASI(angler, anglephi, anglez, E=E, Lz=Lz, I3=I3)
aASI.Freqs(E=E, Lz=Lz, I3=I3)
```

Naming them is what makes this work *without* units as well. Dimensions alone could tell `(E, Lz, I3)` from the actions when Quantities are supplied — `(energy, action, energy)` against `(action, action, action)` — but in internal units both are just floats, and the unitless path is the one the tests and any performance-sensitive caller use. The mechanism is generic, so the other four inverse classes can adopt it by declaring their own labels.

### E and I3 on an interpolated torus

Reading the integrals off an interpolated torus and asking for that torus back did not return the same torus. `_interp_torus` took E and I3 from their *interpolated* values while `_grid_coords` inverts the grid *relations*, and between nodes the two drift apart:

```
E interpolated as a stored profile : -1.556639606440543
E from the grid definition         : -1.556643848981158
```

That 4.2e-06 showed up as the energy error of a torus requested by its integrals, against 1.6e-15 by actions. Taking both from the grid definition makes the two directions exact inverses, and is *also* more accurate, since E then comes from the circular- and outer-energy splines at the exact L_z rather than from a B-spline across the L_z grid:

| | before | after |
|---|---|---|
| round trip (grid index) | 5e-05 | 1e-15 |
| δH by integrals | 4.2e-06 | 1e-15 |
| max abs ΔJ_R vs forward code | 5.5e-05 | 2.2e-05 |
| max abs ΔJ_z vs forward code | 8.0e-05 | 6.9e-05 |

### Also

- Quantity inputs for `Es`/`Lzs`/`I3s` and for `Rmin`/`Rmax`/`Rinf`.
- A Sphinx page for the class, listed with the other inverse modules.

### Tests

Three added, covering the named form across `__call__`/`xvFreqs`/`Freqs`, integrals as Quantities agreeing with internal units, the missing-integral error, and the round trip. 62 inverse tests and all 244 of `test_quantity.py` pass; `actionAngleStaeckelInverse.py` and `actionAngleInverse.py` are at 100% and the new decorator has no uncovered lines.

### Measured: this is worth ~an order of magnitude in accuracy

Not a convenience feature. Going through the actions costs accuracy, because `_integrals_from_actions` inverts them to grid coordinates through a chain of `map_coordinates` lookups over the rho-hat/zeta tables; naming the integrals skips that chain entirely.

Round-trip error for three KuzminKutuzov orbits (worst over R, vR, vT, z, vz), same grid, same angles, the two entry points side by side:

| grid | orbit | via actions | via integrals | ratio |
|---|---|---|---|---|
| 11^3 | 1 | 1.32e-05 | 1.32e-06 | 10x |
| 11^3 | 2 | 8.21e-06 | 1.12e-07 | 73x |
| 11^3 | 3 | 1.62e-05 | 2.86e-06 | 5.7x |
| 15^3 | 1 | 2.32e-06 | 3.74e-07 | 6.2x |
| 15^3 | 2 | 1.61e-06 | 2.07e-07 | 7.8x |
| 15^3 | 3 | 6.84e-07 | 1.10e-07 | 6.2x |
| 17^3 | 1 | 2.38e-06 | 3.60e-07 | 6.6x |
| 17^3 | 2 | 1.16e-06 | 4.34e-08 | 27x |
| 17^3 | 3 | 7.26e-07 | 8.76e-08 | 8.3x |

So on an 11^3 grid the difference is roughly 1e-5 against 1e-6. The action inversion is the dominant error term of the interpolated evaluation, not the interpolation of the tori themselves.

The speed argument, by contrast, is weak: the shortcut saves 25.9 us against a ~1950 us evaluation, about 1.3%. Accuracy is the reason to have it.

(A residual plateau survives on both paths -- orbit 1 stalls at 3.6e-7 by integrals as well -- so the inversion contributes a roughly constant factor rather than being the whole error budget. Ruled out as causes of that plateau: the stored chi mesh (`nchi_store` 201 vs 801 identical at 15^3 and 17^3), `grid_pad` (2.34/2.32/1.99e-6 across a 10x range), and the measurement itself (the direct, non-interpolated inverse round-trips to 1.2e-10 through the same metric).)

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -12,6 +12,16 @@ v1.12.1 (expected around 2026-11-01)
   the interpolation of the actions is useless rather than merely less
   accurate.
 
+- The inverse action-angle methods can label a torus by its integrals of
+  motion by name where a class supports it, e.g.,
+  ``aASI(angler,anglephi,anglez,E=E,Lz=Lz,I3=I3)`` for
+  ``actionAngleStaeckelInverse``, rather than by its actions; this saves
+  inverting the actions when the integrals are what one has. Naming the
+  integrals works the same with and without units, unlike relying on their
+  dimensions, because in internal units the integrals of motion cannot be
+  told apart from the actions. ``actionAngleStaeckelInverse`` also accepts
+  Quantity inputs for the tori it is given and for the extent of its grid.
+
 - ``OblateStaeckelWrapperPotential`` chooses its reference curve ``u0`` from
   the focal length when none is given, placing it at R=1 rather than on the
   symmetry axis. V(v) is built along u=u0 and only represents the wrapped
--- a/doc/source/reference/aa.rst
+++ b/doc/source/reference/aa.rst
@@ -59,6 +59,7 @@ Specific actionAngle modules
 
    actionAngleHarmonicInverse <aaharmonicinverse.rst>
    actionAngleIsochroneInverse <aaisochroneinverse.rst>
+   actionAngleStaeckelInverse <aastaeckelinverse.rst>
    actionAngleTorus <aatorus.rst>
    actionAngleVerticalInverse <aaverticalinverse.rst>
 
--- a/doc/source/reference/aastaeckelinverse.rst
+++ b/doc/source/reference/aastaeckelinverse.rst
@@ -1,0 +1,7 @@
+.. _api_aa_staeckelinv:
+
+actionAngleStaeckelInverse
+==========================
+
+.. autoclass:: galpy.actionAngle.actionAngleStaeckelInverse
+   :members: __init__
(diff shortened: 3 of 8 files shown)


jobovy (approved):

jobovy:
pre-commit.ci run

PR #1374, opened 2026-08-24
special: backend-op gammainc/gammaincc, differentiable in the order

jobovy-agent:
Makes `gammainc`/`gammaincc` differentiable **with respect to the order** on torch, which they are not today, without slowing the forward pass.

## The gap

`torch.special.gammainc`/`gammaincc` have no derivative w.r.t. the order:

```
NotImplementedError: the derivative for 'igamma: input' is not implemented
```

(and `igammac` likewise), so any gradient through `a` is blocked upstream. d/dx works; jax has both and keeps its natives; numpy keeps scipy, bit-identically. **This is a torch-only change.**

## Design: native forward, custom backward

The obvious approach — evaluate the whole thing in backend ops — was measured and rejected: the series/continued-fraction is **485× native on a scalar** and 41× on n=1000, and `PowerSphericalPotentialwCutoff` (a component of `MWPotential2014`) calls this on nearly every evaluation, with scalars. That regressed 3 torch orbit shards, a torch potential shard, and OOM-killed the `test_backend` shard.

The loop does not have to run on the forward pass:

- **d/dx has a closed form**, `dP/dx = x^(a-1) e^-x / Γ(a) = prefix(a,x)/x`, and `prefix` is a cheap loop-free helper;
- **d/da has no closed form** and does need the series/CF — but only when the order actually requires grad, which no hot path does.

So torch gets a `torch.autograd.Function` whose forward is the native kernel and whose backward supplies the analytic d/dx and, on demand, the series/CF d/da. Calls that differentiate nothing skip the wrapper entirely.

| µs/call | native | routed | uniform-fallback (rejected) |
| --- | --- | --- | --- |
| scalar | 11.7 | **24.4** | 5216.8 |
| n=1000 | 253.5 | **269.7** | 6912.4 |

## Correctness

- d/da matches a central difference to ~6e-11, and equals jax's independent native d/da to 10 digits.
- d/dx is **exact** (2.1e-16) — it is the closed form, not differentiated quadrature.
- x = 0 and x = inf return exact limits; both are reachable (total mass and potential-at-infinity evaluate at r = inf).
- numpy path untouched and bit-identical to scipy.

The series/CF itself was validated against mpmath at 60 digits over a∈[0.5,200], x/a∈[0.3,4]: worst 7.3e-15 (P) and 2.4e-14 (Q), against scipy's own 1.0e-13 / 1.1e-13. It ships as the d/da backward.

## Scope: this does NOT fix the a > 20 precision gap

torch switches algorithm near a = 20 and loses ~6 digits above it (~5e-10 vs ~1e-16 below). Keeping the native forward means that behaviour stays. It is **pre-existing, not a regression** — torch used the native forward there before this PR too — but it is worth being explicit that this PR buys differentiability only. The accurate series/CF is in the tree and exercised, so routing the a>20 forward to it later is a routing change, not a rewrite.

The large-order parity test therefore pins native's accuracy (rtol 1e-8) with that reason inline, rather than asserting a precision galpy does not currently buy on this path.

## Tests

Four, in `tests/test_backend_special.py`. The pre-existing `_A` grid tops out at a=3.5 — entirely below the failing regime — so the new parity test crosses a=20 explicitly. `test_gammainc_grad_wrt_order` was verified to fail without the change (the real `NotImplementedError`); `test_gammainc_grad_wrt_argument_still_works` guards the derivative torch already had; `test_gammainc_endpoints_zero_and_infinity` pins both endpoints exactly.

functorch needs both the split `forward`/`setup_context` form and `generate_vmap_rule` — the spherical DFs reach this through `torch.func` vmap/grad, and each was required to get the suite green.

### DIFF
--- a/galpy/backend/special/_fallback/gammainc.py
+++ b/galpy/backend/special/_fallback/gammainc.py
@@ -1,0 +1,152 @@
+###############################################################################
+#   Fallbacks for the regularized incomplete gamma functions P(a,x) and
+#   Q(a,x) = 1 - P(a,x). Needed on torch, which has ``torch.special.gammainc``
+#   but (a) no derivative w.r.t. the ORDER a -- ``NotImplementedError: the
+#   derivative for 'igamma: input' is not implemented`` -- and (b) a discrete
+#   algorithm switch at a ~ 20 that costs ~6 digits above it (|dQ| ~ 5e-10 for
+#   a >= 21 vs ~1e-16 below). Both are cured at once by evaluating in backend
+#   ops: autodiff then supplies d/da and d/dx alike. jax's native versions are
+#   accurate and differentiable in both arguments, so the router does not reach
+#   this fallback for jax; numpy uses scipy.
+#
+#   Standard split, each branch converging where it is used:
+#     - x <  a+1: the ascending series  P = prefix * sum_n x^n / (a...(a+n))
+#     - x >= a+1: the modified-Lentz continued fraction for Q.
+#
+#   The delicate part is the shared prefix  x^a e^-x / Gamma(a),  NOT the
+#   series. Writing it as exp(-x + a ln x - lnGamma(a)) sums three terms that
+#   are each O(a ln a) and cancel to O(1) -- at a = x = 80 that is
+#   -80.0000 + 350.5621 - 269.2911 = 1.2710. Each term carries |term|*eps, so
+#   exp() returns ~1e-13 relative error however many series terms are taken;
+#   more iterations cannot fix it. Instead use the Stirling form, in which the
+#   large parts cancel ANALYTICALLY:
+#
+#       x^a e^-x / Gamma(a) = e^{-a*(lam - 1 - ln lam)} * sqrt(a/2pi) * e^{-w(a)}
+#
+#   with lam = x/a and w the Stirling remainder. Now the exponent is O(1) by
+#   construction. That form needs a >~ 14 for w's asymptotic series; below it
+#   the naive exponent is itself accurate (lnGamma(a) is O(1) there, so there is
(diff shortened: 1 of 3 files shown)


jobovy-agent:
**Moving to draft — the uniform fallback is too expensive for the hot path.**

CI found it and the measurement is unambiguous:

| torch `gammaincc` | native | this fallback | ratio |
| --- | --- | --- | --- |
| scalar | 10.8 µs | 5216.8 µs | **485×** |
| n=1000 | 170.4 µs | 6912.4 µs | **41×** |

`PowerSphericalPotentialwCutoff` is a component of `MWPotential2014` and calls `gammainc` on essentially every evaluation, mostly with scalars. Replacing one native call with 170 unrolled backend ops there is what produced the failures: 3 torch orbit shards, a torch potential shard, and both `test_backend` twins (the latter killed with exit 137 — OOM from the autograd graph retained across those iterations).

The accuracy work stands (worst 7.3e-15 / 2.4e-14 vs scipy's 1.0e-13 / 1.1e-13, verified against mpmath at 60 digits). What was wrong was applying it *uniformly* — I validated accuracy exhaustively and never timed the function or asked who calls it.

**Redesign:** pay the loop only where native actually fails.
- **forward** — native torch (exact for a ≤ 20); route to the series/CF only for the a > 20 precision regime.
- **d/dx** — closed form, `dP/dx = x^(a-1) e^-x / Γ(a)`, which is exactly the prefix already computed. No loop.
- **d/da** — no closed form, so it needs the differentiated series/CF, but only when `a` actually requires grad, which no hot path does.

A `torch.autograd.Function` with a native forward and that backward keeps both wins (d/da, and the a>20 precision) at native cost on every call that does not differentiate with respect to the order.

Issue #29, opened 2011-12-07
higher order moments of diskdf


jobovy:
skewness and kurtosis

PR #1384, opened 2026-08-25
actionAngleVerticalInverse: canonical evaluation mode, on by default

jobovy-agent:
Stage 1 of the canonical-inverse program (fast-orbits `STAECKEL_CANONICAL_MATH.md` §5.1/§8) — the smallest complete instance of the architecture, repairing measured defects in merged code.

### The problem, measured

The interpolating evaluation stores coefficient values (`_nSn`) and their J-derivatives (`_dSndJ`) as **separate** tables. Exact tables satisfy the symplectic consistency relation between them; independently interpolated ones don't, and the assembled map's symplectic defect sits at the derivative-table interpolation error:

| | min / median / max over a sweep |
|---|---|
| old mode | 2.9e-06 / **2.5e-04** / 1.5e-03 |
| canonical mode | 8.0e-12 / **1.3e-10** / 4.9e-10 |
| canonical + 1e-5 table noise | unchanged |

The map's own *accuracy* is orders better than its old defect — accuracy and canonicity are independent axes — but a symplectic integrator built on the old chart cannot converge to machine precision by shrinking its step.

### The canonical mode

Everything the evaluation uses is an interpolant differentiated exactly, or a closed form:

- `d(S̄_n)/dj` = the exact derivative of the **value** table (cubic B-spline stencil along the energy axis) chained through `E(j)`;
- the compensation for the energy-dependent auxiliary frequency is closed form — `∂W^A/∂ω` at fixed `(x, J^A)` equals `x_a v_a/2ω = J^A sin(2θ^A)/2ω` — and enters the angle Newton;
- the returned frequency is the `E(j)` interpolant's **own derivative**: exactly consistent with the chart, which is what a symplectic splitting needs (`H = H̃ + (H−H̃)` with the kick carrying all approximation error).

The noise-invariance line is the defining property, now a unit test: perturbing every stored coefficient changes *which* canonical map the tables define, never *whether* it is one.

### Scope and defaults

- **Default ON** for interpolating instances without a point transformation (Jo-authorized behavior change). There the stored `_nSn` are gauge-compatible as-is: the identity PT has no shear, so the equal-time and cotangent-lift gauges coincide.
- `canonical=False` restores the old behavior exactly.
- **Point-transformed tori keep the old path** (with `canonical=True` raising a clear error): their equal-time gauge carries the ξ-shear, so the canonical form needs recomputed tables — the next PR.
- The old derivative tables are not deleted: they become the consistency diagnostic `check_canonical_consistency()`, verifying the three-way relation `D_n = dS̄_n/dj + (ω′/2ω)·g_n`. With the compensation, interior-node mismatch is 1e-6–7e-5 against a 3.4e-2 table scale; **without it, 5e-2–7.5e-2 — larger than the tables themselves** (the compensation term is the dominant content of `D_n` at n=2, as the closed form predicts).

### Tests

Five new: the defect sweep asserted at the test floor, the noise-injection manifest test (with a check that the noise really moved the evaluation), the consistency relation with a sharpness check (dropping the compensation must degrade it >10×), frequency-consistency, and the two error paths. One legacy tolerance widened 1e-10 → 1e-9 with the reason in a comment: the chart-consistent frequency is a shade less accurate against the truth than the separately interpolated one it replaces (3.8e-10 relative here).

47 vertical tests pass; `actionAngleVerticalInverse.py` at 100% coverage. Diff ≈ 330 lines including tests.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -12,6 +12,19 @@ v1.12.1 (expected around 2026-11-01)
   apart by ~1e-15, which the symplectic dxdv bit-identity test caught
   intermittently depending on which machine it ran on.
 
+- ``actionAngleVerticalInverse`` gained a canonical evaluation mode, the
+  default for interpolating instances without a point transformation. The
+  angle map now uses the exact derivative of the interpolated coefficient
+  tables plus a closed-form compensation for the energy-dependent auxiliary
+  frequency, and the returned frequency is the interpolated Hamiltonian's
+  own derivative, so the assembled (J, angle) -> (x, v) map is exactly
+  symplectic for any stored tables: the measured symplectic defect drops
+  from the interpolation error of the old separately stored derivative
+  tables (median 2.5e-4) to the finite-difference test floor (~1e-10), and
+  is invariant under noise injected into the tables. The old behavior is
+  available with canonical=False; the derivative tables are kept as a
+  consistency diagnostic (check_canonical_consistency).
+
 - ``OblateStaeckelWrapperPotential`` chooses its reference curve ``u0`` from
   the focal length when none is given, placing it at R=1 rather than on the
   symmetry axis. V(v) is built along u=u0 and only represents the wrapped
--- a/galpy/actionAngle/actionAngleVerticalInverse.py
+++ b/galpy/actionAngle/actionAngleVerticalInverse.py
@@ -52,6 +52,7 @@ class actionAngleVerticalInverse(actionAngleInverse):
         maxiter=100,
         angle_tol=1e-12,
         bisect=False,
+        canonical=None,
         **kwargs,
     ):
         """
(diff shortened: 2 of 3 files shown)


jobovy-agent:
Closing as superseded by #1415 (merged today as 70032808, stack #1518). This PR canonicalized the older Fourier evaluation by differentiating the interpolants it reads; what landed instead makes the momentum-matched canonical map the default evaluation, which is manifestly symplectic for any table content by the same principle, and keeps the older evaluation unchanged behind momentum_matched=False. Two things from here carried over: the principle itself (the evaluation must differentiate the same interpolant it evaluates; separately stored derivative tables are exactly what breaks canonicity) and the Freqs routing decision, which on main means Freqs returns the map's own dE/dJ so it is exactly the frequency of the returned orbits. The older evaluation's symplectic defect (~1e-2 at nodes as well as between them) is measured in fast-orbits #29; the derivation of the map is fast-orbits #26.

PR #1385, opened 2026-08-25
actionAngleVerticalInverse: canonical evaluation for all point-transformation modes; old interpolated path removed

jobovy-agent:
Second and final PR of the 1D stack (base `aa-vertical-canonical`, #1384; gh stack #1386); review #1384 first. Now covers **every interpolating configuration**: no PT, polynomial PT, and — per Jo's requirement that everything be canonicalized, the exact PT being what works for extreme orbits — the exact PT, canonically lifted.

### Why the PT modes need more than #1384

#1384 reused the stored `_nSn` because the identity PT has no shear. With a real PT the class works in the equal-time gauge (`va = v/π′`). A correction from review: that gauge **is** a proper canonical transformation (generated by S_π = πv + Ξ, and the implemented stage is its on-torus restriction — the sheared momentum collapses to v/π′ exactly on the torus). What fails under interpolation is the same thing that always fails: cross-table consistency. Old-mode measured defects: poly-PT **6.1e-4 / 8.9e-4 / 1.1e-3**, exact-PT **1.4e-8 / 3.8e-6 / 6.8e-4**.

### What this PR does

- **One gauge for every mode — the shear-free cotangent lift** — with the mapping coefficients recomputed at setup by direct sampling (x_a on the aux loop, x = π(x_a), p_a = v·π′; the PT is never inverted). Canonical action labels are the zero modes ⟨J^A⟩ (gauge-independent loop actions) and become the public `J(E)`/`E(j)`.
- **The exact PT keeps its extreme-orbit position map**, held as ONE order-3 table whose π′, ∂²π/∂u², and torus-direction derivatives are its own stencil derivatives — the legacy quintic value/derivative *pair* is precisely the contingent storage this mode exists to avoid. The price of the gauge: S̄ is no longer ~0 (table scale 4e-2, comparable to no-PT; 1-D tables are cheap).
- **Compensations in closed form**: the ω-chain (J^A sin2θ^A/2ω) and the PT-parameter chain (−v·∂π/∂E·E′(j)) assembled from the stored interpolants' own derivatives.

### A failed attempt worth recording

The first exact-PT implementation canonicalized the equal-time gauge itself through a characteristic-function shortcut; the defect harness rejected it at **1.6e-1** — the shortcut's toy (tangent-push of the HO through π) is not canonical, and the equal-time family genuinely needs Ξ explicitly, branch-glued. That derivation stays open as a refinement (it would recover S̄ ≈ 0 tables — the equal-time sampling machinery verified S̄ = 6.4e-10 during development, so the ingredients exist). The harness catching a wrong derivation at 1e-1 while passing right ones at 1e-9 is the review process working.

### Measured

| mode | old defect | canonical defect |
|---|---|---|
| no-PT | 2.2e-05 / 2.7e-04 / 1.5e-03 | 1.1e-11 / 1.6e-10 / 8.2e-10 |
| poly-PT | 6.1e-04 / 8.9e-04 / 1.1e-03 | 4.6e-11 / 3.1e-10 / 6.6e-10 |
| exact-PT | 1.4e-08 / 3.8e-06 / 6.8e-04 | **9.7e-10 / 9.8e-09 / 3.2e-08** |

**Timing (the requested time test)**: exact-PT canonical measured 79 µs/pt vs 15 µs old before the Newton fix. To correct a possible misimpression: **no quadrature happens at evaluation time** — the landed shear-free compensation is closed form; the per-point Gauss–Legendre integral existed only in the rejected equal-time attempt. The slowdown was the finite-difference Newton Jacobian (three residual evaluations per iteration) plus the Python-side stencil; the final commit replaces the Jacobian with the approximate-analytic one (main terms exact, the small compensation derivative dropped — this only affects convergence *rate*; the residual equation defining the map is exact, so canonicity is untouched).

### The old interpolated path is removed (Jo: no backwards-compat concerns)

It was strictly dominated — worse symplectic defect at the same accuracy class — so interpolating instances now *always* evaluate canonically and the `canonical` kwarg is gone. The per-torus non-interpolating mode is unchanged (exact at its nodes). `pt_only` — an equal-time-gauge construct — now requires `setup_interp=False` (tested error). The `dSndJ` tables and PT-table accessors survive only as diagnostics (`check_canonical_consistency()`, which refuses PT instances with an explanatory error since their stored tables live in the other gauge). Two legacy tolerances adjusted with reasons in comments. 48 vertical tests pass; module coverage 100%.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -30,10 +30,12 @@ v1.12.1 (expected around 2026-11-01)
   compensation for every energy-dependent parameter chain -- the auxiliary
   frequency and the point transformation's shape -- is applied in closed
   form, and the canonical action labels (the gauge-independent loop
-  actions) become the public J(E)/E(j). The old behavior is available with
-  canonical=False; pt_only, an equal-time-gauge construct, is incompatible
-  with the canonical mode; the derivative tables are kept as a consistency
-  diagnostic (check_canonical_consistency).
+  actions) become the public J(E)/E(j). The old interpolated evaluation is
+  removed outright (it was strictly dominated: worse symplectic defect at
+  the same accuracy class); the per-torus non-interpolating mode is
+  unchanged and exact at its nodes; pt_only, an equal-time-gauge construct,
+  now requires setup_interp=False; the derivative tables survive only as a
+  consistency diagnostic (check_canonical_consistency).
 
 - ``OblateStaeckelWrapperPotential`` chooses its reference curve ``u0`` from
   the focal length when none is given, placing it at R=1 rather than on the
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -22,16 +22,18 @@ v1.12.1 (expected around 2026-11-01)
   from the interpolation error of the old separately stored derivative
   tables (median 2.5e-4) to the finite-difference test floor (~1e-10), and
   is invariant under noise injected into the tables. The canonical mode
-  covers instances with and without a polynomial point transformation: the
+  covers every interpolating configuration -- no point transformation, the
+  polynomial one, and the exact one (canonically lifted, preserving its
+  extreme-orbit position map): the
   mapping coefficients are recomputed in the shear-free gauge by direct
   sampling (no inversion of the point transformation is needed), the
   compensation for every energy-dependent parameter chain -- the auxiliary
   frequency and the point transformation's shape -- is applied in closed
   form, and the canonical action labels (the gauge-independent loop
   actions) become the public J(E)/E(j). The old behavior is available with
-  canonical=False; the exact point transformation keeps the old path for
-  now; the derivative tables are kept as a consistency diagnostic
-  (check_canonical_consistency).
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Closing as superseded. Its two aims were reached by other routes: (1) canonicity for the point-transformation modes is now provided by the momentum-matched map (#1415, merged today, stack #1518), which is itself a point transformation and is the default evaluation; (2) the polynomial PT's mislabelled tori are fixed on main by #1510 (actual-action labels) and #1516 (the sheared-gauge offset J^A - J = (1/2pi) Int v (pi'^-2 - 1) dx computed in closed form at construction and applied), so the old modes now return the requested torus at their nodes. The old interpolated path was not removed: it stays reachable with momentum_matched=False and use_pointtransform, as the comparison baseline in fast-orbits #29, where its ~1e-2 symplectic defect at the nodes is documented. If canonicalizing the old modes is ever wanted, this branch is the reference.

Issue #22, opened 2011-10-19
Add API for adding potentials to web docs


jobovy:
Also on how to add C implementations

PR #1387, opened 2026-08-25
actionAngleSphericalCanonical: the spherical inverse, built the Stäckel-canonical way (S1)

jobovy-agent:
**S1 of the spherical canonical program** (`STAECKEL_CANONICAL_MATH.md` §9, fast-orbits #22): the spherical inverse with no auxiliary torus in the construction — exact node tori by direct radial quadrature, lifted onto a manifestly canonical Fourier transform against an isochrone toy in the target's own coordinates. `aainv-spherical-dev` is the frozen baseline this must match or beat (S2's comparison).

### The construction

- **Exact nodes**: turning points by root-finding on the exact momentum relation; samples placed on the torus by construction (p_r from the energy relation, never a fit).
- **The closed-form correspondence**: the toy shares the target's coordinates, so L^A = L pointwise and the lattice is *purely radial* — the legacy 1-D lattice shape, derived. σ by 1-D spectral pullback; canonical action labels are the gauge-independent zero modes (verified against independent loop-action quadratures — the Stokes identity); sine-only parity at 1e-16 with the cosine coefficients kept as the diagnostic.
- **The frozen toy, position-matched (the winding condition)**: the correspondence τ → θ^A only winds once per radial period when the toy's *circular point* lies inside the target loop in the (r, p_r) plane — near-circular tori make this a tight constraint that frequency-magnitude matching (and any blind GM escalation) violates, producing sawtooth tables. So S1 pins the toy's circular radius at the central L to the target's (closed form), takes b from the frequency-ratio rule, and escalates for boundness *along the rc-pinned family* (b and GM together), with the winding condition checked for every torus. Frequency magnitude stays free — the correspondence absorbs it exactly. (The per-torus frequency-matched toy returns in S2 as the varying, compensated form.)
- **Evaluation**: 1-D Newton for θ^A_r; the two companion angles shift by the lattice's L-chain; full 3-D reconstruction delegates to the analytic `actionAngleIsochroneInverse` (the design win of L^A = L). Discrete mode reads per-torus spectral correspondence tables (θ_r(τ), J^A(τ), ψ-shift by regular quadrature in τ — all periodic and finite).
- **Interpolation mode**: J_r, E, S_n tables on a rectangular (u, L) grid (quadratic energy spacing at the circular edge, the phase-2 lesson); labels invert implicitly (u from (J_r, L) by root-finding on the stored J_r interpolant — exact-in-the-family); every chain is the stored splines' own derivative; frequencies are the stored energy interpolant's derivative through the label chain (the integrator contract).

### Measured (log halo; tests assert all of this)

- **Symplectic defect (6×6, central differences): 3.7e-10 – 1.6e-9 across tori and angles — indistinguishable from the analytic isochrone inverse through the same harness** (2.5e-10 – 2.0e-9). 
- **Manifest**: 1e-5 relative noise in every stored table moves the evaluation by 6e-6 and leaves the defect at the floor.
- **Discrete reconstruction**: E conserved to 1e-14 on returned points; forward round trip (actionAngleSpherical) recovers actions at 1.8e-14 and all three angles at ~1e-9 (the forward code's own quadrature tolerance).
- **Interpolated (coarse 8×8 test grid)**: round-trip action error 3e-5, angles 1.3e-4 — the honest interpolation error, cleanly separated from canonicity.
- 100 % module coverage; six tests, ~9 s.

### S2 (next PR, stacked)

The varying compensated toy + the canonical radial alignment PT (support-matched, cotangent-lifted, compensated through the family interpolant), eccentricity coverage, the match-or-beat comparison against the frozen `aainv-spherical-dev` baseline, units + docs.

### DIFF
--- a/galpy/actionAngle/__init__.py
+++ b/galpy/actionAngle/__init__.py
@@ -9,6 +9,7 @@ from . import (
     actionAngleIsochroneApprox,
     actionAngleIsochroneInverse,
     actionAngleSpherical,
+    actionAngleSphericalCanonical,
     actionAngleStaeckel,
     actionAngleStaeckelGrid,
     actionAngleTorus,
@@ -40,6 +41,9 @@ actionAngleStaeckelGrid = actionAngleStaeckelGrid.actionAngleStaeckelGrid
 actionAngleIsochrone = actionAngleIsochrone.actionAngleIsochrone
 actionAngleIsochroneApprox = actionAngleIsochroneApprox.actionAngleIsochroneApprox
 actionAngleSpherical = actionAngleSpherical.actionAngleSpherical
+actionAngleSphericalCanonical = (
+    actionAngleSphericalCanonical.actionAngleSphericalCanonical
+)
 actionAngleTorus = actionAngleTorus.actionAngleTorus
 actionAngleIsochroneInverse = actionAngleIsochroneInverse.actionAngleIsochroneInverse
 actionAngleHarmonic = actionAngleHarmonic.actionAngleHarmonic
--- a/galpy/actionAngle/actionAngleSphericalCanonical.py
+++ b/galpy/actionAngle/actionAngleSphericalCanonical.py
@@ -10,6 +10,7 @@
 #   (STAECKEL_CANONICAL_MATH.md section 9 in the fast-orbits repository.)
 ###############################################################################
 import numpy
+from scipy.interpolate import RectBivariateSpline
 from scipy.optimize import brentq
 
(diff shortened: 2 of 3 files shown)


jobovy-agent:
Note: the class lands as `actionAngleSphericalCanonical` in this PR and is renamed to `actionAngleSphericalInverse` at the top of the stack (#1389, per review) — it replaces the frozen `aainv-spherical-dev` implementation of that name and does not depend on it.

PR #1396, opened 2026-08-26
actionAngleStaeckelInverse: the canonical family (T2)

jobovy-agent:
**T2 of the Stage-3 canonical program** (stacked on #1395; `STAECKEL_CANONICAL_MATH.md` §10 per the fo#23 review decisions): the 3-D canonical family, its compensated evaluation, and the removal of the interpolated-direct machinery.

### The family (part 1)

The direct grid's rectified (L_z, w_E, w_I) node lattice; all node tori in one vectorized canonical construction; prefiltered 3-D tables (labels J_R/J_z, energy, supports, both anomaly-map sine series) evaluated by an explicit four-tap tensor stencil whose value/derivative weights make **every chain the stored tables' own derivative**. Implicit-inverse labels (Q10.1: grid in the integrals) round-trip at 1e-16.

### The compensated evaluation (part 2)

A 2-D angle Newton on the compensated θ-equations — one term per degree per action chain, p^A_d(∂q^A_d/∂J_i)|_τ − p_d(∂q_d/∂J_i)|_τ, flux-grouped so every factor is closed-form and regular at turning points; toy phases from the isochrone's own ψ/Λ relations; delegation to the analytic isochrone inverse; per-degree un-lift; frequencies = the E-table's own chains (the integrator contract).

**The M2 adjudication**: the harness initially said 3.5e-2, h-flat, concentrated in the {J_φ, J_z} bracket — while frozen-map mechanics, all table chains, all closed chains, and comp ≡ −Σp∂π/∂J each verified exactly. The tie-breaker was an FD adjudicator on the generating function itself (branch-aware W by quadrature), which exposed the single ported bug: the toy's vertical momentum missing its 1/sin ϑ factor (a latitude-vs-polar convention, 2.4%, self-consistently wrong in every test that used it on both sides). Fixed: **defect 5.2e-10 – 2.2e-9 across tori and angles, indistinguishable from the analytic control; invariant under 1e-5 noise in every stored table** (evaluation moves by 8.5e-5).

### The removal (part 3, per Q10.4) — with M5 recorded first

Same-budget comparison before deletion: round-trip accuracy comparable (~1e-5-grade both), **symplectic defect 3.7e-4–1.1e-3 (old) vs 1.8e-9 (canonical)**, ~3× per-point cost. Then: χ-profile storage, `_ProfileSet`, the ρ/ζ action lookup, and the batched path deleted (net −231 lines); `setup_interp=True` always builds the canonical family. **Preserved with less machinery**: `integrals=True` and by-name integral labels (the rectified coordinates are closed forms — no inversion at all), directional out-of-grid messages, and the degenerate edges, which now carry their analytic limits (shell u, midplane, vanishing anomaly maps) with guarded degenerate-degree evaluation — J_z = 0 and J_R = 0 evaluate cleanly.

**Contract change, recorded in the adapted tests**: machine-constant H along off-node tori was the removed path's p = √W virtue, bought with contingent canonicity; the canonical family's H-constancy is its (converging) interpolation error while its defect sits at the FD floor for any stored tables.

28 tests, 100% module coverage, no pragmas; the under-resolution warning gives actionable npt/ncanon advice.

### DIFF
--- a/galpy/actionAngle/actionAngleStaeckelInverse.py
+++ b/galpy/actionAngle/actionAngleStaeckelInverse.py
@@ -1139,6 +1139,10 @@ class actionAngleStaeckelInverse(actionAngleInverse):
                 self._grid_coords(jr, jphi, jz), jphi, angler, anglephi, anglez
             )
         if self._interp:
+            if self._canonical:
+                return self._xvFreqs_canonical_interp(
+                    jr, jphi, jz, angler, anglephi, anglez
+                )
             return self._xvFreqs_index(
                 self._coords_from_actions(jr, jphi, jz),
                 jphi,
@@ -1331,6 +1335,10 @@ class actionAngleStaeckelInverse(actionAngleInverse):
             scal = self._interp_torus(self._grid_coords(jr, jphi, jz))[0]
             return (scal[3], scal[5], scal[4])
         if self._interp:
+            if self._canonical:
+                x = self._canon_coords(float(jr), float(jphi), float(jz))
+                _, dq = self._canon_family_chains(x)
+                return (dq[2, 0], dq[2, 1], dq[2, 2])
             self._interp_Lz = jphi
             scal = self._interp_torus(self._coords_from_actions(jr, jphi, jz))[0]
             return (scal[3], scal[5], scal[4])
@@ -1695,26 +1703,53 @@ class actionAngleStaeckelInverse(actionAngleInverse):
             )
             for kk in range(6):
                 out[kk, jj] = oo[kk][0]
+        Rt, vRt, vTt, zt, vzt, phit = self._canon_unlift(
+            out,
+            self._can_a[ii],
+            self._can_e[ii],
+            self._can_LA[ii],
+            self._can_thmin[ii],
+            self._can_Dmu[ii],
+            self._can_Dmv[ii],
+            self._umins[ii],
+            self._umaxs[ii],
+            self._vmins[ii],
+            Lz,
+        )
+        return (
+            Rt,
+            vRt,
+            vTt,
+            zt,
+            vzt,
+            phit,
+            self._OmegaR[ii],
+            self._Omegaphi[ii],
+            self._Omegaz[ii],
+        )
+
+    def _canon_unlift(self, out, a, e, LA, thmin, Dmu, Dmv, umin, umax, vmin, Lz):
+        """Map the toy-chart reconstruction to the target chart: per-degree
+        anomaly inversions through the stored maps (closed-form radius and
+        linear polar-angle inversions), momenta through the per-degree
+        action-flux groups (regular at all turning points), then prolate ->
+        cylindrical exactly as the direct path"""
         R, vR, vT, z, vz, phi = out
-        # un-lift per degree through the stored anomaly maps
         rA = numpy.sqrt(R**2 + z**2)
         vrA = (R * vR + z * vz) / rA
         pAth = vR * z - vz * R
-        a, e = self._can_a[ii], self._can_e[ii]
-        LA, thmin = self._can_LA[ii], self._can_thmin[ii]
         # u-degree: eta from the closed-form radius inversion, tau from the
         # stored map, u from the cosine anomaly; p_u through the flux group
         y = (numpy.sqrt(self._bc**2 + rA**2) - self._bc) / a
         coseta = numpy.clip((1.0 - y) / e, -1.0, 1.0)
         sineta = numpy.sign(vrA) * numpy.sqrt(numpy.clip(1.0 - coseta**2, 0.0, None))
         etau = numpy.arctan2(sineta, coseta) % (2.0 * numpy.pi)
-        tuu = self._tau_of_eta(etau, self._can_Dmu[ii])
-        Du = self._umaxs[ii] - self._umins[ii]
-        u = self._umins[ii] + Du * numpy.sin(tuu / 2.0) ** 2
-        s = numpy.sqrt(self._bc**2 + rA**2)
+        tuu = self._tau_of_eta(etau, Dmu)
+        Du = umax - umin
+        u = umin + Du * numpy.sin(tuu / 2.0) ** 2
         gA = a * e * (y + self._bc / a) / numpy.sqrt(y * (y + 2.0 * self._bc / a))
         ms = self._nforDm
-        detau = 1.0 + numpy.cos(tuu[:, None] * ms[None, :]) @ (ms * self._can_Dmu[ii])
+        detau = 1.0 + numpy.cos(tuu[:, None] * ms[None, :]) @ (ms * Dmu)
         sintu = numpy.sin(tuu)
         sru = numpy.where(
             numpy.fabs(sintu) > 1e-12,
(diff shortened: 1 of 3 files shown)


jobovy-agent:
**Back to draft — the family's angle map is missing content, found while benchmarking against the old code.**

Gridded comparison (KuzminKutuzov amp=4, ac=5, Δ=1.3; R∈[0.7,1.6], R∞=8; max over 4 tori × 12 angles):

| n³ | setup | eval/pt | action err | angle err | ΔH/H | defect |
|---|---|---|---|---|---|---|
| old 9³ | 36 s | 0.32 ms | 1.0e-4 | 7.4e-5 | 1.6e-15 | 7.5e-3 |
| old 11³ | 64 s | 0.32 ms | 2.9e-5 | **9.2e-6** | 1.9e-15 | 2.7e-3 |
| canon 9³ | 57 s | 1.58 ms | 1.0e-4 | 3.2e-3 | 1.6e-4 | 1.0e-8 |
| canon 11³ | 104 s | 1.61 ms | 2.5e-5 | **3.1e-3** | 3.9e-5 | 1.4e-8 |

Actions track the old code at every grid size and the defect is 5–8 orders better, but the **angle error stalls at ~3.1e-3** instead of converging.

Not the map resolution: at fixed 9³, raising `ncanon`/`npt` drives `node_maxdev` 2.5e-2 → 9.7e-6 → 2.3e-9 while the angle error stays 0.0031592 — identical to five digits.

Decisive test, at a torus **exactly on a grid node** (stored maps interpolate exactly there):

| chart | δθ_R | δθ_φ | δθ_z | δJ_r | ΔH |
|---|---|---|---|---|---|
| T1 discrete (this stack's #1395) | 3.2e-8 | 1.0e-8 | 1.1e-8 | 1.2e-11 | 2.7e-15 |
| T2 family (this PR) | **4.0e-3** | 1.6e-4 | **4.2e-3** | 1.2e-11 | 3.9e-13 |

Same torus, same actions recovered, energy fine — the family's *angles* are five orders worse than the discrete chart's. So it is a systematic error in the family's θ-equation, not interpolation.

**Cause**: this PR's family stores no S_n (only labels, energy, supports and the two anomaly maps), on the reasoning that the lattice collapses to ~1e-14 at every node. That is the KB94a trap: S_n vanishing does not make ∂S_n/∂J vanish, and it is that angle content the family needs. The discrete path gets angles right because it stores the measured correspondence fields (θ^A−θ per degree) instead of reconstructing them from the PT alone.

Why nothing here caught it: the symplectic-defect harness cannot — dropping a gradient term leaves the map perfectly canonical, merely a *different* canonical chart — and the PR's checks leaned on action round trips, which are clean. The angle round trip was under-tested; a converging-angle-accuracy test will be part of the fix.

Next: measure the missing field directly (family compensation vs the discrete Δ at the same node, resolved in θ^A), then extend the family storage to carry that angle content with its own derivatives.

jobovy-agent:
### Accuracy investigation concluded

Two commits added since the draft went up:

**Analytic d/dJ of every stored table** (turning points and both anomaly-map
degrees), from differentiating W(q; alpha) = 0 along the level set and the
action-matching condition A^A(eta) = A_t(tau) at fixed tau. The latter wants
to divide by p^A dq^A/deta, which vanishes at *both* turning points, so the
derivative deta/dJ|_tau is expanded in the same sine basis the map uses --
exact, since the maps are odd -- and a Galerkin system is solved instead.
Verified against finite differences of independently built neighbouring
tori: 5.6e-10 (supports), 1.2e-9 (u-degree), 4.7e-10 (v-degree).

**The supports are stored as regular combinations.** Storing the turning
points directly loses the half-width to cancellation wherever it vanishes:
one cell from the shell edge, umin and umax interpolate with errors that are
equal and opposite to three digits, and the half-width comes out 50% wrong.
Stored instead are the midpoint and K_u = (umax-umin)^2/4J_R,
K_v = (pi/2-vmin)^2/J_z -- bounded and smooth to the edge (0.50-0.89 and
0.37-1.11 over the grid) -- with the vanishing carried by the action, which
is known exactly at evaluation. On the 11^3 test grid, near-degenerate tori
improve 8-33x (worst 2.2e-2 -> 2.6e-3) and the planar-edge action round trip
goes 5.7e-5 -> 1.8e-9. The degenerate edges stop needing a special case.

### Against the old interpolated-direct code, at 13^3

| metric | old | new |
|---|---|---|
| symplectic defect | 4.7e-4 | **4.4e-9** |
| action error (rel) | 6.7e-6 | **1.1e-6** |
| angle error | **6.2e-6** | 7.8e-5 |
| setup / eval | 107 s / 0.32 ms | 172 s / 1.67 ms |

The old code also places points on an exact energy surface (spread 1e-15 vs
1.4e-6), since it builds the exact torus from interpolated integrals. That is
the trade: an exact torus with a non-canonical map, against an exactly
canonical map on an approximate torus.

### What the residual angle error is not

Measured, not assumed:

- **Not the anomaly-map resolution.** ncanon/npt 128/32 -> 512/96 drops
  node_maxdev 2200x and changes the angle error by *zero* significant
  figures (1.93e-2 -> 1.93e-2 at the worst grid point). The node_maxdev
  warning's implication that accuracy near those tori is limited to that
  scale is not borne out -- grid interpolation dominates there.
- **Not derivative-constrained interpolation.** Cubic Hermite tied to the
  exactly known node derivatives is still O(h^3) in the derivative; measured
  on the real tables it is worth 1.2-2.7x, so it moves the constant, not the
  rate.
- **Not interpolation order.** Quintic B-splines are 8-34x better on a 1-D
  interior test but a 3-5x *regression* in the family. Per-axis sweep places
  the whole regression on the energy axis: (3,5,3) gives 4.11e-4 against
  (3,3,3)'s 7.83e-5, while (5,3,3) and (3,3,5) are neutral.
- **Not the grid boundary.** The lowest-energy cell is among the most
  accurate.

What it *is*: error grows steeply and smoothly with energy -- 1.1e-4 at
w_E = 0.32 against 1.9e-2 at w_E = 0.92 -- because the high-energy tables
carry genuinely more structure per node. It converges at about h^3. Sizing
Rinf to what is actually needed is worth ~2x in the median and ~5x in the
actions (13^3, Rinf 8 -> 2.5), which is a documentation matter rather than a
code one.

jobovy-agent:
### Why this family stores no S_n

Worth recording, since the natural question about a canonical torus map is
where the generating-function lattice went.

Give the chart the EXACT stored values and derivatives of the exact torus at
the requested actions, and measure what is left.  That residual is the only
thing an S_n lattice could correct; everything above it is
parameter-interpolation error, which more stored content cannot fix.

  w_E    J_R      interpolated   exact params   ratio
  0.20   0.0185     1.16e-04       1.47e-07      789x
  0.32   0.0493     1.08e-04       1.63e-06       66x
  0.44   0.0989     4.18e-04       1.25e-07     3347x
  0.56   0.1738     1.02e-03       7.89e-08    12936x
  0.68   0.2858     3.36e-03       1.12e-07    30131x
  0.80   0.4549     7.26e-03       1.18e-07    61710x
  0.92   0.7159     1.93e-02       1.06e-06    18239x

The momentum-matched PT already carries essentially all of the content: the
chart reproduces the exact torus to ~1e-7 across the whole energy range,
including where the interpolated error is 1.9e-2.  The floor is also flat in
energy while the interpolated error grows 170x, so the difficulty is in
interpolating the parameters, not in representing the torus.  S_n would be
further interpolated tables, each adding its own interpolation error, to
correct a residual four to five orders below the error being chased.  This is
the same fact the spherical code shows from the other side, where the stored
S_n come out at ~1e-11.

Nor is there another reparametrization of the kind that (umin, umax) ->
(u_c, K_u) already exploited.  Substituting truth for one group of tables at
a time buys under 3x each -- supports 1.6-2.8x, D^u_m 1.1-1.3x, D^v_m and the
energy table ~1x -- and those do not compound anywhere near the gap to the
floor.  The error is spread evenly over every stored quantity, all of which
vary rapidly along the energy axis.  What remains is grid resolution, which
converges at about h^3.

PR #1400, opened 2026-08-27
integrate_dxdv: transform the base state in C, restoring bit-identity with plain integrate

jobovy-agent:
Fixes the intermittent `test_symplectic_dxdv_base_bit_identity` failures on CI — at the root, so bit-identity holds by construction rather than by coincidence.

### What was actually happening

The two paths built the *same* orbit from *different* code:

- `Orbit.integrate` passes **cylindrical** ICs into C, which converts them with `cyl_to_rect_galpy` (and back with `rect_to_cyl_galpy`) — i.e. the **C library's** `cos`/`sin`/`atan2`.
- `Orbit.integrate_dxdv` converted the base **in Python** (`coords.cyl_to_rect`/`cyl_to_rect_vec`, and `rect_to_cyl`/`rect_to_cyl_vec` on the way out) — i.e. **numpy's** vectorized trig.

The formulas are identical, but numpy dispatches its vectorized trig on CPU features, so its results need not match libm's in the last bit. Where they didn't, the two runs started from ICs differing by 1 ulp, and the integrators — which really are bit-identical; `evalRectForce_dxdv`'s base block is textually the same as `evalRectForce`, and the drift/kick kernels are element-wise over the base block — faithfully propagated that into a ~1e-15 divergence.

That explains the observed behaviour: same numpy (2.5.2) in both passing and failing runs, so it wasn't a version bump; identical failure value on repeat runs; and always reported for `leapfrog_c` only because the assert fires on the first method in the loop.

### The fix

The C dxdv entry now takes the base in cylindrical coordinates and applies the same `cyl_to_rect_galpy`/`rect_to_cyl_galpy` as the plain path (in, and on every output row); the caller no longer transforms the base for the C integrators. The deviation block is unchanged (rectangular in/out, chain-rule transforms in Python — it never feeds back into the base). The pure-Python (`odeint`/`dop853`) dxdv path keeps its numpy transform, which is unrelated to the C bit-identity claim.

### Verification

Perturbing the Python-side transforms by 1 ulp — a direct simulation of the CPU-dependent disagreement:

| method | before | after |
|---|---|---|
| leapfrog_c | 9.3e-15 | **0** |
| symplec4_c | 6.2e-15 | **0** |
| symplec6_c | 1.0e-15 | **0** |
| rk6_c | 7.1e-15 | **0** |

(Note all methods were affected, not just `leapfrog_c`.) That probe is now a test — `test_symplectic_dxdv_base_independent_of_python_coord_transforms` — so the invariant can't silently regress. Local runs: the dxdv suites (22 tests) plus `test_orbit.py -k "dxdv or lyapunov"` (116) and the `test_orbits.py`/`test_quantity.py` dxdv tests (4), all passing.

Left alone deliberately: the planar dxdv path, whose `getOrbit`-level difference is a *convention* difference (numpy `arccos` on [0,2π) vs C `atan2` on (−π,π]), already documented in `test_symplectic_dxdv_planar_base_bit_identity` and worked around there by asserting on the raw rectangular base. Changing that would change the returned φ range for planar dxdv, which is a user-visible behaviour change and out of scope here.

### DIFF
--- a/galpy/orbit/integrateFullOrbit.py
+++ b/galpy/orbit/integrateFullOrbit.py
@@ -755,9 +755,12 @@ def integrateFullOrbit_dxdv_c(
     """
     Integrate an ode for a fullOrbit+phase space volume dxdv in C.
 
-    Both the input state ``yo`` and deviation ``dyo`` as well as the output are
-    in the rectangular frame (x,y,z,vx,vy,vz | dx,dy,dz,dvx,dvy,dvz); the
-    cylindrical<->rectangular transforms are handled by the calling
+    The base state ``yo`` is given, and returned, in the cylindrical frame
+    (R,vR,vT,z,vz,phi): the C entry converts it with the same helpers the
+    plain ``integrateFullOrbit`` uses, so the base trajectory is
+    bit-identical to a plain integration. The deviation ``dyo`` is given, and
+    returned, in the rectangular frame (dx,dy,dz,dvx,dvy,dvz); its
+    (chain-rule) transforms are handled by the calling
     ``integrateFullOrbit_dxdv``.
 
     Parameters
@@ -922,7 +925,6 @@ def integrateFullOrbit_dxdv(
     -----
     - 2026-06-03 - Written based on integratePlanarOrbit_dxdv - Bovy (UofT)
     """
-    # Go to the rectangular frame: base state (R,vR,vT,z,vz,phi) -> (x,y,z,vx,vy,vz)
     R, vR, vT, z, vz, phi = (
         yo[:, 0],
         yo[:, 1],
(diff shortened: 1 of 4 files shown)


galpy/orbit/integrateFullOrbit.py
```diff
@@ -943,8 +942,24 @@ def integrateFullOrbit_dxdv(
             this_dyo[ii] = numpy.dot(jac, dyo[ii])
     else:
         this_dyo = dyo
-    this_yo = numpy.hstack((this_yo, this_dyo))
-    if int_method.lower() == "dop853" or int_method.lower() == "odeint":
+    _pure_python = int_method.lower() == "dop853" or int_method.lower() == "odeint"
```
jobovy: Usually, we detect C integration using the _c ending of the method (which is robust against new methods). So switch to that

jobovy (changes requested):

galpy/orbit/integrateFullOrbit.py
jobovy-agent: Switched to the `_c` ending (`int_method.lower().endswith("_c")`) in 2d56d35c — every C method ends in `_c` and no pure-Python one does, so this also stays correct as methods are added. Verified both branches still route correctly (`dop853`/`odeint` through the Python path, `dop853_c`/`rk6_c`/`leapfrog_c` through C).

jobovy (approved):

PR #1402, opened 2026-08-27
hyp2f1: one tanh-sinh rule for every B, and drop the c-B >= 1 constraint

jobovy-agent:
Follow-up to #1401, which routed only `B < 0.25` to tanh–sinh and kept Gauss–Legendre above it. This removes the split: **one rule for every B**, and the `c - B >= 1` constraint goes with it.

### The threshold was only protecting the less accurate rule

Measured against mpmath over **768** (A,B,c,z) cases with B *above* the old threshold — i.e. exactly the regime GL was being kept for:

| rule | worst rel err |
|---|---|
| GL (shipping) | **1.5e-04** |
| tanh–sinh | **2.9e-15** |
| cases where tanh–sinh is worse | **0** |

### Why it was kept, and why that reason dissolved

The traced-parameter path also routes through `_euler_quad`, and tanh–sinh could not size a grid from a tracer. That turned out not to be a real constraint: **one fixed grid is as accurate as the per-call adaptive one** (worst 2.67e-15 vs 2.89e-15 over 720 cases spanning B = 0.001…5.0, nothing above 1e-13). Being independent of the parameters is precisely what lets a traced B use the rule, so the concreteness guard disappears entirely — and with it the `X = xi^k` substitution, the `k <= 12` cap, the `expm1` crossover guard, and the tiny-|z| series branch. `(1 - z t)^{-A}` is simply 1 at z = 0, where GL's `1/|z|` substitution was singular and needed its own branch.

### Dropping `c - B >= 1` is part of the same change, not scope creep

That bound exists *only* to keep `(1-t)^{c-B-1}` non-singular for a fixed-order rule. The tanh–sinh weight cancels that endpoint analytically and needs just what the Beta integral needs, `c - B > 0`. Measured across the whole block the old bound excluded (`c - B` in (0, 1]): **2.3e-15**.

It also removes a float64 knife-edge. At `(a=5.0, b=0.001, c=1.001)` the old test asked `1.001 - 0.001 >= 1.0`, which is False by one ulp, so the Euler route was refused and a far worse one ran. Worst relative error over 900 cases through the **public** entry (including z = 0):

| tree | worst |
|---|---|
| develop (pre-#1401) | 8.807e-01 |
| #1401 | 4.577e-03 |
| here | **2.665e-15** |

### Three tests encoded the old contract — the middle one is the interesting one

* `test_fallback_unsupported_regimes_raise` asserted `_euler_labeling(2,2,2.5)` raises. `c - P = 0.5` is now supported, so it **failed** and the suite caught it. Replaced with both genuine ways out of the domain: `c - P <= 0` for every P, and no positive P at all.
* `test_hyp2f1_fallback_alt_labeling` still **passed**, and had quietly stopped testing what it exists for. It used `(2.0, 0.8, 2.5)`, chosen because `c-a=0.5` failed the old bound so labelling *had* to fall through to `b`. Under `c - P > 0`, `a` is admissible, so the case moved to the **first** branch and the second branch lost its only coverage — passing throughout, because either labelling gives the right value. Now `(-1.0, 0.8, 2.5)`: `a < 0` disqualifies `a` for good, whatever the bound.
* `test_hyp2f1_large_B_route_is_unchanged` guarded a route that no longer exists, with bars up to 1e-08 taken from GL's own accuracy. Renamed to `..._is_machine_precision` and retightened to the measured 1e-14; `(-1.6, 1.2, 3.6)` goes 5.3e-08 → 1.55e-15.

### Gates

* **numpy byte-identical**: sha256 `245bc0e3…` over 315 values matches develop `00a7153c` exactly, with the arms grep-asserted to differ (GL references 2 → 0, the `>= 1.0` guard 3 → 0). numpy routes hyp2f1 to scipy, so the fallback never runs there.
* **385 passed, 1 skipped** across `test_backend_special` / `_constantbetadf` / `_quadrature`, re-run *after* the rebase onto the merged base.
* Ledger delta: **0**.
* Net **deletion**: 91 insertions, 156 deletions.

### DIFF
--- a/galpy/backend/special/_fallback/hyp2f1.py
+++ b/galpy/backend/special/_fallback/hyp2f1.py
@@ -31,12 +31,10 @@ from ..._namespaces import (
     has_concrete_truth_value,
     is_backend_array,
 )
-from ._quadrature import gauss_legendre_01
 
 # 128 nodes: ~1e-10 or better vs scipy at realistic radii (|z| = r/a <~ 50);
 # accuracy degrades smoothly to ~1e-6 at the extreme |z| ~ 500 (r/a ~ 500, far
 # beyond any realistic galactic radius) for awkward exponent combinations.
-_NODES = 128
 
 
 # Terms for the series route below. Its argument is z/(z-1), so accuracy is set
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Closing a gap in my own evidence above.

Relaxing `c - B >= 1` also changes `_in_regime`, so cases with `0 < c - B < 1` now take the **Euler route instead of the Pfaff series**. Neither sweep in the description actually covered that: the 900-case public-entry sweep used `c - B` in {1.0, 1.3, 1.7, 2.5, 4.0}, all ≥ 1, and the `c - B ∈ (0,1]` sweep called `_euler_quad` *directly*, bypassing route selection. So the rerouted band was the one region where a regression could hide and I hadn't measured it end-to-end.

A/B through the public entry, 320 cases with `c - B ∈ (0, 1)`, vs mpmath (dps=40):

| tree | worst rel err |
|---|---|
| develop `00a7153c` (Pfaff series) | **4.33e-03** |
| this PR (Euler / tanh–sinh) | **1.55e-15** |

Worst case on develop is `(A=5.0, B=0.05, c=0.95, z=-40)` at 4.33e-03, and the same parameters land at ~1e-15 here. So the reroute is a large improvement rather than a risk — the series was the weaker option in that band too, which is consistent with the main result that the `>= 1` bound was a Gauss-Legendre artifact rather than a real domain limit.

Nothing in the PR changes; this is additional evidence for a region the description did not cover.

PR #1403, opened 2026-08-27
hyp2f1: the traced Euler-label selector must use the same rule as the concrete one

jobovy-agent:
Fixes the blocker in #1204 comment 5440895453.

### The miss

#1402 relaxed the concrete admissibility rule from `c - B >= 1` to `c - B > 0` in **three** places — `_in_regime`, `_euler_labeling`, `_in_regime_mask` — but there is a **fourth** copy inside `_nonpositive_traced`:

```python
oka = (ea > 0) & ((c - ea) >= 1.0)
```

so under tracing a first label with `0 < c-a < 1` was still rejected. The traced route then fell through to the **second** label, which in these cases is non-positive, and `T**(B-1)` with `B <= 0` is `inf` at the `t -> 0` node. Eager was exact; jit returned `inf`/`nan`.

Reproduced before fixing, and after:

| (a, b, c, z) | scipy | traced BEFORE | traced AFTER |
|---|---|---|---|
| (2.0, −0.2, 2.5, −0.7) | 1.09213476189 | `inf` | rel 0.0 |
| (2.0, −1.0, 2.5, −0.7) | 1.56 | `nan` | rel 1.1e-16 |
| (2.0, −0.5, 2.2, −0.7) | 1.27837324696 | `inf` | rel 2.4e-15 |
| (3.0, −0.2, 3.4, −2.0) | 1.2241297706 | `inf` | rel 3.6e-15 |

**Why my audit missed it:** I grepped for `c - a) >= 1.0`, and this site spells the same predicate with different variable names — `(c - ea) >= 1.0`. Re-swept with a pattern that cannot care about the names (`\(c - [a-z_]+\)|>= 1\.0`); this was the only remaining site, and the two surviving `>= 1` strings are historical prose explaining why the old bound existed.

### The stale comments were not only cosmetic

You asked for the stale GL comments to go. One of them encoded an assumption the relaxation had silently broken:

> "… and `_euler_labeling` guarantees c-B >= 1, so B is the binding one."

Under `c - B > 0` the **t=1 exponent can now be the small one**, which it never could before — so that sentence no longer justifies the fixed half-width. Measured rather than assumed: `c - B` down to **1e-4 is accurate to 2.3e-15**, because the tanh–sinh rule is *symmetric in u* — a half-width that resolves a small `B` resolves a small `c-B` identically. The comment now states that reason, and `test_hyp2f1_small_c_minus_B_is_accurate` pins it.

The module header's description of the two Gauss–Legendre substitutions, and the `xi^k` / cap-at-12 / 0.25-threshold block, are deleted outright — that machinery no longer exists after #1402. The remaining GL mentions are deliberate rationale ("the old requirement was…", "This REPLACED…").

### Tests

* `test_hyp2f1_traced_label_matches_the_concrete_rule` — the four cases above, asserting the traced arm is finite *and* matches scipy to 1e-13. The parameters all have `0 < c-a < 1` **and** `b <= 0`, which is what makes the mismatch reachable: the first label is the only admissible one, so wrongly rejecting it is fatal rather than merely suboptimal.
* `test_hyp2f1_small_c_minus_B_is_accurate` — guards the symmetry argument above.

### Gates

* **numpy byte-identical**: sha256 `245bc0e3…` over 315 values matches develop `8c320b39` exactly, arms grep-asserted to differ (traced selector 1 → 0).
* **391 passed, 1 skipped** across `test_backend_special` / `_constantbetadf` / `_quadrature`.
* Ledger delta: **0**.

### DIFF
--- a/galpy/backend/special/_fallback/hyp2f1.py
+++ b/galpy/backend/special/_fallback/hyp2f1.py
@@ -11,15 +11,12 @@
 #   Method: Euler's integral representation, valid for all z not in [1, inf):
 #       2F1(a,b;c;z) = Gamma(c)/(Gamma(B) Gamma(c-B))
 #                      * int_0^1 t^{B-1} (1-t)^{c-B-1} (1 - z t)^{-A} dt
-#   where {A, B} = {a, b} is chosen so B > 0 and c - B > 0 (preferring
-#   c - B >= 1 so the t=1 endpoint is non-singular). For z <= 0 the factor
-#   (1 - z t) = (1 + |z| t) >= 1 is smooth, but for large |z| it forms a thin
-#   boundary layer near t=0; two substitutions resolve it for fixed-order
-#   Gauss-Legendre quadrature:
-#     1. t = ((1+|z|)^X - 1)/|z| maps the boundary layer to a uniform X-grid;
-#     2. X = xi^k (k = ceil(1/B) when B < 1) regularizes the t^{B-1} endpoint
-#        singularity so plain Gauss-Legendre converges.
-#   Pure arithmetic + log1p/expm1/pow, so it differentiates under jax and torch.
+#   where {A, B} = {a, b} is chosen so B > 0 and c - B > 0 -- just what the Beta
+#   integral needs to converge. The quadrature is a tanh-sinh (double-
+#   exponential) rule that cancels BOTH endpoint singularities analytically
+#   against its own weight, so neither exponent has to be regularized by a
+#   substitution and neither endpoint has to be preferred over the other.
+#   Pure arithmetic + log1p/pow, so it differentiates under jax and torch.
 ###############################################################################
 import math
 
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Additional evidence for a risk this fix carries that the reported cases don't cover.

Changing `oka` doesn't only rescue the broken cases. Where **both** labels are admissible under the new rule, the traced path now picks label `a` where it previously picked `b` — so cases that already **worked** change route. That is the class worth measuring, and none of the four reported repros are in it (they all have `b <= 0`, i.e. only one admissible label).

Constructed the flip class explicitly: `0 < c-a < 1` (so `a` was rejected by the old traced rule and is accepted by the new one) **and** `b > 0` with `c-b >= 1` (so the old fallback to `b` was valid and worked). A/B through the traced entry, 60 cases, vs mpmath (dps=40):

| tree | worst rel err |
|---|---|
| develop `8c320b39` (label b) | 2.220e-15 |
| this PR (label a) | 2.887e-15 |

Both at the noise floor, so the flip is safe — the two labellings are equally accurate here, which is what you'd expect since the tanh–sinh rule cancels either endpoint analytically. No regression, and the previously-broken cases go inf/nan → ~1e-15.

Worth stating the limitation: this samples `a` in {1.6, 2.0, 3.0} × `c-a` in {0.9, 0.5, 0.2} × `b` in {0.5, 1.0, 2.0} × z in {−0.6, −3, −40}, not the whole space.

PR #1404, opened 2026-08-27
actionAngleStaeckel: chi-anomaly quadrature at the turning points in C

jobovy-agent:
The C action, frequency and angle integrals used a `t^2` substitution around
each turning point. That removes the sqrt branch point but not the reason the
integrand is hard there: `S = p^2/(2 delta^2)` is a difference of O(1) terms
cancelling to zero, so its *relative* error blows up while its absolute error
stays at machine epsilon, and the `1/sqrt(S)` of the frequency and angle
integrands amplifies it.

Worse, `t^2` clusters Gauss-Legendre nodes as `1/order^4` onto the turning
point, so **raising the order moved them further into the cancellation and made
the answer worse**:

| | order 200 | order 800 |
|---|---|---|
| `dOmega_z/Omega_z`, C vs pure-Python | 1.8e-9 | **3.3e-8** |

against ~1e-12 for the pure-Python path at both orders. Degrading with order is
the signature of cancellation rather than truncation, and it is why this could
not be fixed by asking for more quadrature points.

### What this does

Uses the chi anomaly of the Python path at the turning points,
`q = qmin + D sin^2(chi/2)`. With `y = sin^2(chi/2)` and `Q = S/[y(1-y)]`,

    int sqrt(S) dq   = (D/4) int sqrt(Q) sin^2(chi) dchi
    int f/sqrt(S) dq = D int f/sqrt(Q) dchi

both regular at the turning point. Near the endpoints `Q` is taken from the
analytic derivative by the trapezoid `S ~ (q-q0)[S'(q0)+S'(q)]/2` instead of
from the cancelling difference -- the same switch, at the same threshold, as
`_staeckelChiQuadratures`.

The midplane half of the `v` integrals keeps its `t^2` form: the midplane is an
interior symmetry point of `S_z`, not a turning point. The `umin`/`umax`/`vmin`
root functions also keep the direct expression, since the turning points they
solve for are not known while they run.

### Measured

Agreement with the pure-Python path near the radial turning points, on an
integrated orbit:

| potential | order 200 | order 800 |
|---|---|---|
| KuzminKutuzov | 6.0e-07 -> **1.5e-09** | 2.0e-06 -> **1.4e-09** |
| MWPotential2014 | 5.9e-08 -> 1.9e-07 | 2.1e-06 -> **9.2e-08** |

Frequencies improve 10-11x for MWPotential2014 (1.6e-4 -> 1.4e-5). The
substantive change is that the result is now **stable in the quadrature order**
rather than degrading with it.

Honest caveat: MWPotential2014 at order 200 moves the wrong way (5.9e-8 ->
1.9e-7). That is a max over a small near-turning-point subsample on a potential
that is not of Staeckel form; the value is stable at ~1e-7 across orders instead
of growing to 2e-6 as before.

All 73 Staeckel tests pass, including the near-circular `interpRZPotential`
cases (`test_actionAngleStaeckel_basic_actions_u0_interppot_c`,
`test_actionAngleStaeckel_basic_freqs_u0`), which an earlier threshold-based
attempt broke -- this needs no tuned constant.

### How it was found

While checking how close `actionAngleStaeckelInverse`'s canonical map gets to
machine precision. It reproduces the exact direct construction to 3e-13, but
measured only ~1e-7 against the forward code -- and the *direct* construction,
which has no canonical machinery at all, showed the same 4.4e-8. The floor was
the forward C path, not the inverse.
---

**Scope note (from review).** The `v` path is deliberately not the literal
Python rule: Python integrates the whole `vmin -> pi/2` interval in chi, while
this converts only the turning-point half and leaves the midplane half on the
old `t^2` panel, since the midplane is an interior symmetry point of `S_z`
rather than a turning point and has nothing to regularize. The chi
parametrization is applied where the cancellation actually is.

### DIFF
--- a/galpy/actionAngle/actionAngle_c_ext/actionAngleStaeckel.c
+++ b/galpy/actionAngle/actionAngle_c_ext/actionAngleStaeckel.c
@@ -161,20 +161,16 @@ void calcVmin(int,double *,double *,double *,double *,double *,double *,int,
 	      double *,double *,double *,double *,double *,int,
 	      struct potentialArg *);
 double JRStaeckelIntegrandSquared(double,void *);
-double JRStaeckelIntegrand(double,void *);
 double JRLowStaeckelIntegrand(double,void *);
 double JRHighStaeckelIntegrand(double,void *);
 double JzStaeckelIntegrandSquared(double,void *);
 double JzStaeckelIntegrand(double,void *);
 double JzLowStaeckelIntegrand(double,void *);
 double JzHighStaeckelIntegrand(double,void *);
-double dJRdEStaeckelIntegrand(double,void *);
 double dJRdELowStaeckelIntegrand(double,void *);
 double dJRdEHighStaeckelIntegrand(double,void *);
-double dJRdLzStaeckelIntegrand(double,void *);
 double dJRdLzLowStaeckelIntegrand(double,void *);
 double dJRdLzHighStaeckelIntegrand(double,void *);
-double dJRdI3StaeckelIntegrand(double,void *);
 double dJRdI3LowStaeckelIntegrand(double,void *);
 double dJRdI3HighStaeckelIntegrand(double,void *);
 double dJzdEStaeckelIntegrand(double,void *);
@@ -1973,12 +1969,6 @@ static inline double chiQvStaeckel(double chi,double vmin,double E,double I3V,
       / 2. / ( 1. - y );
   return Q;
 }
-double JRStaeckelIntegrand(double u,
-			   void * p){
-  double out= JRStaeckelIntegrandSquared(u,p);
-  if ( out <= 0.) return 0.;
-  else return sqrt(out);
-}
 double JRStaeckelIntegrandSquared(double u,
 				  void * p){
   struct JRStaeckelArg * params= (struct JRStaeckelArg *) p;
@@ -2010,16 +2000,6 @@ double JRHighStaeckelIntegrand(double chi,
   double D= params->umax - params->umin;
   if ( Q <= 0. ) return 0.;
   return D / 4. * sqrt(Q) * sc * sc;
-}
-double JRStaeckelIntegrandSquared4dJR(double u,
-				      void * p){
-  struct dJRStaeckelArg * params= (struct dJRStaeckelArg *) p;
-  double sinh2u= sinh(u) * sinh(u);
-  double dU= (sinh2u+params->sin2v0)
-    *evaluatePotentialsUV(u,params->v0,params->delta,
-			  params->nargs,params->actionAngleArgs)
-    - (params->sinh2u0+params->sin2v0)*params->potu0v0;
-  return params->E * sinh2u - params->I3U - dU  - params->Lz22delta / sinh2u;
 }
 
 double JzStaeckelIntegrand(double v,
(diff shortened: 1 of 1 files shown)


jobovy (commented):
Findings:

- Low: `galpy/actionAngle/actionAngle_c_ext/actionAngleStaeckel.c:627` / `:1160` declare `midchi` outside OpenMP loops, but the `private(...)` clauses at `:644` and `:1177` omit it while assigning it inside the loop. The value is constant, so this is unlikely to change results, but it is still a shared write/data race under OpenMP. Make it private or compute it once outside the parallel region.
- Nit: `midchi` is also declared but unused in `calcJRStaeckel` / `calcdJRStaeckel`, and the new derivative integrands declare unused `sc = sin(chi)` locals around `:2067`, `:2078`, `:2095`, etc. Not a behavior issue, but avoidable C hygiene.

Correctness notes:

The math looks right to me. The `D/4 * sqrt(Q) * sin^2(chi)` action integrands and `D * f/sqrt(Q)` derivative integrands match the Python `_staeckelChiQuadratures` formulas. The analytic endpoint derivative reconstruction also matches the Python implementation, with the radial high-side sign adjusted appropriately for the reversed `umax - D*y` mapping.

The C vertical path is not literally the same quadrature as Python: Python integrates the full `vmin -> pi/2` interval in chi, while this C patch converts only the turning-point half and keeps the midplane half as the old `t^2` panel. That split still covers the right interval and appears appropriate, but the PR wording "matching the Python path" is a little stronger than the implementation.

Checks run:

- Built the C extension in a temporary PR worktree.
- Ran `pytest -q tests/test_actionAngle.py::test_actionAngleStaeckel_actions_order_c tests/test_actionAngle.py::test_actionAngleStaeckel_freqs_order_c`: passed.
- Spot-checked C vs pure-Python actions/frequencies/angles for MWPotential2014 and Kuzmin-Kutuzov cases; agreement was around `1e-11` or better at order 50.

jobovy-agent:
Thanks -- all three points were right, and the first was a real bug. Fixed in
the push just now.

**The `midchi` data race.** Correct, and worse than the "unlikely to change
results" framing: it was declared outside the OpenMP loops in
`calcJzStaeckel` and `calcdJzStaeckel`, assigned inside them, and absent from
both `private()` clauses. Every thread happened to write the same value, so it
would not have shown up as wrong answers, but it is a shared write either way.

Rather than adding it to `private()`, it turns out not to need to be a
variable at all: the midpoint of `[vmin, pi/2]` sits at `y = 1/4`, so
`chi = 2 asin(1/2) = pi/3` exactly. It is now `STAECKEL_CHI_MID`, which
removes the race and the two unused declarations in the `u` routines in one
move. The macros moved to the top of the file next to `M_PI`, since the `v`
routines that use them appear before the chi helpers that had defined them.

**Unused `sc` locals.** Also correct -- an artifact of generating the twelve
integrand bodies from one template. Dropped from the nine derivative
integrands; only the three action integrands need `sin^2(chi)`.

**"Matching the Python path" is too strong.** Agreed, and this is the most
useful of the three, because it is the sort of thing that misleads a later
reader. Python integrates the whole `vmin -> pi/2` interval in chi; this patch
converts only the turning-point half and leaves the midplane half on the old
`t^2` panel, deliberately, because the midplane is an interior symmetry point
of `S_z` rather than a turning point and so has nothing to regularize. The
title and body have been reworded to say the quadrature is chi-anomaly *where
it matters*, rather than implying the two paths are now literally the same
rule.

Only behaviour change is a last-ulp difference between `2 * asin(0.5)` and
`M_PI / 3`. The KuzminKutuzov near-turning-point agreement with the Python
path is unchanged at 1.5e-09 (order 200) and 1.4e-09 (order 800), and the 73
Staeckel tests pass. I cancelled the superseded CI runs before pushing.

PR #1407, opened 2026-08-28
surfdens: eager backends take the batched vertical quadrature (ledger -6)

jobovy-agent:
Both vertical surface-density integrals sent eager jax/torch to
`scipy.integrate.quad`, whose adaptive loop calls the integrand one **scalar**
node at a time — 1113 calls at ~48 ms each for a wrapped MWPotential2014. The
batched Gauss–Legendre rule the **traced** path already uses evaluates it once on
the whole node array.

Measured against scipy on eager jax, 3 wrappers × 18 (R, |z|, φ), scipy
converged to 1.4e-16:

| potential | worst rel err | scipy | GL | |
|---|---|---|---|---|
| `mockOffsetMWP14Wrapper` | 1.6e-14 | 147 s | 7 s | 22× |
| `mockRotatedAndTiltedMWP14` | 4.2e-15 | 171 s | 2 s | 83× |
| `mockRotatedTiltedOffsetMWP14` | 8.4e-16 | 140 s | 2 s | 70× |

Machine precision — not a precision trade, and it is the same rule at the same
order the traced path has always shipped.

## Ledger −6, measured end to end

Whole `tests/test_potential.py` under `--backend jax`, list emptied for these:

    before   4 failed, 1284 passed, 102 skipped   7144 s (1:59:04)
    after    0 failed, 1288 passed, 102 skipped   2733 s (0:45:33)

| test | before | after | CI ≈ 0.71× |
|---|---|---|---|
| `poisson_surfdens[mockRotatedAndTiltedMWP14]` | >900 s **FAIL** | 20.5 s | 14.5 s |
| `poisson_surfdens[mockRotatedTiltedOffset…]` | >900 s **FAIL** | 19.1 s | 13.6 s |
| `poisson_surfdens[mockOffsetMWP14Wrapper]` | >900 s **FAIL** | 17.1 s | 12.1 s |
| `poisson_surfdens[…MWP14wInclination]` | >900 s **FAIL** | 16.5 s | 11.7 s |
| `poisson_surfdens[…TriaxialLogHalowIncl]` | 349.8 s | 8.0 s | 5.7 s |
| `test_DehnenBinney98` | 249.8 s | 77.9 s | 55.3 s |

All six clear the 300 s cap with >80% margin, so all six come off the list. The
file is now **faster with them running** than it used to be with them skipped —
every other surfdens caller sped up too.

## The gate is not `_force_accepts_arrays`

Using it did nothing, and the smoke test caught that (`symmetric_quad calls=0`):
`_surfdens_poisson` runs on the inner `RotateAndTiltWrapperPotential`, whose
force methods carry the scalar-only marker. That marker is right for the **numpy**
contract — but the potential also declares `_backend_accepts_arrays = True`,
which is exactly the escape hatch `check_potential_inputs_not_arrays` implements:
it admits an array when the potential opts in **and** every array argument is a
backend array, which is precisely this path. The gate now asks the question the
guard itself asks.

Still on scipy: potentials marked scalar-only that do **not** opt in — e.g.
`AnySphericalPotential`, whose force closes over `scipy.integrate.quad` with a
scalar upper limit, so the node array would silently collapse. Pinned by a test.

The density route is rerouted too, and asks about `_dens` rather than the force
methods it never calls — hence `_accepts_node_array(pot, *methods)`, the marker
scan `_force_accepts_arrays` already had, factored out rather than duplicated.

**numpy is byte-identical**: the gate returns `False` for numpy on its first
line, and a test pins that numpy is never diverted off scipy.

## Gates

| | |
|---|---|
| numpy `test_potential.py` | 1288 passed, 102 skipped |
| jax `test_potential.py` | 1288 passed, 102 skipped, **0 failed** |
| jax+torch `test_backend_quadrature.py` | 103 passed, 1 skipped |
| A/B | the 6 new route tests **fail on base** and pass here; the 2 invariant tests (scalar-only stays on scipy, numpy never diverted) pass on **both** sides, which is what they are for |

Tests assert the **route**, not only the value — a value check cannot tell
"took the batched rule and got the right answer" from "silently fell back to
scipy".

Audit: adds no data-dependent branch (the gate reads Python attributes and
namespace identity, both static — no `if`→`xp.where` class); removes duplication
rather than adding it; helpers are potential-specific and sit beside the ones
they generalise, so `Potential.py` is the right home, not `galpy.backend`.

Ledger delta: **−6** (slow-skip 57 → 51; total 65 → 59).

### DIFF
--- a/galpy/potential/Potential.py
+++ b/galpy/potential/Potential.py
@@ -161,6 +161,52 @@ def _quad_needs_backend(xp, *coords):
     return any(under_trace(c) or requires_backend_grad(c) for c in coords)
 
 
+def _vertical_quad_needs_backend(
+    xp, pot, *coords, methods=("_Rforce", "_zforce", "_rforce")
+):
+    """Whether the vertical surface-density integrals should use the backend rule.
+
+    Under a trace, always -- scipy cannot be evaluated there at all, which is
+    what `_quad_needs_backend` decides. Outside one, whenever the namespace is
+    not numpy AND ``pot``'s forces broadcast: scipy's adaptive loop calls the
+    integrand one scalar node at a time (~1100 calls for a wrapped
+    MWPotential2014, ~48 ms each on eager jax), while the backend rule evaluates
+    it ONCE on the whole node array. Measured against scipy on eager jax, over 3
+    wrapper potentials x 18 (R, |z|, phi):
+
+        mockOffsetMWP14Wrapper          worst 1.6e-14   147 s -> 7 s
+        mockRotatedAndTiltedMWP14       worst 4.2e-15   171 s -> 2 s
+        mockRotatedTiltedOffsetMWP14    worst 8.4e-16   140 s -> 2 s
+
+    so it is 22-83x faster at the same accuracy, not a precision trade.
+
+    Array-safety is asked in the terms `check_potential_inputs_not_arrays` itself
+    uses. That guard rejects an array UNLESS the potential opts in with
+    ``_backend_accepts_arrays`` and every array argument is a backend array --
+    which is exactly the situation here, since ``xp`` is not numpy. So a
+    potential carrying the opt-in is safe to batch even though its methods are
+    marked scalar-only for the numpy contract, and asking only
+    `_accepts_node_array` would wrongly exclude it: RotateAndTiltWrapper
+    decorates all of its force methods and is the wrapper under every one of the
+    measurements above.
+
+    Potentials with neither the opt-in nor a scalar-only marker (the common case,
+    and every plain analytic potential) broadcast anyway. What is left out is a
+    potential that is marked scalar-only AND does not opt in -- e.g.
+    AnySphericalPotential, whose force closes over ``scipy.integrate.quad`` with
+    a scalar limit -- which keeps scipy, as it must.
+
+    numpy always keeps scipy, so the numpy path is byte-identical.
+    """
+    if xp is numpy:
+        return False
+    return (
+        _quad_needs_backend(xp, *coords)
+        or getattr(pot, "_backend_accepts_arrays", False)
+        or _accepts_node_array(pot, *methods)
+    )
+
+
 def _force_accepts_arrays(pot):
     """Whether ``pot``'s force methods broadcast over an array of positions.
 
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Torch arm measured locally as well, so the gate now covers all three backends
on the whole file. `tests/test_potential.py` has **zero** entries in either
ledger for torch, so that arm runs the file unfiltered — it was the real
regression surface for this change and it is clean:

| backend | result | wall |
|---|---|---|
| numpy | 1288 passed, 102 skipped | 14:24 |
| jax | 1288 passed, 102 skipped, **0 failed** | 45:33 |
| torch | 1288 passed, 102 skipped, **0 failed** | 17:02 |

Identical counts on all three. The same tests under torch:

| test | jax | torch |
|---|---|---|
| `poisson_surfdens[mockRotatedAndTiltedMWP14]` | 20.5 s | 3.2 s |
| `poisson_surfdens[mockRotatedTiltedOffset…]` | 19.1 s | 3.2 s |
| `poisson_surfdens[mockOffsetMWP14Wrapper]` | 17.1 s | 3.2 s |
| `poisson_surfdens[…MWP14wInclination]` | 16.5 s | 3.2 s |
| `poisson_surfdens[…TriaxialLogHalowIncl]` | 8.0 s | 1.2 s |
| `test_DehnenBinney98` | 77.9 s | 44.3 s |

One more check worth stating explicitly: the `-jit` arms are **unaffected**.
The gate still evaluates `_quad_needs_backend` first, which is True under a
trace, so the traced route is byte-identical to before. The only two `-jit`
ledger entries touching this test are
`test_poisson_surfdens_potential[AnyAxisymmetricRazorThinDiskPotential]`
(xfail, jax-jit and torch-jit) and they are for a different potential and a
different reason, so they stay as they are.

Ledger delta is therefore exactly **−6**, all in the `jax` arm (30 → 24):
slow-skip 57 → 51, plus 8 xfail = **59** total.

jobovy-agent:
Verification matrix complete — all five arms of the whole `tests/test_potential.py`,
run locally on this branch's committed content.

| arm | result | wall |
|---|---|---|
| numpy | 1288 passed, 102 skipped | 14:24 |
| jax eager | 1288 passed, 102 skipped, **0 failed** | 45:33 |
| torch eager | 1288 passed, 102 skipped, **0 failed** | 17:02 |
| jax `--jit` | 1285 passed, 102 skipped, 2 xfailed, 1 failed¹ | 60:49 |
| torch `--jit` | 1286 passed, 102 skipped, 1 xfailed, 1 failed¹ | 63:38 |

¹ The single traced failure is `test_pickling_potential_unitful_dens`, and it is
**not this PR's**. A/B, same command, same content otherwise:

    this branch (with the Potential.py change)   FAILED
    develop + comments-only (without it)         FAILED   <- identical

It is a pre-existing units-are-not-traceable failure on **both** traced arms
(`TracerArrayConversionError` / the torch equivalent), and it passes on both
eager arms — exactly the behaviour `backend_xfail.txt`'s own comment describes
for a units-based density reaching `as_numpy()`. Separately filed as a ledger
correction to make; it is unrelated to this change.

The traced arms matter here because I claimed in the description that the traced
route is byte-identical by construction (`_quad_needs_backend` is evaluated
first and is True under a trace). That is now measured, not just argued — and
since CI never runs jit mode, local is the only place it can be.

The xfail counts also came out as predicted: the AnyAxisymmetricRazorThinDisk
traced entries are untouched by this change.

PR #1412, opened 2026-08-29
actionAngleVerticalInverse: the momentum-matched anomaly map (construction)

jobovy-agent:
First step of bringing the one-dimensional case onto the construction the
spherical and Staeckel cases use. This adds the **map construction only** —
it is deliberately not wired into evaluation yet, for a reason given at the
bottom.

### What it does

`_momentum_matched_map` builds the map onto an equal-action harmonic
auxiliary torus by matching cumulative actions, returning the sine
coefficients and the storage variable `K = xmax^2 / J`.

Three properties were established numerically before writing it, and the
test pins each:

- **The map is a pure sine series.** On an isothermal disc the constant term
  measures ~1e-13 and the largest cosine coefficient ~1e-12, which is the
  parity argument holding to round-off. For a potential symmetric about the
  midplane only even harmonics survive, halving the coefficients.
- **The harmonic frequency cancels.** `A^A(eta) = J (eta - sin eta cos eta)`,
  since `omega` cancels against the squared amplitude `2J/omega`, so the map
  depends on the auxiliary only through its action. Verified at 5e-13.
- **Fitting beats inverting.** `dA^A/deta` vanishes like `sin^2` at both
  turning points, so a pointwise inverse of the matching condition is
  ill-conditioned exactly where the anomaly was introduced to be regular:
  measured, it converges as `1/nta` and floors the construction at 3e-4
  *regardless of how many harmonics are kept* — the signature of an error
  that is not truncation. Fitting the coefficients leaves that derivative as
  a factor and reaches **3e-16** at E=0.5 and **1e-14** at E=2.0.

The cumulative action uses a spectral antiderivative; a cumulative trapezoid
is second order and floors the fit at ~4e-9.

`K = xmax^2/J` tends to a finite constant as `J -> 0` rather than vanishing.
On held-out nodes of an energy grid a cubic spline reproduces `xmax` to
2.5e-2 relative when `xmax` is stored and to 1.4e-5 when `K` is — a factor
of 1800.

### Why evaluation is not wired up here

I tried the cheap route and measured it failing, which is worth recording.
The existing `"exact"` point transformation already uses an equal-action
harmonic auxiliary and stores its map as a sampled mesh, so momentum
matching looked like a two-expression change to *which* quantity is matched
— time (`int dx/p`) versus action (`int p dx`). That change builds and all
49 vertical tests pass.

It is also wrong: actions degrade to `dJ ~ 1e-1` against the exact PT's
3e-12. The mesh is not a neutral representation — the downstream angle
machinery *assumes* time matching, because for a harmonic auxiliary the
angle advances linearly in time, so time matching makes the two sets of
angles correspond directly with no compensation at all. Momentum matching
makes the *action* correspond instead. In the Staeckel and spherical cases
that gap is exactly what the compensation term absorbs; the 1D code has no
compensation because it never needed one.

So completing the 1D case means porting the compensation, not exchanging one
stored map for another. That is the next PR; this one is the piece it needs,
self-contained and tested.

49 vertical tests pass.

### DIFF
--- a/galpy/actionAngle/actionAngleVerticalInverse.py
+++ b/galpy/actionAngle/actionAngleVerticalInverse.py
@@ -394,6 +394,77 @@ class actionAngleVerticalInverse(actionAngleInverse):
             self._pt_deriv2_coeffs[ii] = polynomial.polyder(self._pt_coeffs[ii], m=2)
         return None
 
+    def _momentum_matched_map(self, ii, npt=16, nta=1024):
+        """
+        Momentum-matched anomaly map of torus ii.
+
+        The auxiliary torus is the harmonic oscillator with the SAME action,
+        and corresponding points are those that have swept the same action:
+        A^A(eta) = A(tau).  Because the two actions agree, eta - tau is
+        periodic, and because the momentum is antisymmetric about the turning
+        points it is odd under tau -> 2 pi - tau, so the map is a pure sine
+        series.  For a potential symmetric about the midplane only the even
+        harmonics survive.
+
+        The auxiliary's cumulative action is A^A(eta) = J (eta - sin eta
+        cos eta): the harmonic frequency cancels between omega and the
+        squared amplitude 2J/omega, so the map does not depend on which
+        harmonic auxiliary is used, only on its action.
+
+        The coefficients are obtained by FITTING them to the matching
+        condition rather than by inverting A^A pointwise.  That inverse is
+        ill-conditioned at both turning points, where dA^A/deta vanishes like
+        sin^2, and a pointwise construction converges only as 1/nta; fitting
+        leaves the vanishing derivative as a factor rather than a divisor.
+
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Closing as superseded by #1415 (merged today as 70032808, stack #1518): the momentum-matched anomaly map was rebuilt fresh off main there, Jo's direction at the time since a stack could not be merged into main. The construction is the same in kind, fitting eta(tau) = tau + sum_m D_m sin(m tau) to the matching condition A^A(eta) = A(tau) rather than inverting it pointwise, with K = xmax^2/J as the storage variable. What landed differs in the fit itself (analytic Jacobian, warm start from the neighbouring torus, #1515), in the samples (spectral antiderivative for A, action and frequency from the same sweep), and in the resolution (forty harmonics by default, with a convergence warning). See #1414's closing note for the family side, fast-orbits #26 for the derivation.

PR #1414, opened 2026-08-29
actionAngleVerticalInverse: the momentum-matched family

jobovy-agent:
Stacked on #1412.

The anomaly map of #1412 describes a single torus. This stores it across the energy grid together with the amplitude, which together are the entire stored content of the canonical map in the new scheme — there is no table of angle-fit coefficients, because the auxiliary torus carries the target's content through the anomaly map alone.

**Why `K = xmax^2 / J` and not `xmax`.** The action vanishes at the bottom of the grid and `xmax` vanishes with it like `sqrt(J)`, so a table of `xmax` carries a square-root cusp there and the interpolant spends its resolution resolving it. `K` instead tends to `2 / omega`, finite and O(1) across the whole grid. At `J = 0` the torus is its own auxiliary, so the anomaly map is the identity and `K` takes that limit exactly.

**Derivatives.** `_mm_tables` differentiates the same interpolants the evaluation reads, chained through `E(j)`. Nothing is finite-differenced and no derivative table is stored separately — that is what keeps the map symplectic for arbitrary stored content rather than only for converged tables. The test checks the chain against finite differences *of the interpolant*, which is the precise statement being claimed.

Tests: 50 vertical-inverse tests pass; the new code is fully covered with no pragmas.

### DIFF
--- a/galpy/actionAngle/actionAngleVerticalInverse.py
+++ b/galpy/actionAngle/actionAngleVerticalInverse.py
@@ -465,6 +465,93 @@ class actionAngleVerticalInverse(actionAngleInverse):
         )
         return sol.x, xmax**2.0 / J
 
+    def _setup_momentum_matched_family(self, npt=16, nta=1024):
+        """
+        Build the momentum-matched family: the anomaly map and the storage
+        variable K on every torus of the energy grid.
+
+        This is the whole stored content of the canonical map in the new
+        scheme.  There is no table of angle-fit coefficients: the auxiliary
+        torus carries the target's content through the anomaly map alone,
+        and the amplitude enters only through K = xmax^2 / J.
+
+        K is stored rather than xmax because xmax vanishes with the action
+        while K does not: in the harmonic limit xmax^2 -> 2 J / omega, so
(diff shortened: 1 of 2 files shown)


jobovy-agent:
Closing as superseded: this family construction, together with #1412's anomaly map, was rebuilt fresh off main as #1415 (merged today as 70032808 in stack #1518, with #1515 and #1517 on top, 058e3d2a and 2db0c6f7). What changed between this draft and what landed: the family tables are Hermite splines in the action, with each torus's own analytic dD_m/dJ and dK/dJ as constraints (#1517), rather than prefiltered B-spline rows on the energy grid (whose mirror boundary spoiled the end intervals); the zero-energy node carries sqrt(Phi''(0)) from #1519's second derivatives; the fit uses an analytic Jacobian with a warm start and the angle relation is inverted by Newton on its closed-form derivative (#1515); forty harmonics by default, mm_nta as a parameter with the Nyquist check, a warning when the truncated map does not reconstruct a torus's momentum; and the tests go through the public interface only. The derivation is in fast-orbits #26 (canonical.tex, the 1D worked example) and the comparison with the older evaluations in fast-orbits #29.

PR #1415, opened 2026-08-30
actionAngleVerticalInverse: the momentum-matched canonical map

jobovy-agent:
**Update (2026-09-18):** the review threads on this PR are addressed at the top of the stack (#1517, section "Review fixes"), where the zero-energy frequency now comes from #1519's `evaluatelinearx2derivs` (the five-point stencil mentioned below is gone), the default is forty harmonics, and the tests use only the public interface. The numbers below are this PR's own at the time; the current ones are in #1517.

A single PR to `main`, built fresh off `main` rather than stacked, per your note that stack-merging into `main` is not available to me.

One degree of freedom is the case in which every step of the canonical construction is elementary, so this implements it end to end — from a requested action and angle to a point — with nothing finite-differenced and no derivative tabulated apart from the interpolants the evaluation already reads. It is now the **default** evaluation, so the inverse methods are consistent with one another; the older one remains reachable.

### The construction

The auxiliary is the harmonic oscillator of the **same** action, and its frequency cancels between the momentum and the amplitude in `p^A dx^A/deta = 2 J sin^2(eta)`. The matching condition `A^A(eta) = A(tau)` is therefore independent of *which* harmonic auxiliary is used — only its action matters.

That condition is **fitted, not inverted pointwise**. The derivative being inverted vanishes at both turning points, so a pointwise construction floors the map near `3e-4` however many harmonics are kept — a conditioning problem, not a resolution one — while fitting leaves the vanishing derivative as a factor and converges spectrally.

Everything else is then closed form: the momentum from the flux identity; the compensation for the moving turning point grouped as `p (dxmax/dJ) cos(tau)`, finite where the factored product returns `nan`; and the angle as the action derivative of the generating function the matching condition makes explicit, `W = J (eta - sin eta cos eta)`. Inverting the angle is a bracketed bisection, guaranteed because the relation is monotone and advances by exactly `2 pi`.

`K = xmax^2 / J` is stored rather than `xmax`, which vanishes like `sqrt(J)` at the bottom of the grid while `K` tends to `2/omega` — a cubic spline reconstructs `xmax` 1800x better through `K`. At `J = 0` the torus is its own auxiliary, so the map is the identity and `K` takes that limit exactly.

`E(J)` is a **Hermite** interpolant through the energies *and* their slopes, because `dE/dJ` is the frequency and is already known exactly on every grid torus. Fitting `E` alone and differentiating discards that and leaves the map disagreeing with the tabulated frequency by `4e-5` on a nine-node grid; the Hermite form makes it exact at the nodes, and `1.3e-11` against the forward transformation independently of the grid.

### Measured, against the forward transformation

Independent of everything above, so it is a yardstick rather than a self-consistency check. Through the public entry point:

| quantity | 9-node grid | 33-node grid |
|---|---|---|
| action | 3.9e-13 | 3.9e-13 |
| angle | 1.3e-3 | 8.1e-9 |
| frequency | 1.3e-11 | 1.3e-11 |

The pattern is the claim, and the tests assert it rather than three thresholds: the action is preserved **whatever the tables contain**, so it needs no fine grid; the angle reads `dD_m/dJ` from the family and converges with it; the frequency is exact at the nodes by construction and so does not depend on the grid at all.

Reconstruction returns a grid node to its own torus to `6e-11` with 28 harmonics and `5e-6` with 12, converging spectrally — the residual is the truncation of the anomaly map, not a defect.

### When the default does not apply

Both documented in the constructor. A family interpolates a cubic spline in the action, so it needs **at least four energies**; with fewer, the older evaluation is the only one available and the default falls back to it rather than refusing to construct (asking for the family explicitly still raises). And an explicit `use_pointtransform` selects the older evaluation, since the momentum-matched map **is** a point transformation and the two cannot both be in force.

Flipping the default initially broke 27 vertical tests, all on the four-energy requirement and all fixed by that fallback; the 28th was the point-transformation test, which the exclusion covers.

### One thing left for you

`Freqs` now answers **between** grid tori, where it used to raise — the canonical map evaluates there, so it has a frequency there. That fills only the case that raised: on a node the tabulated frequency still answers and `momentum_matched=False` is unchanged.

Whether `Freqs` should prefer the map’s frequency where **both** are available is a real trade I have deliberately not taken: the map differentiates an interpolated energy while the table interpolates the frequency directly, which is one order better off the nodes. Routing it costs ~4e-10 there and trips `test_actionAngleVerticalInverse_freqs_wrtVertical`; I did not want to relax a pre-existing tolerance to fit my own change. The two now agree to 1.3e-11 rather than 1e-5, so the inconsistency is six orders smaller than when I first hit it.

### Notes

- 8 new tests; the new code is at 100% coverage with **no pragmas** — including the stencil’s top-of-grid branch, which is reachable when the requested action sits on the last node and is now tested rather than excluded.
- 51 vertical-inverse tests pass in total.
- The math is written up in fast-orbits #26, where the one-dimensional case is now the first worked example of the general construction.
- Supersedes the momentum-matched parts of #1412/#1414; those and #1384/#1385 can be closed if you prefer this route.

### Review round (2026-09-16): between the grid tori

Reviewing this PR with Jo turned up that every measurement above is at a grid node, and measuring **between** the nodes exposed two defects confined to the end intervals of the family (the interior was at 1e-5 on the same nine-node grid):

- the tables were prefiltered B-splines on the energy-grid row with a **mirror boundary**, which imposes a zero slope at both ends of the grid — but D grows linearly out of the harmonic bottom, so the first and last intervals were off by ~1e-2 in the action, converging only at first order;
- the **zero-energy node's frequency was a 2018 placeholder** copied from the first torus ("should use sqrt(2nd deriv. pot)"), off by O(J_1) — 13 percent on nine nodes — which the family then interpolated through.

D and K are now not-a-knot cubic splines in the action itself, differentiated by the same spline objects the evaluation reads (canonicity stays manifest). That also removes the hand-rolled stencil and the row coordinate, and with it a silent assumption that the energies are equally spaced: a non-uniform grid of four or more energies previously gave wrong answers even at its own nodes (2.6e-2), where the old default never needed uniform spacing. The zero-energy node now carries the midplane's harmonic frequency from a five-point stencil of the force (~1e-10), for both evaluations.

Measured between grid tori, max relative action error, uniform / non-uniform grids:

| nodes | before | after |
|---|---|---|
| 9 | 1.3e-2 / 1.1e-2 | 2.5e-5 / 3.0e-5 |
| 17 | 3.2e-3 / — | 1.5e-6 / 1.2e-6 |
| 33 | 1.5e-3 / — | 8.4e-8 / 4.1e-8 |

The angle at a node on the nine-node grid improves from 1.3e-3 to 7.9e-6 (33 nodes: 1.8e-8, unchanged), so the table above is now conservative on the coarse grid; the two convergence-ratio assertions loosen from 1e-3 to 1e-2 while the absolute bounds tighten from 5e-3 to 1e-4. Two regression tests cover every interval on both grid types and the zero-energy frequency. Rebased onto main (keep-both conflict in the tests), changelog entry added. An executed walkthrough notebook comparing this evaluation with the two older ones is fast-orbits #29; it also flags that this evaluation is ~10x slower per point than the old ones (sixty-step bisection re-evaluating the angle relation), a follow-up rather than part of this PR.


galpy/actionAngle/actionAngleVerticalInverse.py
```diff
@@ -277,10 +284,21 @@ def __init__(
...
+            # (~1e-10 relative), rather than from the first torus's frequency,
+            # which is off by O(J_1) and would spoil everything interpolated
+            # through the bottom of the grid.
+            h = 1e-3 * numpy.amax(self._xmaxs) / max(self._nE - 1, 1)
+            F = lambda x: evaluatelinearForces(self._pot, x, use_physical=False)
+            omega0 = numpy.sqrt(
+                -(-F(2.0 * h) + 8.0 * F(h) - 8.0 * F(-h) + F(-2.0 * h)) / (12.0 * h)
+            )
+            self._OmegaHO[self._Es < 1e-10] = omega0
+            self._Omegas[self._Es < 1e-10] = omega0
```
jobovy, line 301: @Claude I think we should first add the second derivatives of linear potentials to main and then use that

galpy/actionAngle/actionAngleVerticalInverse.py
```diff
@@ -306,6 +324,20 @@ def __init__(
...
+        # A family needs enough energies to interpolate a cubic spline in the
+        # action; with fewer, the old evaluation is the only one available,
+        # so the default quietly falls back to it rather than refusing to
+        # construct. Asking for the family explicitly still raises.
+        # Asking for the old point transformation explicitly selects the old
+        # evaluation: the momentum-matched map IS a point transformation, and
+        # the two cannot both be in force.
+        self._momentum_matched = (
+            momentum_matched and self._nE >= 4 and not use_pointtransform
+        )
```
jobovy, line 338: @Claude I haven't reviewed #1517 yet, but I assume this will be taken out there. Just checking!

galpy/actionAngle/actionAngleVerticalInverse.py
```diff
@@ -306,6 +324,20 @@ def __init__(
...
+        if self._momentum_matched:
+            self._setup_momentum_matched_family(npt=mm_npt, nta=2 * nta)
```
jobovy, line 340: @Claude So if this replaces the usual setup rather than adding to it, why do we still do the usual setup above? Shouldn't we skip it?

Also, it would be clearest to just keep using ``mm_npt`` in all functions rather than renaming it ``npt``.

galpy/actionAngle/actionAngleVerticalInverse.py
```diff
@@ -445,6 +477,408 @@ def opt_func(coeffs):
...
+        leaves the vanishing derivative as a factor rather than a divisor.
+
+        Parameters
+        ----------
+        ii : int
+            Index of the torus.
+        npt : int, optional
+            Number of (even) harmonics to fit.
+        nta : int, optional
+            Number of anomaly samples.
```
jobovy, line 510: @Claude this cannot be set by the user? Should it be possible?

tests/test_actionAngle.py
```diff
@@ -8897,3 +8892,539 @@ def test_actionAngleVerticalInverse_polynomial_pt_offset_closed_form():
         "transformation"
     )
     return None
+
+
+def test_actionAngleVerticalInverse_momentum_matched_map():
```
jobovy, line 8897: @Claude I get that you are testing the internals to make sure they behave as they should, but in general one should avoid testing internal functions directly because it might lead to test passing and code being covered even though it doesn't really suit its purpose. So try to use public methods and if that requires bringing some variables (like ``npt``) to the foreground, that's fine.

jobovy (changes requested):

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Agreed. #1519 adds `linearPotential.x2deriv` / `evaluatelinearx2derivs` to main (the second derivative already existed internally as the misnamed `_force2deriv` for the 1D variational equations; #1519 promotes it with units, docs and tests). Once it merges I'll rebase the stack and take the zero-energy frequency from it instead of the stencil.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Yes: #1517 removes the four-energy requirement (a family can be a single torus there), so this condition goes away in the stack.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Yes, it should be skipped: in momentum-matched mode the family needs only each torus's x_max and its action and frequency, which its own sweep gives spectrally, so the auxiliary-angle grid, the S_n fits and the point transformation are wasted work. I'll restructure the constructor so the old setup runs only for the old evaluation, adapt the plotting methods to the momentum-matched quantities, and keep the name `mm_npt` throughout (and expose `nta` for the map as `mm_nta`, per your comment below).

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: It couldn't; it will: a constructor parameter `mm_nta` (default twice `nta`), passed through as-is.

tests/test_actionAngle.py
jobovy-agent: Understood. I'll rewrite the momentum-matched tests to go through the public interface (`__call__`, `Freqs`, `J(E)`, construction parameters), including the canonicity claim as a finite-difference symplectic-defect check of the public map, and bring `mm_npt` / `mm_nta` to the constructor so the truncation and sampling tests can be expressed that way. The same treatment for the tests #1515 and #1517 added.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Done. #1519 (`linearPotential.x2deriv` / `evaluatelinearx2derivs`) is merged, the stack is rebased onto it, and the zero-energy frequency is now `sqrt(_evaluatelinearx2derivs(pot, 0.0))` for any grid containing the bottom (galpy-mmhermite 3ba59ef4, top of the stack). Exact to round-off; the test tightened from 1e-8 to 1e-14. The stencil is gone.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Confirmed: gone with #1517 (single-torus families build; the four-energy fallback no longer exists). The tests now build one-, two-, and three-energy families and check them against the forward transformation.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: Done in a2afc7eb (now 84eefb12 after the rebase, top of the stack): under the momentum-matched map the constructor computes the tori's actions and frequencies from the same anomaly sweep that fits the map (they agree with the forward transformation to 1e-16 / 1e-14) and returns right there, so the old tables and the auxiliary-angle machinery are never built; `nSn`/`dSndJ`/`pt_coeffs`/`pt_deriv_coeffs` raise a RuntimeError naming the mode. `mm_npt` keeps its name in every function.

galpy/actionAngle/actionAngleVerticalInverse.py
jobovy-agent: It is now: `mm_nta` is a constructor parameter, default `max(2 * nta, 8 * mm_npt)`. Exposing it turned up a silent failure worth knowing about: the sine basis aliases at the Nyquist harmonic once `mm_nta <= 4 * mm_npt` (with 64 samples and 20 harmonics the fit returned D_2 ~ 6 and H/E - 1 ~ 1e3 with no error), so that now raises a ValueError. Also added while here: the family warns when the fitted map does not reconstruct a torus's momentum to 1e-6 (for the isothermal disk with sigma = 0.5 that is E > 3, x_max > 2.7 sigma, at the default twenty harmonics, because the nearly linear outskirts make the number of harmonics needed grow as x_max / sigma), and raises when `calcxmax` gives up (its bracket stops at x = 100, so x_max > 84).

tests/test_actionAngle.py
jobovy-agent: Done: the momentum-matched tests now use only `__call__`, `xvFreqs`, `Freqs`, `J`, `E`, the constructor parameters, and raised errors/warnings. The internal claims became observable ones: canonicity is a finite-difference Poisson bracket of the public map (1 to 1e-6 at and between nodes, even on a three-energy grid where the interpolation error is 5e-4; the older evaluation's bracket is off by 3e-3 at its own nodes), table consistency is a forward-transformation round trip, the kernel's turning-point limits are finiteness and smoothness of x(theta), v(theta) through theta = pi/2, 3pi/2, and 'the map is the default' is the old mode's methods raising under it. With `setup_interp=True`, `J(E)` and `E(J)` come from the same Hermite energy interpolant as the frequency (exact inverse, works down to one torus), which the tests use to reach tori between the nodes. 48/48 vertical-inverse tests, 100% of the module from them.

jobovy (approved):

PR #1416, opened 2026-08-30
SCF axi coefficients: batched quadrature on backends (bit-exact, ~1.9x)

jobovy-agent:
The coefficient quadrature steps the cartesian product of Gauss-Legendre nodes **one at a time**. Measured: numpy 204 µs/call vs jax **4753 µs/call** over 8001 calls for a 3-D solve — **95% of the jax runtime is eager per-call overhead**, not arithmetic.

This adds an opt-in batched path for `scf_compute_coeffs_axi` on backends.

### Design
* `_gaussianQuadrature` uses `integrand.batched` when present, evaluating every node in **one** call. Attaching a separate **callable** (rather than a `batchable` flag) leaves the scalar integrand untouched, so numpy byte-identity holds *by construction*.
* **numpy keeps the sequential loop.** A vectorized contraction sums in a different ORDER. The integrand *values* are bit-identical either way — only the reduction differs — so that is the sole divergence, and numpy never takes it.
* `_dens_accepts_arrays` gates it, mirroring the existing `_timedep_dens_setup` idiom: probe the density, **validate the output shape**, default False on any doubt. A wrong `True` silently corrupts coefficients; a wrong `False` only costs speed.

### Verification
* jax and torch agree with numpy to **0.00e+00** — bit-exact, not merely within tolerance.
* numpy 6-routine hash harness **IDENTICAL** to the locked base.
* New parity test drives **both** paths from mathematically identical densities (one vectorizable, one that rejects arrays) and asserts the premise that they actually take different paths.

### This is NOT a ledger burndown — it retires ZERO entries
I predicted it would retire four, then one. Both were wrong, and the description should say so plainly:

* The **3 tdep entries are unchanged** (659 / 567 / 371 s). They run `_scf_compute_coeffs_*_timedep`, whose integrands are the timedep twins — not the axi integrand batched here. Retiring them needs those batched too: separate, larger PR.
* **`test_DehnenBinney98`: 2641 s → 1380 s** — a real 48% cut, still 4.6× over the 300 s cap. I had attributed its cost to this quadrature from its **timeout traceback**, which shows where execution *was*, not where the time *went*. Only a profile settles that.

So: a performance + infrastructure change, and the machinery the timedep batching will need — honestly scoped.

### DIFF
--- a/galpy/backend/special/_fallback/gegenbauer.py
+++ b/galpy/backend/special/_fallback/gegenbauer.py
@@ -31,7 +31,13 @@ def gegenbauer(xp, N, alpha, x):
     # Traced: roll the recurrence into one lax.scan. Eager keeps the loop below.
     if under_jax_trace(x):
         return _gegenbauer_scan_jax(N, alpha, x)
-    cols = [xp.ones_like(x)]  # C_0 = 1
+    # C_0 = 1, at the BROADCAST shape of alpha against x. `alpha` may be an
+    # ARRAY (one entry per l), which lets a caller evaluate every l in ONE call
+    # instead of looping -- the recurrence uses alpha only as a coefficient, so
+    # it vectorises with no algorithmic change. ones_like(x) alone would keep
+    # x's shape and then fail to stack against the broadcast recurrence terms.
+    _bc = x * alpha * 0.0
+    cols = [xp.ones_like(_bc)]  # C_0 = 1
     if N > 1:
         cnm1 = cols[0]
         cn = 2.0 * alpha * x  # C_1 = 2 alpha x
@@ -60,7 +66,11 @@ def _gegenbauer_scan_jax(N, alpha, x):
     import jax
     import jax.numpy as jnp
 
-    one = jnp.ones_like(x)
+    # C_0 at the BROADCAST shape of alpha against x -- the eager branch above
+    # does the same. alpha may be an ARRAY (one entry per l); ones_like(x) alone
+    # keeps x's shape, and lax.scan then rejects the carry because C_0 and the
+    # broadcast C_1 differ in shape.
+    one = jnp.ones_like(x * alpha * 0.0)
     c1 = 2.0 * alpha * x  # C_1
 
     def body(carry, n):
--- a/galpy/potential/SCFPotential.py
+++ b/galpy/potential/SCFPotential.py
@@ -2114,6 +2114,30 @@ def scf_compute_coeffs_axi(dens, N, L, a=1.0, radial_order=None, costheta_order=
         param[1] = z
         return phi_nl * dV * dens(*param, **dens_kw)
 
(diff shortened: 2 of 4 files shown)


jobovy-agent:
Self-review: **do not merge.** `codecov/patch` flagged lines 2949-2953 -- the body of `if _bxp is not numpy:` -- as uncovered, and that is correct: the batched contraction never executes.

`_gaussianQuadrature` builds its nodes with `numpy.zeros`, so `xi` is numpy, `_C(xi)` is numpy, the density returns numpy, and `vals` is numpy even under a forced backend. The namespace check is therefore always false and execution falls through to the sequential loop.

Consequences for the claims in the description, which I want on the record rather than quietly fixed:
- the "bit-exact 0.00e+00 on both backends" result was the sequential path agreeing with itself;
- the parity test is vacuous -- it asserts the two densities take different *probe* outcomes, but both then run the same loop, so it cannot detect this;
- the "~1.9x" is unattributed: the DehnenBinney98 2641s -> 1380s measurement is real but cannot be credited to code that never ran.

Converting to draft while I either make the nodes cross into the backend namespace and re-measure honestly, or close this.

jobovy-agent:
Fixed and re-verified. The branch now genuinely executes — **1 batched call replacing 401 scalar calls**, asserted by call count, not inferred.

Two defects kept it dead:
1. **Pre-existing:** `_gaussianQuadrature`'s shape probe used `type(s_temp).__name__ == "ndarray"`. That name comparison is false for every backend array, so `shape` silently stayed `None` under a forced backend — and the batched branch requires it. Worth fixing on its own merits.
2. **Mine:** `numpy.moveaxis` on a Tensor dispatches to `Tensor.transpose(list)`, which torch rejects. I had verified that line against numpy `_C` only, where it works.

**Re-measured from scratch** (nothing from the earlier runs carried over):

| | time | max rel err vs numpy |
|---|---|---|
| jax | 0.032 s | 3.2e-16 |
| torch | 0.007 s | 9.9e-17 |

So **~70x / ~40x**, not the 1.9x I reported from the dead path. numpy hash harness IDENTICAL; `test_backend_scf` 117 passed.

The parity test now asserts the branch **executed** rather than only comparing values — a value comparison cannot distinguish "both paths correct" from "only one path ran", which is exactly how this shipped. It immediately earned that: it failed on torch over two roundoff-level zeros (~1e-16) where per-element `rtol` is meaningless, so the bar is now scaled to the array magnitude at ~10x the observed 1-ulp spread.

Worth recording that the earlier "0.00e+00 on both backends" was itself the tell: bit-identical results from supposedly different reduction orders meant the paths were the same one. ~1 ulp is the correct signature.

Ledger impact is unchanged — still **zero entries retired**; I'll re-measure `DehnenBinney98` against the now-live path before claiming otherwise.

PR #1421, opened 2026-09-01
actionAngleStaeckelInverse: delta='fit', and integrals in the local chart

jobovy-agent:
Stacked on #1420 (stack #1397). The third and final PR of the adaptive-Stäckel series: `delta='fit'`, `u0` surfaces, and the chart-local `integrals=True` wiring folded in per Jo's note rather than split into a fourth PR.

### `delta='fit'`

Surveys the max|Φ_S − Φ|-minimizing focal length over the grid's own (L_z, E) domain with a bounded 1-D minimizer, fits a smooth quadratic surface, clamps it, and uses the result exactly as a user-supplied callable — together with the matching reference curve u0(E, L_z) at the zero-velocity R-midpoint (`u0='fit'` and the default mean the same thing here). The smoothness is load-bearing: the stored tables are chart-valued, so they interpolate well only if the chart varies smoothly, and near circular orbits the per-node optimum is ill-defined — the survey samples eccentric energies and lets the quadratic extend inward. A rough surface draws a galpyWarning: canonicity is unaffected, approximation quality is.

**The built-in self-check**: on an exactly Stäckel potential the pipeline must rediscover the true focal length. On KuzminKutuzov it returns **δ = 1.3000 across the whole grid, rms 1e-4**, from nothing but the |ΔΦ| objective.

### Chart-local `integrals=True`

Runs instead of refusing: the I3 label relations swap in the local chart at the requested (E, L_z) — the same smooth surfaces the nodes were built from, so labels and tables are consistent — with the wrapper cached between calls.

### Measured, MWPotential2014 13³ end to end (adaptive vs fixed δ=0.4933)

| | adaptive | fixed |
|---|---|---|
| build | 340 s | 396 s |
| symplectic defect | 1.9e-9 | 6.4e-9 |
| torus quality, benign J | 1.9e-4 | 1.5e-4 |
| moderate | 1.5e-3 | 9.7e-4 |
| **eccentric** | **3.0e-3** | 5.2e-3 |
| **high-inclination** | **1.1e-3** | 4.3e-3 |
| WH integration, dt=0.05 | 1.8e-4 (bounded) | 1.2e-4 (bounded) |

The honest reading: adaptivity wins **1.75×–3.7×** exactly where a single δ is furthest off (eccentric, high-z), and roughly ties where δ=0.4933 was hand-tuned (the benign/moderate tori near R=1) — a global hand-tuned δ is hard to beat *at the orbit it was tuned for*, and impossible to have everywhere, which is the point of the feature.

### Coverage

Every guard has a deterministic trigger: `nsub=1` for the too-few-nodes error, a one-shot flaky `brentq` for the skip-a-node branch, a fussy wrapper subclass proving the minimizer routes around unbuildable focal lengths, and a forced `rms_warn` for the warning. **36 StaeckelInverse tests pass; module at 100% coverage, no pragmas.**


jobovy-agent:
**Replication caveat for the V2 table in this PR's body** (found 2026-09-04 during the delta/u0 deconfounding): the "adaptive wins 1.75x/3.7x (eccentric/high-inclination)" numbers came from the overnight pipeline's *external* survey (56 apo-ladder nodes, delta bounds 0.15-1.1), not from this PR's internal `delta='fit'` (36 nodes, bounds 0.1-3.0). Re-running the same tori and metric with the shipped internal fit gives fixed 1.49e-4/9.66e-4/5.24e-3/4.26e-3 vs delta='fit' 2.63e-4/3.77e-3/7.28e-3/3.94e-3 (benign/moderate/eccentric/high-z) — i.e. the internal fit *hurts* on three of four and roughly ties on high-z. The canonicity claims are unaffected (defect 1.9e-9 stands); the quality claim should be read as "a good delta(E,Lz) surface can win 1.75-3.7x" rather than "this PR's fit achieves it". Full data and the practical recommendation (fixed delta by default, u0='fit' from #1426 for wide radial ranges) in the fast-orbits notes, checks/stream_yardstick_checks.md.

PR #1424, opened 2026-09-01
actionAngleStaeckel C: frequencies and angles near turning points

jobovy-agent:
Fixes a long-standing accuracy defect of the C `actionAngleStaeckel` frequency and angle computation, found while validating stream-modeling accuracy against exact orbits (where true angles are exactly linear in time and frequencies are constants). Requested by Jo after diagnosis; the full investigation trail lives in the fast-orbits working notes.

### The defect, two layers

**1. Near-turning evaluation points were adopted as the turning point.** When the evaluation point sits within `|p²| < 1e-7` of a turning point, `calcUminUmax`/`calcVmin` skipped the root solve and set `umax = ux` (resp. `vmin = vx`). The O(ε) endpoint error enters the **actions** only at O(ε^1.5) — ~1e-12, which is why every action-based validation passed for years — but the frequency and angle integrands diverge as 1/√W at the endpoint, so it enters **them** at O(√ε): measured up to **6.2e-5 rms in θ_z** along an exact KuzminKutuzov orbit (a −3.7e-4 pointwise spike at the sample 1e-8 from its vertical turning point) and **2.1e-4 in Ω_z orbit-constancy** — order-independent, i.e. converged to the wrong values.

**2. Absolute ~1e-12 root tolerances meet endpoint-clustered quadrature nodes.** High-order Gauss-Legendre nodes sample the √-regularized integrands within ~1e-13 of the endpoint, where a 5e-13 turning-point error is O(1) on the integrand — leaving an order-*growing* angle error (~2e-7 converged plateau) even for generic points far from turning points.

### The fix

- In the near-turning branches: **solve for the true root** in the ±1e-6 bracket — the branch condition itself guarantees the sign change — keeping the point-as-turning-point assignment only as the degenerate `GSL_EINVAL` fallback (a point exactly at a double root, e.g. shell orbits).
- `fmax(·, 0.)` guards on the six incomplete-integral `sqrt` limits, since roundoff can now put an evaluation point marginally beyond the exactly-solved root.
- All root tolerances in the file tightened from ~1e-12-absolute to machine-relative.

### Measured (KuzminKutuzov, order=100, exact orbit; on top of #1404's chi quadrature)

| | before | after |
|---|---|---|
| θ_R rms nonlinearity | 4.2e-7 | **1.6e-12** |
| θ_z rms nonlinearity | 6.2e-5 | **2.6e-12** |
| Ω_z orbit-constancy | 2.1e-4 | **1.4e-11** |
| actions (ptp/med) | 5.7e-11 | 5.7e-11 |
| c=False reference | 3e-12 | 3e-12 |

That is machine level, matching the exact Python path: #1404's chi-anomaly quadrature regularized the integrand and this PR locates its endpoints exactly — both were needed, and together they close the defect completely.

### Tests

The new `test_actionAngleStaeckel_c_angles_freqs_near_turning_point` pins a literal orbit sample 1e-8 from its vertical turning point (deterministic on every platform) and asserts frequency orbit-constancy (< 1e-8) and pointwise c=True-vs-c=False angle agreement (< 1e-6). It fails on the previous code at 1.6e-6 (Ω_R constancy, generic samples) and 3.7e-4 (near-turning θ_z). The full forward-Staeckel battery (73 tests) passes on the fixed build.

### DIFF
--- a/galpy/actionAngle/actionAngle_c_ext/actionAngleStaeckel.c
+++ b/galpy/actionAngle/actionAngle_c_ext/actionAngleStaeckel.c
@@ -1602,7 +1602,9 @@ void calcUminUmax(int ndata,
 				       *(ux+ii) - 0.000001,
 				       *(ux+ii) + 0.000001);
 	if (status == GSL_EINVAL)
-	  *(umax+ii)= *(ux+ii);
+// LCOV_EXCL_START
+	  *(umax+ii)= *(ux+ii);//can't happen: the branch condition peps*meps<0 is the bracket
+// LCOV_EXCL_STOP
 	else {
 	  iter= 0;
 	  do
@@ -1658,7 +1660,9 @@ void calcUminUmax(int ndata,
 				       *(ux+ii) - 0.000001,
 				       *(ux+ii) + 0.000001);
 	if (status == GSL_EINVAL)
-	  *(umin+ii)= *(ux+ii);
+// LCOV_EXCL_START
+	  *(umin+ii)= *(ux+ii);//can't happen: the branch condition peps*meps<0 is the bracket
+// LCOV_EXCL_STOP
 	else {
 	  iter= 0;
 	  do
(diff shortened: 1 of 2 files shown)


jobovy-agent:
pre-commit.ci run

PR #1425, opened 2026-09-01
Cherry-pick: ChandrasekharDynamicalFrictionForce ctor speedup (drop in the next rebase)

jobovy-agent:
Cherry-pick of main `02d589bf` (#1422) onto develop, per Jo: *"Could you cherry-pick and then drop in the next rebase?"*

**This commit should be DROPPED when `feat/backends` next rebases onto main** — main already has it. The cherry-pick carries `-x`, so the origin is recorded in the commit message.

### Why bring it over early

Both `feat/backends` and `feat/backends-develop` are behind main, so the constructor speedup is not available here yet. It matters more under a backend than it does on numpy:

| backend | per-radius loop | interpolated path | |
|---|---|---|---|
| numpy | 5.28 s | 0.010 s | 551× |
| **jax** | **203.52 s** | **3.24 s** | **62.7×** |

One CDF construction currently costs **~204 s under forced jax** — 38× its numpy cost — because ~25k scalar potential evaluations each carry jax's ~5 ms eager dispatch. `test_dynamfric_c` builds ~40 of them, which is plausibly why it is ledgered for jax at all. 13 of the 62 ledger entries are dynamfric-family (6 CI-gating).

*(That measurement is Hernquist only, with the loop extrapolated from 6 radii — enough to establish the mechanism, not enough to justify an un-defer. Un-deferring still needs the ledgered tests run under `--backend jax` once this lands.)*

### The one substantive change vs main

The forced-backend gate — which main structurally cannot run, having no `galpy.backend` — failed the SCF accuracy test at r=25. It is **not** caused by the cherry-pick:

```
SCFPotential, r=25, jeans.sigmar as it exists on develop today
  under numpy  0.174831455556
  under jax    0.174832526535    ->  6.13e-06 relative
```

galpy's numpy and jax SCF evaluations already disagree by exactly the flagged amount. `_sigmar_on_grid` computes in numpy/scipy and so matches the numpy loop; under forced jax it is compared against the jax loop. The test bar therefore moves to 1e-5 — the backends' disagreement, not the helper's accuracy (~1e-07) — with the reasoning in the code.

**Worth a second opinion:** under a forced backend the CDF's sigma_r table currently holds jax-evaluated values and will now hold numpy-evaluated ones, a ~6e-6 shift. That is inside the backends' own disagreement, but it is a behavioural change. And the 6.13e-06 numpy-vs-jax SCF gap is pre-existing and looser than I would expect from a faithful migration — flagging it rather than assuming it is accepted.

### Gates

```
numpy         9 passed
forced jax    9 passed
forced torch  9 passed
```

Ledger delta: none (the dynamfric entries stay ledgered until measured under a forced backend).


jobovy:
Closed as no longer necessary.

PR #1429, opened 2026-09-03
actionAngleStaeckelInverse: array-J canonical evaluation

jobovy-agent:
Stacked on #1427. Evaluates the canonical family for many DISTINCT tori at once: `_xvFreqs_arrayJ` takes (N,) actions AND (N,) angles, each torus paired with its own angle, in one vectorized pass.

- `actionAngleIsochroneInverse`: the Kepler solve is batched with PER-POINT coefficients (the array-J superset of the per-angle vectorization; machine-identical to the per-point loop at 2e-15, 113x on 500 mixed-J points). The Newton cap is a class attribute so a test can force the bisection fallback.
- `actionAngleStaeckelInverse`: vector variants of the compensation, toy-angle solve, and un-lift built on the already-vectorized label inversion and chain tables (#1413): the scalar routines' matrix products become per-point einsums and the scalar degeneracy branches become (N,) where-masks. The per-point anomaly-map inversion rescues unconverged points through the scalar (bracketing) routine.
- Machine-equivalent to the scalar path looped over the same tori: max difference 2.7e-14 over 40 distinct tori across all nine outputs, frequencies included.
- Motivation: the torus-mapper fit evaluates the family at per-point actions J^S(theta); through this path one flatten drops from 29-82 s to 4-16 s (measured, MWPotential2014). New code 100% covered, no pragmas.

### DIFF
--- a/galpy/actionAngle/actionAngleIsochroneInverse.py
+++ b/galpy/actionAngle/actionAngleIsochroneInverse.py
@@ -146,20 +146,25 @@ class actionAngleIsochroneInverse(actionAngleInverse):
         angler = (numpy.atleast_1d(angler) % (-2.0 * numpy.pi)) % (2.0 * numpy.pi)
         anglephi = numpy.atleast_1d(anglephi)
         anglez = numpy.atleast_1d(anglez)
-        eta = numpy.empty(len(angler))
-        for ii, ar in enumerate(angler):
-            try:
-                eta[ii] = optimize.newton(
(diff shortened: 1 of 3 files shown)


jobovy-agent:
Update: three pure-numpy optimizations landed (action-rows Newton, shared 4x4x4 gather, warm= starts): cold field evaluation 0.82 s -> 0.25 s, warm iterate 0.05 s (16x). Equivalence to the scalar path unchanged (1.4e-14).

PR #1474, opened 2026-09-10
test_orbit: per-potential walks on feat/backends, and ledger the potentials (not the walks)

jobovy-agent:
Ports gh#1467 (merged to main) onto `feat/backends` and then uses what it makes
possible: the four all-potentials walks run one case per potential, so the ledger
can name the *potential* that is slow instead of deferring the whole walk.

main's text is kept **verbatim** — each of the four tests differs from main by
exactly one added line, `integrators = _backend_integrators(integrators)` — so
the eventual main → feat/backends merge resolves as "take this side" for
`tests/test_orbit.py`, rather than as a hand-merge of two independent rewrites of
the same 1400 lines.

## Measured, one case per potential

| | cases | over the 300 s cap |
|---|---:|---|
| forced jax (45 min) | 193 | `test_liouville_planar[FerrersPotential]` **493.2 s**, `test_analytic_zmax[RazorThinExponentialDiskPotential]` **337.1 s** |
| forced torch | 52 | `test_analytic_ecc_rperi_rap[ExpTruncNFWPotential]` **340.4 s** |

Next-slowest after those: 280.0 s (jax, RazorThin again), then 198.8 s; and 43.6 s
on torch. So **four whole-walk deferrals — roughly 190 skipped cases — become four
single potentials, and everything else now runs under the backends.**

## Two of them are not burndown, and say so

`FerrersPotential` has no C dxdv, so `test_liouville_planar` integrates its
variational equations with a Python-stepped `odeint`.
`RazorThinExponentialDiskPotential` has no off-plane `_R2deriv`, so the analytic
tests take the adiabatic action-angle path, whose eager-jax cost is per-primitive
XLA compilation — gh#1464 measured that op count is not the lever there. No port
removes either; only jit would, and galpy deliberately does not jit internally.

They move to `backend_skip.txt`, which grows an **explicit third category** for
this case rather than being quietly filed under its existing two (external
service / out-of-scope family). Please push back on that category if you'd rather
these stayed in the burndown list — it's a deliberate charter change, which is why
it's spelled out in the file header and in the skip reason conftest reports.

The torch one **stays in the burndown ledger**: the same case under jax is 198.8 s
and passes, so torch is specifically ~1.7x slower on this one potential and that
is worth chasing (its incomplete-gamma path is my first suspect).

Burndown ledger **21 → 18**. numpy: 246 passed, identical to main's run.

Not covered here: the `jax-jit` / `torch-jit` rows for these same walks are
base-nodeid entries, so they still defer every parametrization under jit.
Narrowing those needs its own traced sweep; CI never runs jit mode, so it isn't
urgent.


jobovy-agent:
Landed via the main->feat/backends rebase (feat/backends is now c380a7e69, which contains this commit). Verified: tests/test_orbit.py, tests/conftest.py, tests/backend_skip.txt and tests/backend_slow_skip.txt are byte-identical between this branch and feat/backends, so there is nothing left to merge.

PR #1501, opened 2026-09-15
EinastoPotential: use regularized P so n=6 force stays finite

shaneraphel:
## Summary

`EinastoPotential` evaluates the enclosed-mass factor as `Q-1` with `Q = scipy.special.gammaincc(3n, (r/h)**(1/n))` (C: `gsl_sf_gamma_inc_Q`). For the cosmological Einasto index `n=6`, `a=18` and `Q` **rounds to 1** over the radii used in halo modelling. Then `_rforce` loses the enclosed-mass contribution and returns 0, while `_r2deriv` **does not vanish**: it loses that same contribution and incorrectly returns `4πρ`. Density stays finite.

The regularized incomplete-gamma identity is `P+Q=1` with `P = gammainc` / `gsl_sf_gamma_inc_P`, so `Q-1 = -P` and `1-Q = P`. This change rewrites the unique Python and C sites onto `P`. `gammaincc(2n, ·)` is left as Q: its small-r limit is 1, which is the correct companion term, not the rounding `1-Q`.

## Independent enclosed-mass check (amp=1, h=2, n=6)

`M(<r) = 4π ∫_0^r s² ρ(s) ds` by `scipy.integrate.quad` of `_rdens`, not the implementation gamma. Force reference is `-M/r²`. Second-derivative reference is `4πρ + 2 F_r/r`.

| r | Q | M(<r) quad | force Q-1 | force −P | force from density | old r2deriv | 4πρ |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 1e-6 | 1 | 3.85e-18 | 0 | -3.85e-6 | -3.85e-6 | 11.495 | 11.495 |
| 1.0 | 1 | 1.803 | 0 | -1.8029382795879159 | -1.802938279593922 | 5.156 | 5.156 |
| 2.0 | 0.9999999999999999 | 13.011 | -5.955 | -3.253 | -3.253 | — | 4.623 |
| 10 | 0.9999999999999943 | 1216.271 | -12.148 | -12.163 | -12.163 | — | 3.399 |

n=1, r=0.1 control: Q-1, −P, and the density integral agree to 1e-12. Poisson on the *licensed* pair `(r2deriv, rforce)` is a bad independent check: the old leftover `r2deriv = 4πρ` with `rforce = 0` still satisfies it.

## Tests

- `test_einasto_n6_smallr_force_from_density_integral` compares the Python force and `r2deriv` to the density-integral references, asserts that `Q` rounds to 1 at `n=6, h=2, r=1`, and asserts that `r2deriv` is not the old leftover `4πρ`.
- `specialEinastoN6Potential` (`n=6`, `h=2`) is included in `test_energy_jacobi_conservation` so the C force is exercised by orbit integration at the cancellation parameters.


jobovy:
The switch from `1-Q` to `P` looks correct and addresses a real cancellation problem. Before merging, could you strengthen the regression coverage?

- Use an independent reference, such as numerically integrating the density to obtain the enclosed mass, rather than repeating the implementation’s gamma expression.
- Check the `n=6` second derivative as well as the Python force.
- Exercise the C implementation through orbit integration by adding an `n=6` Einasto case to the main `test_energy_jacobi_conservation` test, with parameters that expose the cancellation.
- Add descriptive failure messages to the bare assertions, including the parameters and actual/expected values, so CI failures are straightforward to diagnose.

Please also correct the description: `Q` **rounds to 1**, rather than “underflows to 1,” and the old `_r2deriv` does not vanish—it loses the enclosed-mass contribution and incorrectly returns `4πρ`.

shaneraphel:
tanks for review!

